PR #15, opened 2026-02-09
Fix a sneaky bug in the fluxes

simone-silvestri:
before the pruning, I ll leave it here so we remember that this fix is required in NumericalEarth.
Hopefully this is the last fix and we can start connecting the two repos


simone-silvestri:
This should be ready to go (it has already been merged in ClimaOcean)

glwagner:
> before the pruning, I ll leave it here so we remember that this fix is required in NumericalEarth. Hopefully this is the last fix and we can start connecting the two repos

why would we leave it here rather than merging?

glwagner (approved):

PR #53, opened 2026-02-23
Add nonhydrostatic ocean, prescribed ocean, and four-ocean example

glwagner:
## Summary

- **`nonhydrostatic_ocean_simulation`**: new constructor for `NonhydrostaticModel`-based ocean LES with WENO(order=9) advection, TEOS-10 EOS, and surface flux boundary conditions. No free surface, no barotropic forcing, no split-explicit timestepping.

- **`PrescribedOcean`**: new prescribed ocean type following the `PrescribedAtmosphere` pattern. Takes `(grid, times)`, stores `FieldTimeSeries` for SST, SSS, and velocities. The ocean state does not evolve in response to fluxes — the atmosphere still feels the prescribed surface. `ComponentExchanger` creates snapshot CenterFields updated each coupling step via `interpolate_state!`.

- **`OceananigansModelSimulations`**: union type `Union{Simulation{<:HydrostaticFreeSurfaceModel}, Simulation{<:NonhydrostaticModel}}` that generalizes dispatch across both Oceananigans simulation types.

- **Example**: `breeze_over_two_oceans.jl` → `breeze_over_four_oceans.jl` with prescribed, slab, hydrostatic, and nonhydrostatic ocean models coupled to independent Breeze atmospheres. Five-column animation layout.

## Test plan

- [ ] `using NumericalEarth; nonhydrostatic_ocean_simulation(RectilinearGrid(size=(16,16), x=(0,100), z=(-50,0), topology=(Periodic,Flat,Bounded)))` constructs without error
- [ ] `PrescribedOcean(RectilinearGrid(size=16, x=(0,100), topology=(Periodic,Flat,Flat)))` constructs without error
- [ ] Run `examples/breeze_over_four_oceans.jl` end-to-end
- [ ] Verify all four SST lines appear in animation


src/Oceans/prescribed_ocean.jl
```diff
@@ -0,0 +1,187 @@
...
+    # Initialize from the first time snapshot
+    interior(T) .= interior(ocean.sea_surface_temperature)[:, :, :, 1]
+    interior(S) .= interior(ocean.sea_surface_salinity)[:, :, :, 1]
+    interior(u) .= interior(ocean.velocities.u)[:, :, :, 1]
+    interior(v) .= interior(ocean.velocities.v)[:, :, :, 1]
+
+    return ComponentExchanger((; u, v, T, S), nothing)
+end
+
+function net_fluxes(ocean::PrescribedOcean)
```
simone-silvestri: we can probably avoid allocating `net_fluxes` since they are not used?

src/Oceans/prescribed_ocean.jl
```diff
@@ -0,0 +1,187 @@
...
+- `velocities`: `NamedTuple` of `FieldTimeSeries` for `(u, v)`.
+"""
+struct PrescribedOcean{FT, G, Clk, SST, SSS, U, TI, F, ρ, C}
+    grid :: G
+    clock :: Clk
+    sea_surface_temperature :: SST
+    sea_surface_salinity :: SSS
+    velocities :: U
+    times :: TI
+    temperature_flux :: F
```
simone-silvestri: also this is maybe not required

src/Oceans/prescribed_ocean.jl
glwagner: but doesn't the atmosphere need the fluxes?

simone-silvestri: yeah, but that is in the interfaces the `net_fluxes` are the sum of the turbulent and non-turbulent fluxes used to force a particular component. So the ocean `net_fluxes` are not seen by the atmosphere because the atmosphere should compile its own net fluxes through `update_net_fluxes!(::Atmosphere...)`

src/Oceans/prescribed_ocean.jl
glwagner: I’m not sure we should let bulk properties like turbulent diffusivity leak into the ComponentInterfaces though…

navidcy (approved):

PR #54, opened 2026-02-23
Couple tightly ocean and sea ice

simone-silvestri:
This PR aims to couple tightly sea ice and ocean in stages rather than full timesteps. It is still a work in progress (I have just checked it works very nicely for the simple example shown in this PR which I will probably remove) because I still need to modify the interface to make sure that the coupling works with all the rest of the infrastructure (for example callbacks in the two specific components)

A video from the example, showing the difference between AB2 + FE with 5 minutes timestep and RK3 + RK3 with per-stage coupling but a 15 minutes timestep (so coupling expenses are equal to the AB2 - FE) 

https://github.com/user-attachments/assets/46e45c1b-e7d2-49dd-8f94-ed123b3df17b

### DIFF
--- a/test/test_earth_system_model.jl
+++ b/test/test_earth_system_model.jl
@@ -2,7 +2,7 @@ include("runtests_setup.jl")
 
 using CUDA
 using Oceananigans.OrthogonalSphericalShellGrids
-using NumericalEarth.EarthSystemModels: above_freezing_ocean_temperature!
+using NumericalEarth.EarthSystemModels: above_freezing_ocean_temperature!, RKCM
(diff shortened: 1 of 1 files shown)


glwagner:
nice

simone-silvestri:
We can use the same strategy also for breeze - ocean coupling when we use an RK3 for both

glwagner:
> We can use the same strategy also for breeze - ocean coupling when we use an RK3 for both

that's already the default, although one wrinkle is we have ssprk3 for the atmosphere. maybe the flavor of rk3 doesn't matter though

PR #56, opened 2026-02-24
Use `cpu_interpolating_time_indices` for `ECCOPrescribedAtmosphere`

xkykai:
Closes https://github.com/NumericalEarth/NumericalEarth.jl/issues/48


simone-silvestri:
Oh you might be right! Well, I would say we can merge this

PR #71, opened 2026-03-01
Added heat and freshwater flux diagnostics

### LINKED ISSUE #78: Output heat fluxes in W/m^2 and salinity fluxes in kg/(m^2*s)
We'd like a way to output heat fluxes in units of W/m^2. At the moment, some of the fluxes are temperature fluxes (meaning, units of K m/s) while some of them are actually heat fluxes (like, e.g., the frazil).

Similarly, for the salinity fluxes (g/kg m/s) we would like to have a way to output them in the more conventional as fresh water mass fluxes (kg / m / s^2).

This PR gives 
_Originally posted by @taimoorsohail in https://github.com/NumericalEarth/NumericalEarth.jl/pull/71#discussion_r2871041621_

taimoorsohail:
Closes #78 

We have added a way to output total heat and freshwater fluxes for comparison with other model outputs and observations, as well as for global budget diagnostics. 

With Codex

### DIFF
--- a/src/Diagnostics/interface_fluxes.jl
+++ b/src/Diagnostics/interface_fluxes.jl
@@ -3,6 +3,11 @@
 ### Temperature fluxes
 ###########################
 
+"""
+    frazil_temperature_flux(esm::EarthSystemModel)
+
+Return the two-dimensional frazil temperature flux (K m s⁻¹) in a coupled `esm`.
+"""
 function frazil_temperature_flux(esm::EarthSystemModel)
     ρᵒᶜ = esm.interfaces.ocean_properties.reference_density
     cᵒᶜ = esm.interfaces.ocean_properties.heat_capacity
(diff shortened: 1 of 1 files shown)


src/Oceans/assemble_net_ocean_fluxes.jl
```diff
@@ -159,6 +164,12 @@ end
     
         # salinity flux > 0 extracts salinity from the ocean --- the opposite of a water vapor flux
         Jˢao = - Sᵒᶜ * ΣFao
+        Jˢ_total = (1 - ℵᵢ) * Jˢao + Jˢio
+
+        # Freshwater-equivalent correction from salinity flux using S₀ = 35
+        # M_fw_from_Js = ρₒ * (-Jˢ/S₀), units kg m⁻² s⁻¹
+        S₀ = convert(eltype(Sₒ), 35)
+        ΣMao_with_salt_equiv = ΣMao - ρₒ * Jˢ_total / S₀
```
glwagner: It looks like this can be computed in a script, right, provided that we have `ΣMao` and `Jˢ_total`? That might be preferred, because then users can decide on their own reference salinity (and we don't compute+store this _every simulation_ if its not always needed)

glwagner:
I have some concern about this. While for an ocean model, this can indeed be thought of as a "unit conversion", this concept does not hold for other Earth system components. There is a state-dependent difference between heat flux and "temperature flux" (as an example) for an atmosphere model, where density is a state variable (not just a constant). Heat capacity is also a state variable (the atmosphere is a mixture of dry air and water species, and the heat capacity depends on the water species state).

Why don't we implement functions that extract diagnostics directly from `ComponentInterfaces`, rather than conceiving this as unit conversion?

glwagner:
> Here is says
> 
> https://github.com/NumericalEarth/NumericalEarth.jl/blob/466c01d3868b987e7f8ebb33d5e2a7b216d225d8/docs/src/appendix/notation.md?plain=1#L13
> 
> But we also use `J` for anything-flux, like tracer flux: Jᵀ is temperature flux, Jˢ is salinity flux. So it's not true that it always has units of mass flux, right?

That's correct. `J^q` for a quantity `q` has units `[q] * velocity` where `[q]` are the units.

PR #89, opened 2026-03-03
Add meridional heat transport diagnostic via ocean heat uptake method

### LINKED ISSUE #90: Add Meridional Heat Transport diagnostic

taimoorsohail:
Added a Meridional Heat Transport via computing the cumulative integral of the difference between the vertical, longitudinally-integrated OHC and the longitudinally-integrated surface heat flux, both in Joules.

(edited by @navidcy)

Closes #90


src/Diagnostics/meridional_heat_transport.jl
```diff
@@ -0,0 +1,189 @@
...
+    compute!(∫heat_flux)
+
+    model_time = esm.ocean.model.clock.time
+    model_iteration = esm.ocean.model.clock.iteration
+    state = get(meridional_heat_transport_states, esm, nothing)
+    state === nothing &&
+        (state = initialize_mht_state!(esm, ∫heat_flux, ∫ohc_tendency, model_time, model_iteration))
+
+    if model_iteration != state.last_iteration
+        Δt = max(0.0, model_time - state.last_time)
```
glwagner:
```suggestion
        Δt = max(0, model_time - state.last_time)
```

examples/meridional_heat_transport_ecco.jl
```diff
@@ -0,0 +1,127 @@
...
+mht_vT_mean  = deepcopy(mht_vT[1][1, :, 1])
+
+for j in 1:Nt
+    @info "iteration $j out of $Nt"
+    mht_OHC_mean += mht_OHC[j][1, :, 1]
+    mht_vT_mean  +=  mht_vT[j][1, :, 1]
+end
+
+@. mht_OHC_mean = mht_OHC_mean / Nt
+@. mht_vT_mean = mht_vT_mean / Nt
```
navidcy: surely there is an easier way to take the mean over one of the FieltTimeSeries dimensions other than what I did here? I think I tried to implement ReducedOperations on FieldTimeSeries dimensions at some point but never finished it...

@simone-silvestri, @glwagner would you do this differently?

src/Diagnostics/meridional_heat_transport.jl
```diff
@@ -0,0 +1,147 @@
...
+    cᵒᶜ = heat_capacity(esm.ocean)
+    ∂T_∂t = temperature_evolution_tendency(esm.ocean.model.timestepper)
+    heat_flux = net_ocean_heat_flux(esm) |> Field
+
+    ∂ℋ_∂t = Integral(ρᵒᶜ * cᵒᶜ * ∂T_∂t, dims=3) |> Field
+    ∫sum_dx = Integral(heat_flux + ∂ℋ_∂t, dims=1) |> Field
+    MHT = CumulativeIntegral(- ∫sum_dx, dims=2)
+
+    return MHT
+end
```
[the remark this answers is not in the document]
glwagner: What is `temperature_evolution_tendency`? Should this be just `temperature_tendency` (is a tendency ever about something other than "evolution"?)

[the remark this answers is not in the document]
glwagner: I propose just calling this diagnostic `meridional_heat_transport`; if there are more ways to compute the heat transport then these can be supported with a kwarg like

```julia
meridional_heat_transport(esm; method=:heat_content)
```

may want to specify this is `ocean_meridional_heat_transport`.

navidcy:
> what's the difference between "transport" and "flux"?

isn't it that flux is transport per unit area?

> I'm guessing these objects are located at `Center, Face, Center`?

Yes

src/Diagnostics/meridional_heat_transport.jl
```diff
@@ -0,0 +1,194 @@
...
+
+function meridional_heat_transport_via_ocean_heat_content(esm)
+    ρᵒᶜ = reference_density(esm.ocean)
+    cᵒᶜ = heat_capacity(esm.ocean)
+    ∂T_∂t = temperature_tendency(esm.ocean.model.timestepper)
+    heat_flux = net_ocean_heat_flux(esm) |> Field
+
+    ∂ℋ_∂t = Integral(ρᵒᶜ * cᵒᶜ * ∂T_∂t, dims=3) |> Field
```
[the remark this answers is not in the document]
navidcy: I think I sorted the notation + the derivatives style now; have a look!

navidcy:
> I have been thinking a bit about this.
> 
> I actually think something like
> 
> esm_regridded = regrid!(esm, Regridder)
> 

But then we create a whole new esm on a lat-lon grid? And we are timestepping the original `esm`? How is the `esm_regridded` moving forward?

Btw, I do think adding the regridding option is out of scope for _this_ PR, right?

taimoorsohail:
> > I have been thinking a bit about this.
> 
> > 
> 
> > I actually think something like
> 
> > 
> 
> > esm_regridded = regrid!(esm, Regridder)
> 
> > 
> 
> 
> 
> But then we create a whole new esm on a lat-lon grid? And we are timestepping the original `esm`? How is the `esm_regridded` moving forward?
> 
> 
> 
> Btw, I do think adding the regridding option is out of scope for _this_ PR, right?

Yes out of scope for this PR. 

The regridded ESM is purely for output writing so would need to be created every time the MHT is output. There wouldn't be any point in time stepping it.

taimoorsohail:
> > Why is there a global heat budget imbalance? Shouldn't the blue line integrate to zero?
> 
> 
> 
> What does "global heat imbalance" mean? What do you see to say so? If it's that the curves should integrate to zero then does that need to hold for every snapshot? I can compute the integrals and check if they are zero; I can't do the integrals by eye. So it should be that $$\int_{-90^\circ}^{+90^\circ} \mathrm{MHT}(\phi) \mathrm{d}\phi = 0$$? It could be that the blue is closer to zero than the orange.
> 
> 
> 
> > You are running the RYF correct? I think we need to make sure the diagnostics match at least globally before we merge?
> 
> 
> 
> I'm running with a prescribed atmosphere:
> 
> 
> 
> https://github.com/NumericalEarth/NumericalEarth.jl/blob/46efae2a8eb68cf7c517e7ceaa8ff6158db020fb/examples/meridional_heat_transport_ecco.jl#L46-L47
> 
> 
> 
> I'm also running on a near-global lat-lon grid from 80S to 80N:
> 
> 
> 
> https://github.com/NumericalEarth/NumericalEarth.jl/blob/46efae2a8eb68cf7c517e7ceaa8ff6158db020fb/examples/meridional_heat_transport_ecco.jl#L19
> 
> 
> 
> Could that be the culprit?
> 
> 
> 
> I'm wondering: computing the heat transport directly via $v T$ could be wrong for a near-global grid but given that $\int v \mathrm{d}A =0$ by construction we may be not getting any apparent imbalance. But the for surface heat fluxes the missing Arctic matters?
> 

The MHT is a cumulative integral in latitude, right? In that case, the blue line should start and end at 0 if the global heat budget is closed. That means that there is no residual between cumulatively integrated heat flux and OHC tendency. Does that make sense?

taimoorsohail:
Just pinging @navidcy -- would be great to merge this!

navidcy:
> The MHT is a cumulative integral in latitude, right? In that case, the blue line should start and end at 0 if the global heat budget is closed. That means that there is no residual between cumulatively integrated heat flux and OHC tendency. Does that make sense?

So given that there is a discrepancy, does that imply that the residual term (see the docstring derivation) is not integrated to zero?

src/EarthSystemModels/earth_system_model.jl
```diff
@@ -336,6 +336,8 @@ function restore_prognostic_state!(osm::EarthSystemModel, state)
     restore_prognostic_state!(osm.atmosphere, state.atmosphere)
     restore_prognostic_state!(osm.sea_ice, state.sea_ice)
     restore_prognostic_state!(osm.interfaces, state.interfaces)
+    auxiliary_state = hasproperty(state, :auxiliary) ? state.auxiliary : nothing
+    restore_auxiliary_state!(osm, auxiliary_state)
```
navidcy: @taimoorsohail is this still needed?

src/Diagnostics/meridional_heat_transport.jl
navidcy, line 1: @giordano, after #135 no file calls `test_diagnostics_1.jl` or `test_diagnostics_2.jl` explicitly.
Do these tests still run because of 

https://github.com/NumericalEarth/NumericalEarth.jl/blob/558bf7605a3ef50277f32ea8bd1d0e89c0fabdf2/test/runtests.jl#L11

?

navidcy: I'm answering this myself saying "yes". I'm pretty confident now it's the case.

giordano: Yes, [`find_tests`](https://juliatesting.github.io/ParallelTestRunner.jl/stable/api/#ParallelTestRunner.find_tests) automatically finds all `.jl` files in the given directory, and its subdirectories recursively.

navidcy: Thanks!

simone-silvestri:
I think our example folder is becoming a bit large. I think we could either add the example to the examples that are run in CI or maybe add a `validation` folder like in Oceananigans. What do you guys think?

examples/meridional_heat_transport_ecco.jl
simone-silvestri: is this still an issue? We can add a method in Oceananigans to deal with reductions in time or just solve the 
```
ERROR: MethodError: conditional_length(::FieldTimeSeries{…}, ::Int64) is ambiguous.
```
which should be rather easy

src/EarthSystemModels/earth_system_model.jl
taimoorsohail: Nope this is a vestige of the old method of saving auxiliary checkpoints

examples/meridional_heat_transport_ecco.jl
simone-silvestri: see https://github.com/CliMA/Oceananigans.jl/pull/5456

navidcy:
#159 added the MHT diagnostic via $\int v T dx$.
This PR adds the MHT via the ocean heat update method.

src/Diagnostics/meridional_heat_transport.jl
```diff
@@ -81,15 +160,48 @@ function meridional_heat_transport(esm::EarthSystemModel; reference_temperature=
     grid isa OrthogonalSphericalShellGrid &&
         throw(ArgumentError("meridional_heat_transport diagnostic does not work on OrthogonalSphericalShellGrid at the moment; use LatitudeLongitudeGrid."))
 
-    FT = eltype(esm)
-    reference_temperature = convert(FT, reference_temperature)
+    if method isa MeridionalHeatFluxMethod
+        FT = eltype(esm)
+        reference_temperature = convert(FT, reference_temperature)
+        return meridional_heat_transport_via_meridional_heat_flux(esm; reference_temperature)
+    elseif method isa OceanHeatContentTendencyMethod
```
glwagner: how about "MeridionalFluxMethod" and "TendencyMethod" ?

if I understand these are both ocean-specific methods, and also "heat" specific method.

also the difference between "heat tendency" and "heat content tendency" isn't completely clear to me

src/Diagnostics/meridional_heat_transport.jl
```diff
@@ -81,15 +160,48 @@ function meridional_heat_transport(esm::EarthSystemModel; reference_temperature=
...
-
     MHT = Integral(ρᵒᶜ * cᵒᶜ * v * (T - reference_temperature), dims=(1, 3))
     return MHT
 end
+
+function meridional_heat_transport_via_ocean_heat_content(esm)
+    ρᵒᶜ = reference_density(esm.ocean)
+    cᵒᶜ = heat_capacity(esm.ocean)
+    ∂t_T = temperature_tendency(esm.ocean.model.timestepper)
+    𝒬ᵃᵒₙₑₜ = net_ocean_heat_flux(esm) |> Field
```
glwagner: what is the difference between `𝒬ᵃᵒₙₑₜ` and `𝒬ᵃᵒ`

src/Diagnostics/meridional_heat_transport.jl
navidcy: Yeah, sounds good!

src/Diagnostics/meridional_heat_transport.jl
navidcy: I think it's part of the discussion why the budget doesn't close; there might be a difference or there may be a double counting of some parts.

But if there isn't any in the end, we should drop the "net".

src/Diagnostics/meridional_heat_transport.jl
navidcy: [658923d](https://github.com/NumericalEarth/NumericalEarth.jl/pull/89/commits/658923ded4d06d9178c07fbce68baece3fe73bcd)

navidcy:
```Julia
using NumericalEarth
using Oceananigans
using Oceananigans.Units
using Dates
using Statistics
using Printf

using CUDA; CUDA.device!(3)

arch = GPU()
Nx = 360
Ny = 180
Nz = 50

depth = 5000meters
z = ExponentialDiscretization(Nz, -depth, 0; scale = depth/4)

underlying_grid = TripolarGrid(arch; size = (Nx, Ny, Nz), halo = (5, 5, 4), z)
underlying_grid = LatitudeLongitudeGrid(arch; size = (Nx, Ny, Nz), halo = (5, 5, 4), z, longitude = (0, 360), latitude = (-80, 80))
bottom_height = regrid_bathymetry(underlying_grid;
                                  minimum_depth = 10,
                                  interpolation_passes = 10,
                                  major_basins = 2)
grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(bottom_height);
                            active_cells_map=true)

free_surface       = SplitExplicitFreeSurface(grid; substeps=70)
momentum_advection = WENOVectorInvariant(order=5)
tracer_advection   = WENO(order=5)
vertical_mixing = NumericalEarth.Oceans.default_ocean_closure()
ocean = ocean_simulation(grid; momentum_advection, tracer_advection, free_surface,
                         closure=(vertical_mixing,))
sea_ice = sea_ice_simulation(grid, ocean; advection=tracer_advection)

date = DateTime(1993, 1, 1)
ecco_set = MetadataSet(:temperature, :salinity,
                       :sea_ice_thickness, :sea_ice_concentration;
                       dataset = ECCO4Monthly(), date)

set!(ocean.model,   ecco_set)   # T, S
set!(sea_ice.model, ecco_set)   # h, ℵ

atmosphere = JRA55PrescribedAtmosphere(arch)
land       = JRA55PrescribedLand(arch)
radiation  = JRA55PrescribedRadiation(arch)
esm = OceanSeaIceModel(ocean, sea_ice; atmosphere, land, radiation)

simulation = Simulation(esm; Δt=20minutes, stop_time=5*365days)

wall_time = Ref(time_ns())

function progress(sim)
    ocean = sim.model.ocean
    u, v, w = ocean.model.velocities
    T = ocean.model.tracers.T
    e = ocean.model.tracers.e
    Tmin, Tmax, Tavg = minimum(T), maximum(T), mean(view(T, :, :, ocean.model.grid.Nz))
    emax = maximum(e)
    umax = (maximum(abs, u), maximum(abs, v), maximum(abs, w))

    step_time = 1e-9 * (time_ns() - wall_time[])

    msg1 = @sprintf("time: %s, iter: %d", prettytime(sim), iteration(sim))
    msg2 = @sprintf(", max|uo|: (%.1e, %.1e, %.1e) m s⁻¹", umax...)
    msg3 = @sprintf(", max(e): %.2f m² s⁻²", emax)
    msg4 = @sprintf(", wall time: %s \n", prettytime(step_time))

    @info msg1 * msg2 * msg3 * msg4

    wall_time[] = time_ns()

     return nothing
end

# And add it as a callback to the simulation.
add_callback!(simulation, progress, IterationInterval(200))

mht = Field(meridional_heat_transport(esm))

ocean.output_writers[:mth] = JLD2Writer(ocean.model, (; mht);
                                        schedule = TimeInterval(3hours),
                                        filename = "ocean_one_degree_mht",
                                        overwrite_existing = true)

run!(simulation)

##

using Oceananigans

mht  = FieldTimeSeries("ocean_one_degree_mht.jld2", "mht"; backend = OnDisk())

times = mht.times
Nt = length(times)

grid = mht.grid
Ny = size(mht.grid, 2)

mht_mean  = deepcopy(mht[1][1, :, 1])

for iter in 1:Nt
    @info "iteration $iter out of $Nt"
    mht_mean  +=  mht[iter][1, :, 1]
end

@. mht_mean = mht_mean / Nt

using CairoMakie

fig = Figure()
ax = Axis(fig[1, 1], xlabel="latitude (deg)", ylabel="MHT (PW)")

φ = φnodes(grid, Face())

lines!(ax, φ, mht_mean[1:Ny+1]  / 1e15, linewidth=4)

save("mht.png", fig)
```

navidcy:
(pasting here the example that I used to make the plots; x-ref: #277)

taimoorsohail:
OK, I have implemented a major refactor here. The below were the identified issues:

1) Because the heat flux field HF(t-1) is coupled to [OHC(t) - OHC(t-1)], there was a need to have an on-the-fly cache to outputthe MHT, which caches the current and previous time step state. This is especially important for the `AverageTimeInterval()` schedule because the heat budget needs to be computed for every model timestep for the rolling average. 

2) The MHT calculation does naive integration on the `TripolarGrid`, so it is quite different to observational estimates of MHT, and doesn't cover the same latitude ranges. So, there was a need to use `ConservativeRegridding` on-the-fly when outputting the MHT. The current implementation of `ConservativeRegridding` was to `regrid!` every time-step - a very costly workflow. 

3) The MHT exposes a broader tracer budget test that we have been working on (not really an issue but more of a comment). 

To resolve these issues I:

1) Introduced a new field in the ESM: `esm.interfaces.budgets.ocean_heat` with the fields:

```Julia
budget.tendency       # ΔOHC / Δt
budget.surface_flux   # flux applied during that interval
budget.radiative_flux
budget.residual       # combined column budget used by MHT
```

Under the hood, these use the same `interface_fluxes.jl` diagnostic functions we coded earlier. But, they should ensure exact closure. The `test_tracer_budget.jl` file can be substantially simplified as a result (`Integral(budget.residual) == 0`) is literally all you need. 

2) A new `AbstractField` type, `RegriddedField` has been added in https://github.com/CliMA/Oceananigans.jl/pull/5792; this type avoids the multiple `regrid!`s

cc @navidcy @simone-silvestri

taimoorsohail:
BTW @simone-silvestri the new mass flux/budget/conservation could also be wired through `esm.budgets`

simone-silvestri:
Can we separate the diagnostics from the model? If we need to have introspection inside the model we can use a callback. I think we can do something like we did for the `VarianceDissipation` diagnostic in Oceananigans which is attached to a simulation but can go pick into it through predefined callbacks:

https://github.com/CliMA/Oceananigans.jl/tree/main/src/Models/VarianceDissipationComputations

the usage is shown here

https://github.com/CliMA/Oceananigans.jl/blob/main/validation/implicit_dissipation/variance_dissipation.jl

something like:
```julia
model = OceanSeaIceModel(....)
sim = Simulation(model, ...)
Tb = BudgetComputation(:temperature, model)
add_callback!(Tb, sim)
```

taimoorsohail:
OK @navidcy @simone-silvestri 

I have again refactored a bit to match the design we were discussing: 

1) `BudgetComputation(:temperature, esm)` is now a callback which is attached to the simulation, and triggers caching of previous heat fluxes, saving in checkpoints etc. Currently only temperature is supported.
2) `meridional_heat_flux` now errors if `grid::OrthogonalSphericalShellGrid` and `method::MeridionalFluxMethod()`, because `v` is not meaningful in this case
3) `meridional_heat_flux` now errors if `grid::OrthogonalSphericalShellGrid` and `method::MeridionalFluxMethod()` and `destination_grid=nothing`
4) `meridional_heat_flux` now defaults to `method::MeridionalFluxMethod()` and `destination_grid=nothing`

The workflow for the two methods is:

```Julia
### Assume some simulation has been created which is a simulation of an EarthSystemModel

temperature_budget = BudgetComputation(:temperature, simulation.model) 
# only :temperature is supported at the moment

add_callback!(simulation, temperature_budget)
# This defaults to IterationInterval(1); which it MUST be to have correct caching behaviour
 
mht_vT = Field(meridional_heat_transport(simulation, MeridionalFluxMethod())) 
# This currently is not supported with Orthogonal grids, only LatitudeLongitudeGrids

mht_OHC = Field(meridional_heat_transport(simulation; destination_grid)) 
# This triggers TendencyMethod() by default. Note that destination_grid is REQUIRED # if grid:: OrthogonalSphericalShellGrid. 

simulation.model.ocean.output_writers[:mht] = JLD2Writer(simulation.model.ocean.model, (; mht_OHC, mht_vT);
                                        schedule = TimeInterval(3hours),
                                        filename = "ocean_one_degree_mht",
                                        overwrite_existing = true)

run!(simulation)
```

What do you think? Tests are passing, but I haven't incorporated this new callback into `test_tracer_budgets.jl` yet. This is a breaking change.

navidcy:
will rerun the ECCO example

navidcy:
I run for 20 days and got this for meridional heat transport using the two methods!

image

src/Diagnostics/meridional_heat_transport.jl
```diff
@@ -63,34 +161,111 @@ atmosphere = PrescribedAtmosphere(grid, [0.0])
...
+    budgets = [callback.func for callback in values(simulation.callbacks)
+               if callback.func isa BudgetComputation]
+
+    if isempty(budgets)
+        throw(ArgumentError("TendencyMethod() requires a BudgetComputation callback. " *
+                            "Create a budget and register it with " *
+                            "add_callback!(simulation, budget) first."))
+    elseif length(budgets) > 1
+        throw(ArgumentError("TendencyMethod() requires exactly one BudgetComputation callback, " *
+                            "but the simulation has $(length(budgets))."))
```
navidcy: What's the issue if I want a simulation in which I also want to compute the salt budget or the u-momentum budget? Why should an error be thrown then?

src/Diagnostics/budget_diagnostic.jl
```diff
@@ -0,0 +1,184 @@
...
+    applied_radiative_heat_flux :: A
+    residual :: B
+end
+
+Base.summary(budget::BudgetComputation) =
+    string("BudgetComputation(:", budget.tracer_name, ") on ", summary(budget.residual.grid))
+
+Base.show(io::IO, budget::BudgetComputation) = print(io, summary(budget))
+
+function BudgetComputation(tracer_name::Symbol, esm::EarthSystemModel)
```
navidcy, line 34: I believe the docstring should be in the constructor.

Also:
- if temperature is hardcoded, why don't we call this TemperatureBudget?
- Can we write down the actual budget that the diagnostic computes?

src/Diagnostics/interface_fluxes.jl
```diff
@@ -34,11 +67,16 @@ end
 """
     net_ocean_temperature_flux(esm::EarthSystemModel)
 
-Return the net temperature flux (K m s⁻¹) at the ocean's surface in a coupled `esm`.
+Return the complete outward-positive temperature flux (K m s⁻¹) for the ocean
```
navidcy, line 70: what does "outward-positive" mean? It's confusing? Is this different that all other conventions in Oceananigans? Outward of what?

Project.toml
```diff
@@ -75,6 +76,7 @@ ArchGDAL = "0.10"
 Breeze = "0.5.3, 0.6, 0.8"
 CDSAPI = "2.2.1"
 CFTime = "0.1, 0.2"
+CUDA = "6.2.1"
```
navidcy, line 79: why do we add CUDA as dependency?

Project.toml
```diff
@@ -8,7 +8,9 @@ authors = ["NumericalEarth contributors"]
 Accessors = "7d9f7c33-5ae7-4f3b-8dc6-eff91059b697"
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 CFTime = "179af706-886a-5703-950a-314cd64e0468"
+CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
```
navidcy, line 11: why do we add CUDA as dependency?

src/Diagnostics/interface_fluxes.jl
```diff
@@ -4,10 +4,43 @@ using ..SeaIces: FreezingLimitedEarthSystemModel
...
+    model = esm.ocean.model
+    fields = merge(Oceananigans.fields(model), (; T))
+    return KernelFunctionOperation{Center, Center, Nothing}(evaluate_additional_temperature_flux,
+                                                            model.grid,
+                                                            model.clock,
+                                                            fields,
+                                                            additional_flux)
+end
+
+function ocean_additional_temperature_flux(esm, T)
```
navidcy, line 27: what is `ocean_additional_temperature_flux`? I can't understand...
Perhaps the confusing bit is the "additional"; additional to what?

glwagner:
> I run for 20 days and got this for meridional heat transport using the two methods!
> 
> image

when do the methods differ?

src/Diagnostics/interface_fluxes.jl
glwagner: We have the convention (better or worse) of fluxes that are positive upward. This is outward positive from the surface though.

glwagner: To me "outward positive" means that the flux convention is positive if it transports a quantity out of a closed domain.

If you imagine a box with top and bottom, outward positive is +z at the surface ("outwards" is "up") and -z at the bottom ("outwards" is "down").

navidcy:
They were very different before because the ocean-heat-content tendency diagnostic method was wrong. In principle, they should be the same if you take into account all terms properly.

A good test for the ocean-heat-content method is whether it closes, that is whether it goes to zero at the northern-most latitude. Before the latest refactor by @taimoorsohail and @simone-silvestri it wasn't; see https://github.com/NumericalEarth/NumericalEarth.jl/pull/89#issuecomment-4056734234

However, when the other day I run for 2 years and plotted, I got results with a bit of a disagreement:

mht-1

This is still much closer compared to https://github.com/NumericalEarth/NumericalEarth.jl/pull/89#issuecomment-4056734234

src/Diagnostics/interface_fluxes.jl
navidcy: OK, gotcha!

So outward-positive from the surface is exactly equivalent to the convention that we currently have anyway, right? That's perhaps why I was a bit confused. Was thinking about it after I commented... I felt that clarification was needed if something is out of the ordinary but then I couldn't understand the difference...

glwagner:
Oh ok. So maybe we only want one method in the end, but having the two together is helpful for validation during development.

PR #111, opened 2026-03-05
Use area-scaled biharmonic viscosity in coarse examples

glwagner:
## Summary
- Add `area_scaled_biharmonic_viscosity()` in `src/Oceans/ocean_simulation.jl` returning a `HorizontalScalarBiharmonicDiffusivity` with `ν = Az² / timescale` (default 15 days), so the effective viscosity adapts to local cell area.
- Export `area_scaled_biharmonic_viscosity` from `NumericalEarth` and the `Oceans` submodule.
- Use it as the viscous closure in the `one_degree_simulation` and `global_climate_simulation` examples, alongside the existing Gent-McWilliams `IsopycnalSkewSymmetricDiffusivity`. In `global_climate_simulation` it replaces the constant `ν=1e12` biharmonic viscosity; in `one_degree_simulation` it is added next to GM and CATKE.

## Test plan
- [ ] Verify the one-degree simulation runs with the new closure
- [ ] Verify the global climate simulation runs with the new closure
- [ ] Check that grid-scale artifacts are suppressed


glwagner:
> out of curiosity why don't we want GM, especially in the 1degree model?

didn't you find that it has no effect? I've been telling people this...

glwagner:
> I would still recommend having some GM at 1 degree as the solutions are less noisy.

Do these solutions have biharmonic viscosity?

glwagner:
@xkykai did you try running with only the symmetric part of the diffusivity? Theoretically we might expect that we don't need the skew part (baroclinic conversion is represented) but we still a symmetric diffusivity (bc tracer stirring not captured if no eddies)

xkykai:
> > I would still recommend having some GM at 1 degree as the solutions are less noisy.
> 
> Do these solutions have biharmonic viscosity?

No they don't. I haven't run comparison of GM with biharmonic viscosity as well.

glwagner:
> > @xkykai did you try running with only the symmetric part of the diffusivity? Theoretically we might expect that we don't need the skew part (baroclinic conversion is represented) but we still a symmetric diffusivity (bc tracer stirring not captured if no eddies)
> 
> no, don't think we ever turned off skew viscosity completely

I'll add ISSD back here, but with only the symmetric turned on. Maybe that will provide some inspiration for future work.

xkykai (approved):
up to your discretion to merge!

simone-silvestri:
then remove the "probably". At one degree we need it.

glwagner:
@xkykai @simone-silvestri changing this PR so that it just adds viscosity to the coarse example, but does not remove GM. Let me know what you think.

xkykai:
lgtm!

simone-silvestri (approved):

navidcy:
Docs break because of Metadataset fixes included in #277 or #283

simone-silvestri:
I think tests should pass now. If docs pass I will merge

PR #113, opened 2026-03-09
New inpainting procedure

simone-silvestri:
which smooths out the inpainted field using a discrete laplacian. It takes a bit more time but it might prevent errors near the coasts

### DIFF
--- a/src/DataWrangling/inpainting.jl
+++ b/src/DataWrangling/inpainting.jl
@@ -16,7 +16,7 @@ end
 
 Diffusive inpainting algorithm. Vertical levels are processed top-to-bottom sequentially:
 at each level, missing values are first filled by horizontal neighbor propagation,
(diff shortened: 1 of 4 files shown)


src/DataWrangling/inpainting.jl
```diff
@@ -11,6 +11,39 @@ struct NearestNeighborInpainting{M}
     maxiter :: M
 end
 
+"""
+    DiffusiveInpainting{M, T}
+
+Diffusive inpainting algorithm. Vertical levels are processed top-to-bottom sequentially:
+at each level, missing values are first filled by horizontal neighbor propagation,
+then smoothed by iteratively solving the discrete Laplace equation ∇²ϕ = 0.
```
navidcy:
```suggestion
at each level, missing values are first filled by horizontal neighbor propagation,
then smoothed by iteratively solving the discrete Laplace equation ``∇²ϕ = 0``.
```

PR #121, opened 2026-03-11
Add filename property to Metadata

### LINKED ISSUE #115: Have `Metadata` path support in DataWrangling. Support multiple variables (names) inside `Metadata`.
The `Metadata` struct inside `src/DataWrangling/metadata.jl` is defined as follows:

```julia
struct Metadata{V, D, B}
    name  :: Symbol
    dataset :: V
    dates :: D
    bounding_box :: B
    dir :: String
end
```

for datasets where we have multiple variables of interest, it would be nice to use one `Metadata` to encapsulate this without having to define a new symbol. Also would be nice to encode the paths of objects rather than just the directory they live in. 

I propose something like this:

```julia
struct Metadata{S, V, D, B, P}
    name  :: S
    dataset :: V
    dates :: D
    bounding_box :: B
    dir :: String
    paths :: P
end
```

cc @glwagner

glwagner:
Plus allow `name` to be any type, permitting `Metadata` that represent multiple variables.

Closes #115 

cc @awiteck


awiteck (approved):

simone-silvestri (approved):

glwagner:
Anyways moving to AWS has benefits in the longer run anyways so I am definitely not opposed to that.

navidcy:
Shall we merge this PR?

simone-silvestri:
That would be a great benefit I think!

src/DataWrangling/ERA5/ERA5.jl
```diff
@@ -186,43 +186,37 @@ function bbox_strs(c)
     return first, second
 end
 
-function metadata_prefix(metadata::ERA5Metadata)
```
ewquon, line 189: @glwagner Why change the function signature rather than extending it?

Issue #28, opened 2026-02-15
Should we spawn more agents for the tests?


simone-silvestri:
At the moment we have 4 agents on tartarus, one connected to each GPU.
In this repo we use only GPU 2 for GPU-CI and GPU 3 for documentation.
GPU 0 is free (benchmarks in Oceananigans use GPU 0) and GPU 1 is used by the Oceananigans docs (and also by ClimaOcean docs).

<img width="903" height="428" alt="Image" src="https://github.com/user-attachments/assets/3b9d7d37-d379-407b-8880-7a32ecd47898" />

One agent can only handle one job at a time, so at the moment, GPU2 is running all the GPU-CI tests for this repo sequentially. However, these tests are quite small in size, so I was wondering if it was a good idea to create new runners all connected to GPU 2 to run more jobs in parallel.

Would this be problematic? I think it is how we do it in oceananigans where one GPU handles more jobs at a time.

@glwagner @giordano @navidcy

giordano:
To understand the setup: this node has 4 GPUs in total?  How many CPUs?

One option could be to use ParallelTestRunner, to parallelise test within a single job, if you have enough CPUs to handle them.  This is what we do in Breeze.jl.

giordano:
Here is the documentation of the package: https://juliatesting.github.io/ParallelTestRunner.jl/stable/.

Here is Breeze's `runtests.jl`: https://github.com/NumericalEarth/Breeze.jl/blob/b764d2ef4ba887def0d4c5724e1c5b411afd948b/test/runtests.jl.

Happy to answer questions about `ParallelTestRunner` 😃

PR #127, opened 2026-03-13
Move CI to self-hosted AWS runners for GPU tests

awiteck:
Use newly-created NumericalEarth Docker images


.github/workflows/ci.yml
```diff
@@ -159,19 +128,29 @@ jobs:
...
+      - name: Copy LocalPreferences
+        run: cp /usr/local/share/julia/environments/numericalearth/LocalPreferences.toml test/. 2>/dev/null || true
+      - name: Add CUDA_Runtime_jll to test environment
+        run: julia --project --color=yes --check-bounds=yes -e 'using Pkg; Pkg.Registry.update(); Pkg.add("CUDA_Runtime_jll")'
+      - name: Run tests
+        run: julia --project --color=yes --check-bounds=yes -e '
+          using Pkg;
+          Pkg.test(; coverage=true,
+            julia_args=["--check-bounds=yes", "--compiled-modules=yes", "-O0"],
+            test_args=["${{ matrix.test_group }}"])'
```
giordano: I believe this should be
```suggestion
            julia_args=["--check-bounds=yes", "--compiled-modules=yes", "-O0"]'
        env:
          TEST_GROUP: ${{ matrix.test_group }}
```

.github/workflows/ci.yml
```diff
@@ -159,19 +128,29 @@ jobs:
...
           - woa
     env:
+      GPU_TEST: "true"
       OPENBLAS_NUM_THREADS: 1
-      JULIA_NUM_PRECOMPILE_TASKS: 8
+      # Redirect temp and scratch to the workspace volume (bind-mounted from the
+      # host's EBS disk) to avoid filling the container's overlay filesystem with
+      # large data downloads and compiled artifacts.
+      TMPDIR: /__w/_temp
+      JULIA_SCRATCH_DIR: /__w/_scratch
```
giordano: What's `JULIA_SCRATCH_DIR`?  I don't think that's a variable actively used anywhere

awiteck:
@giordano mind taking another look here?

giordano (commented):
Looks overall good to me (minus investigating why the precompile cache is not hit at all, but can be done later at this point), but I'd prefer an approval from someone who follows this project better than me 🙂

giordano:
> Looks overall good to me (minus investigating why the precompile cache is not hit at all, but can be done later at this point)

I investigated and resolved all the issues (#129, https://github.com/NumericalEarth/numerical-earth-docker-images/commit/759448110c92708c8fe4ecec1690be47aafe45d0, https://github.com/NumericalEarth/numerical-earth-docker-images/commit/a25dc35a4e72ddfbb814a3ca9c70c47ecf976772), and now [the CI jobs here are only precompiling NumericalEarth](https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/23092567454/job/67079589654#step:7:617), as desired.

This is ready to go from my point of view, a follow up work should be to move tests to ParallelTestRunner.jl (#28) to use resources more efficiently.

navidcy:
> This is ready to go from my point of view, a follow up work should be to move tests to ParallelTestRunner.jl (#28) to use resources more efficiently.

I'm keen to approve based on this!

simone-silvestri (approved):
Looks good! Now thew bottleneck is the ecco_en4 test which takes 1 hour 20 minutes on CPU and 47 minutes on GPU. We can split it further in ECCO4 and EN4 tests to speed up the test suite.

PR #130, opened 2026-03-14
Require a much more recent version of Reactant

giordano:
I just realised we were requiring a year-old version of Reactant, which was causing problems with setting up the GPU runner in #127 (the preference set in the docker image wasn't compatible with such an old version of Reactant)

### DIFF
--- a/test/Project.toml
+++ b/test/Project.toml
@@ -31,7 +31,7 @@ NumericalEarth = {path = ".."}
 
 [compat]
 Breeze = "0.4"
-CDSAPI = "2"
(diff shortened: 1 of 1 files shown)


giordano:
> We were not requiring that version, that's just the lower bound, right?

It's "requiring at least", not "requiring exactly", yes. But the point is that an old version of Reactant is being used here in CI, and it wasn't clear to me why that was the case (that became clear after I opened the PR: it was a conflict with CDSAPI, hopefully solved now).

PR #135, opened 2026-03-16
Using ParallelTestRunner to (immensely [edit by navidcy]) speed up tests

### LINKED ISSUE #28: Should we spawn more agents for the tests?

simone-silvestri:
with @giordano


simone-silvestri:
> fix download of the GPU tests

If we separate the downloads we will effectively double the size of the cached data, is this ok?

giordano:
Oh, with 502abe8b61fab863679d7219b34ec095633c3edd I disabled caching of the scratch space anyway: each cache object was 16 GB, way larger than the 10 GB for the entire repo, so the cache was evicted as soon as it was saved. And caching files to be downloaded probably doesn't help much anyway: we have to download them in some way at some point in any case.

simone-silvestri:
let me see if we can reduce the number of files we download, we probably just need one for dataset to check the links are working.

giordano:
Still same errors :smiling_face_with_tear: ***Edit***: we eventually found the problem: the `test_downloading` test was removing/downloading some test files needed by other test files, so messing up with the other tests.  Moving `test_downloading` to a separate CI job solved the issue.

navidcy (approved):

test/test_ocean_only_model.jl
```diff
@@ -0,0 +1,62 @@
...
+        radiation = Radiation(arch)
+
+        # Fluxes are computed when the model is constructed, so we just test that this works.
+        @test begin
+            coupled_model = OceanOnlyModel(ocean; atmosphere, radiation)
+            time_step!(coupled_model, 1)
+            true
+        end
+    end
+end
```
navidcy:
```suggestion
end

```

giordano: :star_struck:

giordano:
It's all green, it's happening :eyes:

simone-silvestri:
Let's go!

PR #136, opened 2026-03-17
Remove various unused dependencies

giordano:
CUDA_Compiler_jll wasn't used anwywhere, and the stale compat bound was causing problems in downstream packages.

I guess I'll need to add Aqua testing at some point, to catch stale dependencies 😈


giordano:
Also the dependency on CUDA was entirely unnecessary: `@allowscalar` comes from `GPUArraysCore`, which is a lot smaller package.  I guess I'll also need to add testing with `ExplicitImports`, which checks we're importing symbols from the right modules 😈😈

giordano:
Before this PR:
```julia-repl
julia> @time @eval using NumericalEarth
  3.927013 seconds (5.74 M allocations: 388.925 MiB, 9.10% gc time, 11.32% compilation time: 82% of which was recompilation)
```
After this PR:
```julia-repl
julia> @time @eval using NumericalEarth
  2.503683 seconds (3.81 M allocations: 241.258 MiB, 12.72% gc time, 3.17% compilation time)
```
🙃

navidcy (approved):

navidcy:
Thanks @giordano!!

PR #140, opened 2026-03-24
Add ocean basin labeling and masking framework from ClimaOcean

simone-silvestri:
## Summary

ClimaOcean has developed a basin labeling and masking framework that enables identifying and isolating individual ocean basins (Atlantic, Pacific, etc.) from bathymetry data. This is needed for basin-specific diagnostics, regional restoring, and inpainting. Meanwhile, NumericalEarth had independently added a bathymetry caching system to avoid recomputing regridded bathymetry on every call.

This PR reconciles the two codebases by porting ClimaOcean's labeling advances into NumericalEarth while preserving the caching infrastructure, preparing for https://github.com/CliMA/ClimaOcean.jl/pull/757 which will have ClimaOcean depend on NumericalEarth's Bathymetry module for this functionality.

### Changes

- Add `Barrier` type for geographically separating connected ocean basins (e.g. closing Drake Passage to isolate the Atlantic)
- Add `label_ocean_basins` with support for periodic and tripolar grid boundary conditions
- Add `OceanBasinMask` with predefined barriers and seed points, plus convenience functions: `atlantic_ocean_mask`, `indian_ocean_mask`, `pacific_ocean_mask`, `southern_ocean_mask`, `arctic_ocean_mask`
- Refactor `remove_minor_basins!` to delegate to `label_ocean_basins` instead of doing its own internal labeling
- Preserve the existing bathymetry caching infrastructure (`BathymetryRegridding`, JLD2 cache)
- Add tests for barrier geometry, basin labeling with/without barriers, and `OceanBasinMask` creation

## Test plan

- [x] Existing bathymetry construction, smoothing, and caching tests still pass
- [x] `Barrier` constructors (explicit, keyword, meridional with width)
- [x] `label_ocean_basins` without barriers produces a single connected ocean
- [x] `label_ocean_basins` with barriers separates basins correctly
- [x] `OceanBasinMask` for the Atlantic excludes Pacific cells


src/Bathymetry/label_ocean_basins.jl
```diff
@@ -0,0 +1,226 @@
...
+    φ = φnode(i, j, 1, grid, Center(), Center(), Center())
+    in_lat = barrier.south <= φ <= barrier.north
+
+    @inbounds zb[i, j, 1] = ifelse(in_lon & in_lat, zero(grid), zb[i, j, 1])
+end
+
+# Since the strel algorithm in `remove_major_basins` does not recognize periodic boundaries,
+# before removing connected regions, we extend the longitude direction if it is periodic.
+# An extension of half the domain is enough.
+function maybe_extend_longitude(zb_cpu::Field, ::Periodic)
```
glwagner:
```suggestion
function copy_periodic_longitude(zb_cpu::Field, ::Periodic)
```

I think "extension" is not quite the right word

src/Bathymetry/label_ocean_basins.jl
```diff
@@ -0,0 +1,226 @@
...
+    Nx = size(zb_cpu, 1)
+    nx = Nx ÷ 2
+
+    zb_data   = zb_cpu.data[1:Nx, :, 1]
+    zb_parent = zb_data.parent
```
glwagner: does this work for very large datasets?

src/Bathymetry/label_ocean_basins.jl
```diff
@@ -0,0 +1,226 @@
...
+
+    @inbounds zb[i, j, 1] = ifelse(in_lon & in_lat, zero(grid), zb[i, j, 1])
+end
+
+# Since the strel algorithm in `remove_major_basins` does not recognize periodic boundaries,
+# before removing connected regions, we extend the longitude direction if it is periodic.
+# An extension of half the domain is enough.
+function maybe_extend_longitude(zb_cpu::Field, ::Periodic)
+    Nx = size(zb_cpu, 1)
+    nx = Nx ÷ 2
```
glwagner: this is used to mean "local Nx" elsewhere.

src/Bathymetry/ocean_basin_mask.jl
```diff
@@ -0,0 +1,375 @@
...
+##### OceanBasinMask struct
+#####
+
+"""
+    OceanBasinMask{M, G}
+
+A mask identifying cells belonging to a specific ocean basin.
+The mask is stored as a 2D Field{Center, Center, Nothing} with values:
+- 1.0: cell belongs to the basin
+- 0.0: cell does not belong to the basin
```
glwagner: the docstring does not mention what `seed_points` are

src/Bathymetry/ocean_basin_mask.jl
```diff
@@ -0,0 +1,375 @@
...
+"""
+struct OceanBasinMask{M, G, S}
```
glwagner:
```suggestion
struct BasinMask{M, G, S}
```

this has nothing to do with oceans, it is just geometry

src/Bathymetry/ocean_basin.jl
glwagner, line 1: consider using the word "Earth". We usually think of "oceans" as generic (Europa has an ocean)

src/Bathymetry/ocean_basin.jl
```diff
@@ -0,0 +1,375 @@
...
+#####
+##### Some useful Basin seeds and barriers
+#####
+
+const SOUTHERN_OCEAN_SEPARATION_BARRIER = Barrier(-180.0, 180.0, -56.0, -54.0)
+
+const ATLANTIC_OCEAN_BARRIERS = [
+    Barrier(20.0, -90.0, -30.0),      # Cape Agulhas (meridional barrier)
+    Barrier(289.0, -90.0, -30.0)      # Drake passage (meridional barrier)
+]
```
glwagner, line 136: why are these all Float64?

src/Bathymetry/ocean_basin_mask.jl
```diff
@@ -0,0 +1,375 @@
...
+add_barrier(::Nothing,         b::Barrier) = b
+
+"""
+    OceanBasinMask(grid;
+                   south_boundary = nothing,
+                   north_boundary = nothing,
+                   seed_points = [(0, 0)],
+                   barriers = nothing)
+
+Create a mask identifying cells in the Ocean basin using connected component analysis.
```
glwagner: what is "connected component analysis"?

src/Bathymetry/ocean_basin_mask.jl
```diff
@@ -0,0 +1,375 @@
...
+The algorithm uses labels restricted by south, north, east, and west boundaries with a fill
+algorithm that retrieves the basin corresponding to labels associated with known `seed_points`.
```
glwagner:
```suggestion
algorithm that retrieves the basin corresponding to labels associated with known `seed_points` and bounded by `barriers`
```

?

why are `seed_points` and `barriers` plural? It is not made clear in the docstring how multiple seed points and barriers interact with each other.

src/Bathymetry/label_ocean_basins.jl
```diff
@@ -0,0 +1,226 @@
...
+============
+- `Barrier(west, east, south, north)`: Create a barrier with explicit bounds
+- `Barrier(; west, east, south, north)`: Keyword argument version
+"""
+struct Barrier{W, E, S, N}
+    west  :: W
+    east  :: E
+    south :: S
+    north :: N
+end
```
glwagner: what is the conceptual difference between `Barrier` and `BoundingBox` ?

glwagner (changes requested):

src/Bathymetry/ocean_basin_mask.jl
glwagner: also why is the mask floats rather than bool?

src/Bathymetry/ocean_basin_mask.jl
glwagner: and also, shouldn't the mask be `Bool`, not float?

glwagner:
Can these tools also be used to identify lakes, or any enclosed body? If so "ocean basin" may not be the right word.

simone-silvestri:
thanks for the review, I ll go through the comments one by one.

simone-silvestri:
I think all the review comments have been addressed:

##  Changes made:

  1. Barrier vs BoundingBox — Added a Note: section to the Barrier docstring explaining the conceptual difference (Barrier blocks water cells for labeling; BoundingBox selects spatial
  data subsets)
  2. maybe_extend_longitude → copy_periodic_longitude — Renamed as suggested
  3. Nx → Nλ, nx → half — Renamed to avoid confusion with grid's Nx
  4. Large dataset concern — Added comment: "This is an O(Nλ × Nφ) CPU allocation, acceptable for the serial labeling step"
  5. OceanBasinMask → BasinMask — Renamed throughout (struct, exports, tests, docstrings)
  6. "Earth" naming — Updated convenience function docstrings to say "Earth's Atlantic Ocean" etc., and section header to "Convenience functions for Earth's ocean basins"
  7. Mask Bool instead of float — Changed all Field{Center, Center, Nothing}(grid) mask allocations to Field{Center, Center, Nothing}(grid, Bool); updated + operator to use .| instead
  of max.
  8. Docstring for seed_points — Added explanation to the BasinMask struct docstring and expanded the constructor docstring
  9. "connected component analysis" clarified — Rewrote the constructor docstring to reference ImageMorphology.label_components explicitly and describe the algorithm plainly
  10. Seed points / barriers interaction clarified — Added explanation of how multiple seeds (tried in order, first hit wins) and multiple barriers (applied simultaneously) work
  11. southern_ocean_mask alignment — Fixed indentation of keyword args to align with the function name

glwagner:
Sorry to be so annoying about `Barrier` -- but can we simply use `BoundingBox` for the same purpose as barrier? They seem to carry the exact same information and _sound_ very similar. I understand they are used for different purposes in different contexts, but that doesn't matter. (It would only matter if we have to support both barrier and bounding box at the same time, then we need them distinct)

src/Bathymetry/label_ocean_basins.jl
```diff
@@ -0,0 +1,232 @@
...
+using Oceananigans.Fields: convert_to_0_360
+
+#####
+##### Barrier type for separating connected regions
+#####
+
+"""
+    Barrier{W, E, S, N}
+
+A rectangular geographic region used to separate connected water regions during labeling.
```
glwagner: bounding box literally represents a rectangular geographic region

src/Bathymetry/label_ocean_basins.jl
```diff
@@ -0,0 +1,232 @@
...
+
+# Since the strel algorithm in `label_components` does not recognize periodic boundaries,
+# before labeling connected regions we copy half the longitude extent on each side so that
+# water cells near the boundary are correctly identified as connected.
+function copy_periodic_longitude(zb_cpu::Field, ::Periodic)
+    Nλ = size(zb_cpu, 1)
+    half = Nλ ÷ 2
+
+    zb_data   = zb_cpu.data[1:Nλ, :, 1]
+    zb_parent = zb_data.parent
```
glwagner: this reads weird. what is the type of `zb_data` here? should we be using `view(zb_cpu, 1:N, :, 1)`? I thought that the `getindex` would allocate a new array. 

either way, def better to use `parent(zb_data)` to be more future proof.

src/Bathymetry/label_ocean_basins.jl
```diff
@@ -0,0 +1,232 @@
...
+
+    # Concatenate a copy of the eastern half on the left and the western half on the right.
+    # This is an O(Nλ × Nφ) CPU allocation, acceptable for the serial labeling step.
+    zb_parent = vcat(zb_parent[half:Nλ, :], zb_parent, zb_parent[1:half, :])
```
glwagner: @giordano has showed cases where rebinding a name (eg reusing the name `zb_parent`) produces compiler problems, so you probably need to introduce a new name like `concatenated_zb`

src/Bathymetry/label_ocean_basins.jl
```diff
@@ -0,0 +1,232 @@
...
+
+    # Update offsets so that index 1 maps back to the original domain start
+    yoffsets = zb_cpu.data.offsets[2]
+    xoffsets = - half - 1
```
glwagner:
```suggestion
    zb_data   = zb_cpu.data[1:Nλ, :, 1]
    zb_parent = zb_data.parent

    # Concatenate a copy of the eastern half on the left and the western half on the right.
    # This is an O(Nλ × Nφ) CPU allocation, acceptable for the serial labeling step.
    half = Nλ ÷ 2
    zb_parent = vcat(zb_parent[half:Nλ, :], zb_parent, zb_parent[1:half, :])

    # Update offsets so that index 1 maps back to the original domain start
    yoffsets = zb_cpu.data.offsets[2]
    xoffsets = - half - 1
```

easier to read when variables are defined just before being used

src/Bathymetry/label_ocean_basins.jl
```diff
@@ -0,0 +1,232 @@
...
+    # The labelling algorithm works only on CPUs
+    cpu_grid = on_architecture(CPU(), grid)
+
+    TX = topology(cpu_grid, 1)
+    zb = cpu_grid.immersed_boundary.bottom_height
+
+    # If barriers are specified, apply them to a copy of the bathymetry
+    if !isnothing(barriers)
+        # Create a temporary field with the modified bathymetry
+        zb_modified = Field{Center, Center, Nothing}(cpu_grid)
```
glwagner: could probably also use `similar(zb)` -- may be more robust to different bathymetry types

src/Bathymetry/label_ocean_basins.jl
```diff
@@ -0,0 +1,232 @@
...
+              mark certain cells as land, allowing separation of connected ocean basins
+              (e.g., separating Atlantic from Pacific via the Southern Ocean).
+"""
+function label_ocean_basins(grid::ImmersedBoundaryGrid; barriers=nothing)
+
+    # The labelling algorithm works only on CPUs
+    cpu_grid = on_architecture(CPU(), grid)
+
+    TX = topology(cpu_grid, 1)
+    zb = cpu_grid.immersed_boundary.bottom_height
```
glwagner: this is specific to `GridFittedBottom` and `PartialCellBottom`. perhaps it will work when we have "top" partial cells too and cut cells?

src/Bathymetry/ocean_basin_mask.jl
```diff
@@ -0,0 +1,388 @@
...
+
+The mask is stored as a 2D `Field{Center, Center, Nothing}` with `Bool` values:
+- `true`:  cell belongs to the basin
+- `false`: cell does not belong to the basin
+
+`seed_points` is a collection of `(λ, φ)` coordinate pairs used to identify which
+connected component corresponds to the desired basin. The algorithm finds the
+label at each seed point and builds the mask from that label.
+"""
+struct BasinMask{M, G, S}
```
glwagner: I think technically this type represents a basin. It is not the mask itself -- the mask is

```
basin.mask
```

if you call this "mask" you have confusing semantics:

```
mask = BasinMask()
actual_mask = mask.mask
```

maybe clear this up?

src/Bathymetry/ocean_basin_mask.jl
```diff
@@ -0,0 +1,388 @@
...
+##### Additive basin masks
+#####
+
+"""
+    Base.:+(mask1::BasinMask, mask2::BasinMask)
+
+Combine two basin masks, returning a new mask that is the union of both.
+Cells that belong to either basin will be `true`.
+"""
+function Base.:+(mask1::BasinMask, mask2::BasinMask)
```
glwagner: should it be `*`?

src/Bathymetry/ocean_basin.jl
```diff
@@ -0,0 +1,388 @@
...
```
glwagner, line 136: why all of these have `.0` eg are float64?

src/Bathymetry/ocean_basin_mask.jl
```diff
@@ -0,0 +1,388 @@
...
+- `include_southern_ocean`: If `true`, extends the Atlantic basin into the Southern Ocean
+                            sector below the standard separation latitude (~55°S). Default: `false`.
+- `south_boundary`: Southern latitude limit. Default: -50.0 (or -90.0 if `include_southern_ocean=true`)
+- `north_boundary`: Northern latitude limit. Default: 65.0
+- Other keyword arguments are passed to `BasinMask`.
+"""
+function atlantic_ocean_mask(grid;
+                             include_southern_ocean = true,
+                             south_boundary = include_southern_ocean ? -90.0 : -50.0,
+                             north_boundary = 65.0,
```
glwagner: f64?

src/Bathymetry/ocean_basin_mask.jl
```diff
@@ -0,0 +1,388 @@
...
+- `include_southern_ocean`: If `true`, extends the Indian basin into the Southern Ocean
+                            sector below the standard separation latitude (~55°S). Default: `false`.
+- `south_boundary`: Southern latitude limit. Default: -50.0 (or -90.0 if `include_southern_ocean=true`)
+- `north_boundary`: Northern latitude limit. Default: 30.0
+- Other keyword arguments are passed to `BasinMask`.
+"""
+function indian_ocean_mask(grid;
+                           include_southern_ocean = true,
+                           south_boundary = include_southern_ocean ? -90.0 : -50.0,
+                           north_boundary = 30.0,
```
glwagner: f64?

test/test_bathymetry.jl
```diff
@@ -140,3 +146,108 @@ end
...
+    @test meridional_wide.west == 18.0
+    @test meridional_wide.east == 22.0
+
+    # Test LatitudeBand
+    band = Barrier(-180.0, 180.0, -60.0, -55.0)
+    @test band.west == -180.0
+    @test band.east == 180.0
+    @test band.south == -60.0
+    @test band.north == -55.0
+end
```
glwagner, line 177: better to compare to int here?

glwagner (commented):
I know I'm being annoying, but it's probably nice to get this right...

1. Can we use `BoundingBox` to represnet barriers?
2. `BasinMask` may be wrong name, you may want "Basin" so the mask is `mask = basin.mask`
3. There's a lot of F64 baked in, it looks dirty to me but maybe has a purpose?
4. Also lots of broadcasting and CPU assumption which has caused issues in the past. Do we need this to work on GPU ever? Perhaps not, but if we have like 40m bathymetry global, would it fail? I'm also ok to merge as is and work on clean up later, if it works.
5. I think there is also an assumption about the type of immersed boundary grid --- the main thing I think we want to consider is supporting ice shelves / top partial cells, which hopefully will come soon.

src/Bathymetry/ocean_basin_mask.jl
simone-silvestri: It actually should be an `or`. So I have changed it to `|`

simone-silvestri:
I should have answered all the comments. I think this is ready to review again.

glwagner:
> I should have answered all the comments. I think this is ready to review again.

The comments may have been addressed in the code but are unanswered in text. Can you please respond to my comments and say what was done to address them?

simone-silvestri:
@glwagner, sorry took me a while.

1. yes (done)
2. done
3. Removed in favor of `Int` values
4. Yeah it's a little difficult for the moment to make labelling work with GPUs since it uses a non-GPU ready package. Maybe we can see how to fix it in a coming PR
5. I am not sure how to consider a "basin" something that has a top boundary grid. However I think this PR sets the pluming and the interface, then, when we encounter this problem we can extend this interface to include other types of "area chunking"

src/Bathymetry/regrid_bathymetry.jl
```diff
@@ -516,22 +475,26 @@ function remove_minor_basins!(zb, keep_major_basins, core_size)
...
+    zb_data = interior(zb_cpu, :, :, 1)
 
-    for ℓ = 1:maximum(labels)
-        remove_basin = all(ℓ != m for m in mm_basins)
-        if remove_basin
-            labels[labels .== ℓ] .= NaN # Regions to remove
+    for j in 1:Ny, i in 1:Nx
+        label = labels[i, j]
+        if label > 0 && !(label in major_basins)
+            zb_data[i, j] = 0  # Flatten this cell (make it land)
```
glwagner:
```suggestion
            @inbounds zb_data[i, j] = 0  # Flatten this cell (make it land)
```

why is this restricted to CPU?

simone-silvestri: the problem is that the labelling process is CPU only

PR #144, opened 2026-03-25
Update ORCAGrid for new RightFaceFolded convention

simone-silvestri:
## Summary

Following [Oceananigans PR #5408](https://github.com/CliMA/Oceananigans.jl/pull/5408), the `RightFaceFolded` topology now behaves like `Bounded`:
- Center fields have `Ny` interior points
- Face fields have `Ny + 1` interior points (fold at `j = Ny+1`)

This removes the old workaround of adding 1 to `Ny` when constructing the ORCA grid.

**Changes:**
- `Ny = Ny_nemo` directly (was `Ny_nemo + 1`)
- Simplified code: `halo_fill_stagger` helper, `continue_south!` loop, `to_arch` helper
- Updated test expectations for new `Ny` values (333 → 332)

## Test plan
- [ ] ORCA grid construction tests pass with new Ny values
- [ ] Metric consistency tests (no NaN/Inf, positive metrics, stagger checks)
- [ ] Bathymetry loading unchanged


src/Bathymetry/orca_grid.jl
```diff
@@ -41,41 +41,50 @@ end
...
+    return (
+        halo_filled_data(CC, helper_grid, bcs, Center, Center, overlap),
+        halo_filled_data(FC, helper_grid, bcs, Face,   Center, overlap),
+        halo_filled_data(CF, helper_grid, bcs, Center, Face,   overlap),
+        halo_filled_data(FF, helper_grid, bcs, Face,   Face,   overlap),
+    )
+end
+
 """
     ORCAGrid(arch = CPU(), FT::DataType = Float64;
```
glwagner, line 89: This isn't a valid constructor because there is no type or type alias called `ORCAGrid`. You should either call it `orca_grid`, or you can define a type alias (I believe this is possible, if you introduce `ORCAConformalMap`).

I also suggest `ORCATripolarGrid` (this is a tripolar grid, right?)

glwagner (commented):
define a type alias `ORCATripolarGrid` (and change the name to indicate it is a tripolar grid)

simone-silvestri:
I have added these changes in #145 to update to the new Oceananigans version, we can keep this PR open, merge #145 then I will address the comments here

src/Bathymetry/orca_grid.jl
simone-silvestri: for a new conformal map we probably should introduce an abstract `AbstractTripolarMap` so that we can define a tripolar grid based on an abstract type rather than a concrete one

simone-silvestri:
Will close this because #153 will already refactor the ORCA grid.

Issue #30, opened 2026-02-16
Add radiation as a top-level EarthSystemModel component


glwagner:
## Motivation

Currently, radiation-related state is split across two separate types:

- **`Radiation`** (in `InterfaceComputations/radiation.jl`): holds surface radiative properties — albedo, emissivity, and the Stefan-Boltzmann constant. Lives inside `ComponentInterfaces`, not as a first-class model component.
- **`PrescribedAtmosphere`** (in `Atmospheres/prescribed_atmosphere.jl`): holds `downwelling_radiation` (a `TwoBandDownwellingRadiation` with shortwave/longwave fields), even though downwelling radiative fluxes are fundamentally a radiation concern, not an atmospheric state variable.

This split makes it difficult to swap in an interactive radiation scheme (e.g. RRTMGP) because the relevant information is scattered across unrelated types. An interactive radiation component would need to provide the same outputs (downwelling fluxes, surface albedo, emissivity, etc.) but there is currently no single abstraction that represents "radiation" as a complete component.

## Proposal

Add a `radiation` field to `EarthSystemModel`:

```julia
mutable struct EarthSystemModel{I, A, O, R, F, C, Arch} <: AbstractModel{Nothing, Arch}
    architecture :: Arch
    clock :: C
    atmosphere :: A
    radiation :: R       # new
    sea_ice :: I
    ocean :: O
    interfaces :: F
end
```

Introduce a `PrescribedRadiation` type that consolidates everything radiative into one place:

```julia
struct PrescribedRadiation
    downwelling_radiation   # TwoBandDownwellingRadiation (shortwave + longwave fields)
    emission                # SurfaceProperties (ocean/sea-ice emissivity)
    reflection              # SurfaceProperties (ocean/sea-ice albedo)
    stefan_boltzmann_constant
end
```

This merges:
- The downwelling flux data currently on `PrescribedAtmosphere.downwelling_radiation`
- The surface radiative properties currently on `Radiation` (emission, reflection, Stefan-Boltzmann constant)

### What changes

1. **`PrescribedAtmosphere`** loses its `downwelling_radiation` field. It becomes purely an atmospheric state provider (velocities, pressure, tracers, freshwater flux, thermodynamics).
2. **`Radiation`** (the current interface-computation type) is replaced by `PrescribedRadiation`, which lives on the model rather than buried inside `ComponentInterfaces`.
[...]

PR #149, opened 2026-03-30
(0.3.0) Add land as a top-level EarthSystemModel component

### LINKED ISSUE #36: Add land as a top-level EarthSystemModel component
## Motivation

Similar to the radiation refactor (#30), land-related data is currently shoehorned into `PrescribedAtmosphere`. Specifically:

- **`freshwater_flux`** (rain, snow) — precipitation that originates in the atmosphere but ultimately represents a surface/land boundary flux
- **`auxiliary_freshwater_flux`** (rivers, icebergs) — these are explicitly land and land-ice sourced. In JRA55 they are even categorized under `"land/day"` and `"landIce/day"` rather than `"atmos/3hr"`

There is currently no concept of "land" in `EarthSystemModel`. River runoff, iceberg calving, and potentially other land-surface processes (soil moisture, land surface temperature, vegetation effects) have no natural home. They end up as auxiliary fields on the atmosphere, handled with special-case code (e.g. `auxiliary_freshwater_flux` is interpolated separately because it lives on a different grid and time frequency than the atmospheric fields).

Adding a land component — starting with `PrescribedLand` — would:

1. Give land-sourced fluxes a proper home
2. Clean up `PrescribedAtmosphere` so it truly represents atmospheric state
3. Provide a path to interactive land models in the future

## Proposal

Add a `land` field to `EarthSystemModel`:

```julia
mutable struct EarthSystemModel{I, A, O, R, L, F, C, Arch} <: AbstractModel{Nothing, Arch}
    architecture :: Arch
    clock :: C
    atmosphere :: A
    radiation :: R
    land :: L            # new
    sea_ice :: I
    ocean :: O
    interfaces :: F
end
```

Introduce a `PrescribedLand` type that holds land-sourced boundary data:

```julia
struct PrescribedLand
    river_runoff        # freshwater flux from rivers
    iceberg_calving     # freshwater flux from calving icebergs
    # Future: land surface temperature, soil moisture, ...
end
```

### What changes

1. **`PrescribedAtmosphere`** loses `auxiliary_freshwater_flux` (and possibly `freshwater_flux`, depending on how we classify precipitation). It becomes purely atmospheric state: wind, temperature, humidity, pressure.
2. **`JRA55PrescribedAtmosphere`** is complemented by a **`JRA55PrescribedLand`** constructor that loads the `friver` and `licalvf` fields (and potentially rain/snow if we decide precipitation belongs to the land surface budget).
3. Similarly, a **`JRA55PrescribedRadiation`** (from #30) would handle shortwave/longwave.
4. **Interface computations** pull land-sourced freshwater from `earth_system_model.land` instead of from `atmosphere.auxiliary_freshwater_flux`.

### Relationship to #30

This is the land-side analog of the radiation refactor in #30. Together, the two refactors decompose `PrescribedAtmosphere` into focused components:

| Component | What it provides |
|-----------|-----------------|
| `PrescribedAtmosphere` | Wind, temperature, humidity, pressure |
| `PrescribedRadiation` | Downwelling shortwave/longwave, surface albedo, emissivity |
| `PrescribedLand` | River runoff, iceberg calving, (future: land surface fields) |

Each would have a `JRA55`-prefixed constructor (`JRA55PrescribedAtmosphere`, `JRA55PrescribedRadiation`, `JRA55PrescribedLand`) for loading reanalysis data.

### Path to interactive land

With land as a first-class component, swapping in an interactive land surface model becomes straightforward — it would satisfy the same interface (providing freshwater fluxes to the ocean, and potentially exchanging heat/moisture with the atmosphere) but compute them prognostically rather than reading them from data.

## Current state of land-related data

| Field | Currently lives on | JRA55 category | Grid/frequency |
|-------|-------------------|----------------|----------------|
| `river_freshwater_flux` | `PrescribedAtmosphere.auxiliary_freshwater_flux` | `land/day` | daily |
| `iceberg_freshwater_flux` | `PrescribedAtmosphere.auxiliary_freshwater_flux` | `landIce/day` | daily |

glwagner:
## Summary

Closes #36.

- Adds a new `PrescribedLand` component (`src/Lands/`) that holds river runoff and iceberg calving freshwater fluxes, previously shoehorned into `PrescribedAtmosphere.auxiliary_freshwater_flux`
- `EarthSystemModel` gains a `land` field; all constructors (`OceanOnlyModel`, `OceanSeaIceModel`, `AtmosphereOceanModel`) accept a `land` keyword
- `JRA55PrescribedLand` constructor loads `:river_freshwater_flux` and `:iceberg_freshwater_flux` via `JRA55FieldTimeSeries`
- Ocean flux assembly sums `atmosphere.freshwater_flux + land.freshwater_flux` (land defaults to zero when absent)
- Removes `auxiliary_freshwater_flux` from `PrescribedAtmosphere` and `include_rivers_and_icebergs` from `JRA55PrescribedAtmosphere`
- Updates checkpointing with backward-compatible restore (`hasproperty` check for old checkpoints without land)

## Test plan

- [x] `test_jra55.jl`: `JRA55PrescribedLand` loads river/iceberg data with correct grid and daily frequency
- [x] `test_ocean_only_model.jl`: `OceanOnlyModel` constructs and time-steps with land
- [x] `test_ocean_sea_ice_model.jl`: `OceanSeaIceModel` constructs and time-steps with land
- [x] `test_checkpointer.jl`: round-trip checkpoint with land component
- [x] `test_surface_fluxes.jl`: model constructs with and without land, exchanger wired correctly
- [x] All existing tests pass with `land=nothing` (default)


examples/single_column_os_papa_simulation.jl
```diff
@@ -306,7 +306,7 @@ scatterlines!(axκz, κn)
 
 axislegend(axuz)
 
-ulim = max(maximum(abs, u), maximum(abs, v))
+ulim = max(maximum(abs, u), maximum(abs, v), 1e-10)
```
bgroenks96, line 316: what is `1e-10`? comment or use an explicit variable.

src/EarthSystemModels/InterfaceComputations/state_exchanger.jl
```diff
@@ -3,26 +3,30 @@ struct ComponentExchanger{S, EX}
...
         S = typeof(sea_ice_exchanger)
-        
-        return new{G, A, O, S}(grid, 
-                               atmosphere_exchanger, 
-                               ocean_exchanger, 
-                               sea_ice_exchanger)
+
+        return new{G, A, L, O, S}(grid,
+                                   atmosphere_exchanger,
+                                   land_exchanger,
```
bgroenks96: What is a land exchanger? Does it exchange only with the atmosphere? Or all other components?

src/EarthSystemModels/earth_system_model.jl
```diff
@@ -333,6 +345,8 @@ function restore_prognostic_state!(osm::EarthSystemModel, state)
     restore_prognostic_state!(osm.clock, state.clock)
     restore_prognostic_state!(osm.ocean, state.ocean)
     restore_prognostic_state!(osm.atmosphere, state.atmosphere)
+    land_state = hasproperty(state, :land) ? state.land : nothing
```
bgroenks96: Why is this only necessary for land?

src/Lands/Lands.jl
```diff
@@ -0,0 +1,27 @@
+module Lands
```
bgroenks96, line 1: I don't love this but I get the desire for consistency.

src/Lands/prescribed_land.jl
```diff
@@ -0,0 +1,73 @@
+mutable struct PrescribedLand{G, T, F, TI}
+    grid :: G
+    clock :: Clock{T}
+    freshwater_flux :: F   # NamedTuple, e.g. (rivers=FTS, icebergs=FTS)
```
bgroenks96, line 9: is this really the only thing needed? what about albedo? skin temperature?

src/Lands/prescribed_land.jl
```diff
@@ -0,0 +1,73 @@
...
+    print(io, "└── freshwater_flux: ", flux_names)
+end
+
+"""
+    PrescribedLand(freshwater_flux; clock=Clock{Float64}(time=0))
+
+Return a `PrescribedLand` component from a `NamedTuple` of `FieldTimeSeries`
+representing freshwater fluxes (e.g. rivers, icebergs).
+"""
+function PrescribedLand(freshwater_flux; clock=Clock{Float64}(time=0))
```
bgroenks96: shouldn't the default type of `Clock` be set to match that of `freshwater_flux`?

src/Oceans/assemble_net_ocean_fluxes.jl
```diff
@@ -45,6 +45,10 @@ function update_net_ocean_fluxes!(coupled_model, ocean_model, grid)
 
     freshwater_flux = atmosphere_fields.Jᶜ.data
 
+    # Extract land freshwater flux if land component is present
+    land_exchanger = coupled_model.interfaces.exchanger.land
+    land_freshwater_flux = isnothing(land_exchanger) ? nothing : land_exchanger.state.freshwater_flux.data
```
bgroenks96, line 48: why `.data`?

src/Oceans/assemble_net_ocean_fluxes.jl
```diff
@@ -66,12 +70,16 @@ function update_net_ocean_fluxes!(coupled_model, ocean_model, grid)
...
             freshwater_flux,
+            land_freshwater_flux,
             atmos_ocean_properties,
             ocean_properties)
 
     return nothing
 end
 
+@inline get_land_freshwater_flux(i, j, ::Nothing) = 0
+@inline get_land_freshwater_flux(i, j, flux) = @inbounds flux[i, j, 1]
```
bgroenks96: `@propagate_inbounds`?

test/test_checkpointer.jl
```diff
@@ -26,11 +26,12 @@ using Oceananigans.OutputWriters: Checkpointer
...
             set!(sea_ice.model, h=hi, ℵ=hi)
 
-            # Create atmosphere and radiation
+            # Create atmosphere, land, and radiation
             backend = JRA55NetCDFBackend(4)
             atmosphere = JRA55PrescribedAtmosphere(arch; backend)
+            land = JRA55PrescribedLand(arch; backend)
 
-            return OceanSeaIceModel(ocean, sea_ice; atmosphere)
+            return OceanSeaIceModel(ocean, sea_ice; atmosphere, land)
```
bgroenks96, line 34: Wait, why does an `OceanSeaIceModel` need `land`?

test/test_ocean_sea_ice_model.jl
```diff
@@ -68,6 +68,22 @@ using ClimaSeaIce.Rheologies
...
                 true
             end
+
+            # Test with land component
+            @info "Testing OceanSeaIceModel with land on $A..."
+            land = JRA55PrescribedLand(arch; backend)
+
+            @test begin
+                ocean_land = ocean_simulation(grid; free_surface)
+                set!(ocean_land.model, T=temperature_metadata[1], S=salinity_metadata[1])
```
bgroenks96: Is `T` here just ocean temperature?

test/test_surface_fluxes.jl
```diff
@@ -176,6 +176,44 @@ end
...
+        @test !isnothing(model_with_land.interfaces.exchanger.land)
+        @test model_with_land.land === land
+
+        # Test PrescribedLand display methods
+        @test summary(land) isa String
+        @test contains(sprint(show, land), "PrescribedLand")
+        @test contains(sprint(show, land), "freshwater_flux")
+
+        # Time-step without land exercises get_land_freshwater_flux(::Nothing) path
+        time_step!(model_no_land, 1)
```
bgroenks96: Is there some kind of condition you could test here? Maybe check the clock?

bgroenks96 (commented):
Looks great! 🎉

test/test_ocean_sea_ice_model.jl
glwagner: yes because perhaps confusingly `ocean_land` is an ocean simulation (eg `ocean_land = ocean_simulation(grid; free_surface)`

test/test_checkpointer.jl
glwagner: the name refers to the prognostic components; eg `OceanSeaIceModel` has prognostic ocean and prognostic sea ice. The other components, such as the atmosphere and land, are prescribed. land and ocean are coupled through river runoff primarily, although reanalysis products also include icebergs/calving as a land-ocean flux.

src/Oceans/assemble_net_ocean_fluxes.jl
glwagner: its passed into a kernel which, if `Field` would adapt to its `data` anyways. We could unwrap explicitly here or implicitly during argument conversion for the kernel... not sure whether one is better than the other

src/Lands/prescribed_land.jl
glwagner: I think you're right that we could conceive of atmosphere-land coupling with prescribed land and in that case the land albedo would be required.

Worth noting that we would like to move to a config where radiation is an independent component, #30. In that case surface albedos may belong to the radiation component. But this isn't settled (or I can't remember)

src/Lands/Lands.jl
glwagner: what do you propose?

src/EarthSystemModels/earth_system_model.jl
glwagner: Meh this is a claudism. It should be symmetric with other components

src/EarthSystemModels/InterfaceComputations/state_exchanger.jl
glwagner: a land exchanger regrids from the land to the exchange grid. For atmos-land coupling it is typically trivial.

examples/single_column_os_papa_simulation.jl
glwagner: hmm, this sometimes comes in because a `ulim=0` causes CairoMakie to fail since `colorrange=(0, 0)` is invalid

test/test_surface_fluxes.jl
glwagner: thats a good idea. we should have a function `time(model_no_land)` that can be interrogated

src/Oceans/assemble_net_ocean_fluxes.jl
bgroenks96: If it makes no difference, I would leave it as a `Field` and let it get unpacked implicitly. It's a small reduction in cognitive load for people less familiar with the code, but a reduction nonetheless.

src/Lands/Lands.jl
bgroenks96: I think I would prefer `LandModels` or `LandComponents`, but then you would need to change others to match it. It's not really that big of a deal, though.

examples/single_column_os_papa_simulation.jl
glwagner: added a comment

src/EarthSystemModels/InterfaceComputations/state_exchanger.jl
glwagner: added docstrings to explain this

src/EarthSystemModels/earth_system_model.jl
glwagner: should be fixed

src/Lands/Lands.jl
glwagner: I'm keeping "Lands".

src/Lands/prescribed_land.jl
glwagner: done

src/Lands/prescribed_land.jl
glwagner: put a comment, this can't be addressed in this PR

src/Oceans/assemble_net_ocean_fluxes.jl
glwagner: I'm leaving this since its a common pattern in NumericalEarth right now. I think we can remove all of the manually unwrapping in a future PR. I think we implemented these to solve some issues with compilation at various times, but it is worth revisiting.

src/Oceans/assemble_net_ocean_fluxes.jl
glwagner: `inbounds` is ubiquitous right now. It is possible to undertake a campaign to remove these and use `@propagate_inbounds`. But not for this PR

test/test_ocean_sea_ice_model.jl
glwagner: changed the name to `ocean_with_land`

test/test_surface_fluxes.jl
glwagner: added a condition that the time was advanced correctly

glwagner:
@giordano do you know why CI is flaky?

src/EarthSystemModels/InterfaceComputations/state_exchanger.jl
ewquon: I think situations where the atmospheric grid is _finer_ than the land grid, which I expect to be ~O(100 m) — O(1 km) depending on the resolution of available land-surface datasets, would be trivial. I see an interpolator implemented but I think we should think about special treatment for the opposite situation with a mesoscale atmospheric grid _coarser_ than the land surface grid, e.g., through tiling (sub-grid fractional approach).

giordano:
I think it's a github issue, because https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/25067674708 just says "The operation was canceled."

src/EarthSystemModels/InterfaceComputations/state_exchanger.jl
```diff
@@ -1,28 +1,46 @@
+"""
+    ComponentExchanger(component, exchange_grid)
+
+Holds a regridder and a buffer of `state` fields used to bring data from a
+component (atmosphere, land, ocean, sea ice) onto a shared `exchange_grid`,
+where atmosphere--ocean and atmosphere--sea-ice fluxes are computed.
+For atmosphere--land coupling the regridder is typically trivial.
```
kaiyuan-cheng: For NWP, yes, but it may not be the case for applications on longer time scales.

glwagner: ok, I will remove this comment. This is not intended to be oriented towards NWP.

glwagner: I think this comment was added in response to a previous question about what the atmosphere-land regridded would be.

glwagner:
```suggestion
```

kaiyuan-cheng (approved):
looks good

ewquon (approved):
Looks good to me.

src/EarthSystemModels/earth_system_model.jl
```diff
@@ -39,6 +40,7 @@ function Base.show(io::IO, cm::ESM)
 
     print(io, "├── ocean: ", ocean_summary, "\n")
     print(io, "├── atmosphere: ", summary(cm.atmosphere), "\n")
+    print(io, "├── land: ", summary(cm.land), "\n")
```
bgroenks96, line 42: Hmm, this is admittedly very nitpicky but shouldn't this print the fields in the same order that they are listed in the `struct`?

src/EarthSystemModels/earth_system_model.jl
```diff
@@ -279,7 +289,8 @@ Construct a coupled atmosphere--ocean model.
 Convenience constructor for [`EarthSystemModel`](@ref) with an atmosphere and ocean
 but no sea ice. All keyword arguments are forwarded to `EarthSystemModel`.
 """
-AtmosphereOceanModel(atmosphere, ocean; kw...) = EarthSystemModel(atmosphere, ocean, nothing; kw...)
+AtmosphereOceanModel(atmosphere, ocean; land=nothing, kw...) =
```
bgroenks96, line 292: Can we already add `AtmosphereLandModel` and `LandOnlyModel`? Or is it too soon? Maybe they are already there and GitHub just doesn't show everything.

src/EarthSystemModels/InterfaceComputations/state_exchanger.jl
```diff
@@ -1,28 +1,45 @@
...
-        return new{G, A, O, S}(grid, 
-                               atmosphere_exchanger, 
-                               ocean_exchanger, 
-                               sea_ice_exchanger)
+
+        return new{G, A, L, O, S}(grid,
+                                   atmosphere_exchanger,
+                                   land_exchanger,
+                                   ocean_exchanger,
+                                   sea_ice_exchanger)
```
glwagner:
```suggestion
                                  atmosphere_exchanger,
                                  land_exchanger,
                                  ocean_exchanger,
                                  sea_ice_exchanger)
```

test/test_ocean_sea_ice_model.jl
```diff
@@ -68,6 +68,22 @@ using ClimaSeaIce.Rheologies
                 time_step!(coupled_model, 1)
                 true
             end
+
+            # Test with land component
+            @info "Testing OceanSeaIceModel with land on $A..."
+            land = JRA55PrescribedLand(arch; backend)
+
+            @test begin
+                ocean_with_land = ocean_simulation(grid; free_surface)
```
bgroenks96, line 77: why does `ocean_simulation` return an `ocean_with_land`?

src/EarthSystemModels/InterfaceComputations/state_exchanger.jl
glwagner: For a subgrid fractional approach, I think that we do not regrid explicitly. Instead, we compute fluxes for each subgrid component based on the same atmospheric state within the same cell, and then net fluxes for the atmosphere accordingly based on the partitioning. This is also what we do for sea ice and ocean which supports an area fraction.

test/test_surface_fluxes.jl
```diff
@@ -176,6 +176,46 @@ end
...
+        ocean_no_land = ocean_simulation(grid;
+                                         momentum_advection = nothing,
+                                         tracer_advection = nothing,
+                                         closure = nothing,
+                                         bottom_drag_coefficient = 0)
+
+        set!(ocean_no_land.model, T = 15, S = 30)
+        model_no_land = OceanOnlyModel(ocean_no_land; atmosphere)
+
+        ocean_with_land = ocean_simulation(grid;
```
bgroenks96, line 191: Again here, I don't understand the difference between these two constructors and why one is with land and one without.

src/Lands/interpolate_land_state.jl
```diff
@@ -0,0 +1,65 @@
...
+end
+
+@kernel function _interpolate_land_freshwater_flux!(land_freshwater_flux,
+                                                     interface_grid,
+                                                     clock,
+                                                     freshwater_data,
+                                                     land_grid,
+                                                     land_times,
+                                                     land_backend,
+                                                     land_time_indexing)
```
glwagner:
```suggestion
                                                    interface_grid,
                                                    clock,
                                                    freshwater_data,
                                                    land_grid,
                                                    land_times,
                                                    land_backend,
                                                    land_time_indexing)
```

bgroenks96 (approved):
A few minor questions but otherwise looks fine to me!

src/EarthSystemModels/earth_system_model.jl
glwagner: its a fine comment -- I fixed it. For reference, the logic (simply because some convention is useful) is to treat "top to bottom" and "left to right" as proceeding physically from the top of the Earth system to the bottom. And for "side by side" situations, like land, sea ice, ocean, we put land first.

src/EarthSystemModels/earth_system_model.jl
glwagner: I'd prefer to add that alias at the time that we implement an example with tests, since otherwise it would be unusable.

test/test_ocean_sea_ice_model.jl
glwagner: this should be read as "ocean component for a coupled model that includes land [for testing]". I can change the name if you have a good suggestion.

test/test_surface_fluxes.jl
glwagner: There is no difference in the constructor. The difference is in how the object is subsequently used for the test.

bgroenks96: Oh, ok. I guess I was expecting `ocean_no_land` to have `land = nothing`?

test/test_ocean_sea_ice_model.jl
bgroenks96: It's fine if the convention is to always initialize the other components to `Prescribed*` in `*_simulation`. But I would expect to see `land = nothing` in the "no land" case, I guess.

test/test_surface_fluxes.jl
```diff
@@ -176,6 +176,46 @@ end
...
+
+        # Test that fluxes work with and without land
+        ocean_no_land = ocean_simulation(grid;
+                                         momentum_advection = nothing,
+                                         tracer_advection = nothing,
+                                         closure = nothing,
+                                         bottom_drag_coefficient = 0)
+
+        set!(ocean_no_land.model, T = 15, S = 30)
+        model_no_land = OceanOnlyModel(ocean_no_land; atmosphere)
```
glwagner, line 189: @bgroenks96 in this case `ocean_no_land` is incorporated into `OceanOnlyModel` with no land component.

Where is the inconsistent appearance of `ocean_no_land`? I can fix it if you point me to it.

test/test_surface_fluxes.jl
glwagner: just a few lines above, that is indeed the case. Below, there are tests that `model_no_land` is correctly set up for the case where `land=nothing`

test/test_ocean_sea_ice_model.jl
glwagner: well, to be clear, it does not become an "ocean with land" until it is actually incorporated into `EarthSystemModel` (aliased by `OceanOnlyModel` below). Whether or not a component is used with another component should not entire into the component constructors (hopefully, if we did our job well)

test/test_surface_fluxes.jl
bgroenks96: I think your comment above clears this up.

glwagner:
ouch

glwagner:
Ideas about this error @simone-silvestri ? It took 86 minutes to reach it:

```
Error During Test at /home/runner/work/NumericalEarth.jl/NumericalEarth.jl/test/test_veros.jl:21
  Test threw exception
  Expression: begin
    #= /home/runner/work/NumericalEarth.jl/NumericalEarth.jl/test/test_veros.jl:22 =#
    ocean = VerosModule.VerosOceanSimulation("global_4deg", :GlobalFourDegreeSetup)
    #= /home/runner/work/NumericalEarth.jl/NumericalEarth.jl/test/test_veros.jl:23 =#
    #= /home/runner/work/NumericalEarth.jl/NumericalEarth.jl/test/test_veros.jl:23 =# @test ocean isa VerosModule.VerosOceanSimulation
    #= /home/runner/work/NumericalEarth.jl/NumericalEarth.jl/test/test_veros.jl:24 =#
    #= /home/runner/work/NumericalEarth.jl/NumericalEarth.jl/test/test_veros.jl:24 =# @test !(isnothing(ocean.setup))
    #= /home/runner/work/NumericalEarth.jl/NumericalEarth.jl/test/test_veros.jl:25 =#
    true
end
  Python: AttributeError: module 'h5py.h5f' has no attribute 'LIBVER_V200'
  Python stacktrace:
   [1] <module>
     @ /tmp/jl_zXNhkv/.CondaPkg/.pixi/envs/default/lib/python3.12/site-packages/h5py/_hl/files.py:45
   [2] <module>
     @ /tmp/jl_zXNhkv/.CondaPkg/.pixi/envs/default/lib/python3.12/site-packages/h5py/__init__.py:60
   [3] <module>
     @ /tmp/jl_zXNhkv/.CondaPkg/.pixi/envs/default/lib/python3.12/site-packages/h5netcdf/core.py:26
   [4] <module>
     @ /tmp/jl_zXNhkv/.CondaPkg/.pixi/envs/default/lib/python3.12/site-packages/h5netcdf/__init__.py:16
   [5] <module>
     @ /tmp/jl_zXNhkv/.CondaPkg/.pixi/envs/default/lib/python3.12/site-packages/veros/setups/global_4deg/global_4deg.py:2
   [6] <module>
     @ /tmp/jl_zXNhkv/.CondaPkg/.pixi/envs/default/lib/python3.12/site-packages/veros/setups/global_4deg/__init__.py:1
  Stacktrace:
   [1] pythrow()
     @ PythonCall.Core ~/.julia/packages/PythonCall/83z4q/src/Core/err.jl:77
   [2] errcheck
     @ ~/.julia/packages/PythonCall/83z4q/src/Core/err.jl:10 [inlined]
   [3] pyimport(m::String)
     @ PythonCall.Core ~/.julia/packages/PythonCall/83z4q/src/Core/builtins.jl:1458
   [4] NumericalEarthVerosExt.VerosOceanSimulation(setup::String, setup_name::Symbol)
     @ NumericalEarthVerosExt ~/work/NumericalEarth.jl/NumericalEarth.jl/ext/NumericalEarthVerosExt/veros_ocean_simulation.jl:212
   [5] top-level scope
     @ ~/work/NumericalEarth.jl/NumericalEarth.jl/test/test_veros.jl:10
   [6] macro expansion
     @ /opt/hostedtoolcache/julia/1.12.5/x64/share/julia/stdlib/v1.12/Test/src/Test.jl:1776 [inlined]
   [7] macro expansion
     @ ~/work/NumericalEarth.jl/NumericalEarth.jl/test/test_veros.jl:21 [inlined]
   [8] macro expansion
     @ /opt/hostedtoolcache/julia/1.12.5/x64/share/julia/stdlib/v1.12/Test/src/Test.jl:677 [inlined]
   [9] macro expansion
     @ ~/work/NumericalEarth.jl/NumericalEarth.jl/test/test_veros.jl:22 [inlined]
Error in testset Veros ocean model interface:
```

simone-silvestri (approved):

simone-silvestri:
> Ideas about this error @simone-silvestri ? It took 86 minutes to reach it:

It should be solved (hopefully) in #195

PR #145, opened 2026-03-26
Update to Oceananigans 0.106.3

simone-silvestri:
to be able to run benchmarks in main, at the moment we are running blind because the two repos are not updated.
This PR adds a `reconcile_state!` for the ESM and a `maybe_prepare_first_time_step!` in line with the new syntax of Oceananigans 0.106.3

PR #152, opened 2026-04-01
Adding a surface layer restoring mask

simone-silvestri:
The usage is rather simple:
```
mask = SurfaceLayerMask()
restoring = DatasetRestoring(; mask, dataset=....)
```


simone-silvestri:
Ups, I have started from #145.

src/DataWrangling/restoring.jl
```diff
@@ -297,3 +297,8 @@ end
     λ, φ, z = node(i, j, k, grid, LX(), LY(), LZ())
     return mask(φ, z)
 end
+
+""" a mask for restoring only the surface layer (Nz) """
+struct SurfaceLayerMask end
+
+@inline stateindex(mask::SurfaceLayerMask, i, j, k, grid, time, loc) = convert(eltype(grid), k == size(grid, 3))
```
glwagner, line 304: this is only the ocean surface, but not the surface for the atmosphere. What does it do for sea ice?

A different name that generalizes to all components would probably be better. `TopLayerMask` perhaps.

glwagner (commented):
find a name that generalizes to all `NumericalEarth` components.

simone-silvestri:
I think this can be closed, the correct approach is #155

PR #153, opened 2026-04-02
Adds ORCA2 tripolar grid and changed ORCAGrid to ORCATripolarGrid

jagoosw:
This PR adds ORCA2 grids to data wrangling and makes some changes to grid construction to get it working. It also renamed `ORCAGrid` to `ORCATripolarGrid`, adds a "coarse" example on ORCA2, and adds the option to remove closed basins int he grid constructor.

ORCA2 is strange because it has regional refinement in the Mediterranean, Red, Black and Caspian Seas, as well as at the equator and south pole. 

The underlying grid:
image

```julia
Nz = 31
z = ExponentialDiscretization(31, -5000, 0; scale = 1100)
grid = ORCAGrid(z = z.faces, Nz, dataset = ORCA2())
```

I had to add Tar and CodecZlib to the dependancies because the Zenodo I found with the grid has them in a tarball.

@simone-silvestri @glwagner


src/DataWrangling/ORCA/ORCA.jl
```diff
@@ -87,4 +89,60 @@ end
...
+function download_dataset(metadatum::ORCA2Metadatum)
+    fileurl  = metadata_url(metadatum)
+    filepath = metadata_path(metadatum)
+    fileroute = metadatum.dir
+
+    @root if !isfile(filepath)
+        @info "Downloading ORCA2 data: $(metadatum.name) to $(metadatum.dir)..."
+        Downloads.download(fileurl, fileroute*"ORCA2L75.tar.gz"; progress=download_progress)
+        open(GzipDecompressorStream, fileroute*"ORCA2L75.tar.gz") do io
+            Tar.extract(io, fileroute*"/ORCA2L75")
```
jagoosw, line 170: `Tar.extract` doesn't like the destination `fileroute` so the files actually live at `fileroute*"/ORCA2L75/ORCA2L75"`

glwagner:
Is this a tripolar grid? Does it have a zipper bc? Can we call it ORCATripolarGrid?

glwagner:
> I had to add Tar and CodecZlib to the dependancies because the Zenodo I found with the grid has them in a tarball.

@giordano is this correct or should we use an extension

@jagoosw one possibility is to put this in a new package (we can put it in NumericalEarth) and then add an extension for the deps

giordano:
> @giordano is this correct or should we use an extension

I think this is correct: if this was in an extension, users would need to know in advance what kind of files they're downloading and explicitly load `Tar`/`CodecZlib` just to be safe (I'd assume at least `Tar` is already loaded via other dependencies anyway), which doesn't seem to be very user-friendly.  Also, pkgextensions usually are meant to extend methods on types defined in the packages they glue, but the downloaded file isn't a different type depending on its format, so there wouldn't be many types for which to extend methods.

jagoosw:
> @jagoosw one possibility is to put this in a new package (we can put it in NumericalEarth) and then add an extension for the deps

Yeah its very similar to ORCA1 which is already in the file I added to, they just have some regional refinement too.

https://www.nemo-ocean.eu/doc/node108.html#Tab_orca_zgr

jagoosw:
> I think this is correct: if this was in an extension, users would need to know in advance what kind of files they're downloading and explicitly load `Tar`/`CodecZlib` just to be safe (I'd assume at least `Tar` is already loaded via other dependencies anyway), which doesn't seem to be very user-friendly. Also, pkgextensions usually are meant to extend methods on types defined in the packages they glue, but the downloaded file isn't a different type depending on its format, so there wouldn't be many types for which to extend methods.

I think you're right that Tar was already loaded

jagoosw:
When I try to run the example I get NaNs on the first time step, it seems ORCA1 does the same too? (@simone-silvestri)

(The example I added is a copy of the one degree example, I don't know if we want it in the examples folder in this PR but there doesn't seem to be a validation folder)

examples/coarse_example.jl
```diff
@@ -27,7 +27,7 @@ Nz = 31
 depth = 5000meters
 z = ExponentialDiscretization(Nz, -depth, 0; scale = depth/4, mutable = true)
 
-grid = ORCAGri(arch, z = z.face, Nz, dataset = ORCA2())
+grid = ORCAGrid(arch; z, Nz, dataset = ORCA2(), halo = (5, 5, 4))
```
glwagner: Can we use `ORCATripolarGrid` ?

glwagner (commented):
ORCATripolarGrid?

simone-silvestri:
There were a couple of bugs that have been solved, now I can run an ORCA1 with 30minutes quite stably

jagoosw:
I figured out the problem is that ORCA2 (and 0.5 and 0.25) use T fold points so I have to get rid of this:
https://github.com/NumericalEarth/NumericalEarth.jl/blob/99c763f6e8269a594f41956e6f293d18aefbfb44/src/Bathymetry/orca_grid.jl#L68-L72
and just do the same as Center for everything.

Is it okay for me to add a fold topology argument to that function and do `(LY === Center)|is_a_T_fold`?

glwagner:
Is this a tripolar grid?

src/Bathymetry/orca_tripolar_grid.jl
glwagner, line 1: `orca_tripolar_grid.jl`

test/test_orca_tripolar_grid.jl
glwagner, line 1: `test_orca_tripolar_grid.jl`

src/Bathymetry/orca_tripolar_grid.jl
```diff
@@ -85,8 +85,11 @@ function halo_fill_stagger(CC, FC, CF, FF, helper_grid, bcs, overlap)
     )
 end
 
+fold_topology(::ORCA1) = RightFaceFolded
+fold_topology(::ORCA2) = RightCenterFolded
```
simone-silvestri: Is this really a RightCenterFolded?

src/DataWrangling/ORCA/ORCA.jl
```diff
@@ -87,4 +89,60 @@ end
 
 default_south_rows_to_remove(::ORCA1) = 35
 
+# ORCA2
+struct ORCA2 end
+
+default_download_directory(::ORCA2) = download_ORCA_cache
+reversed_vertical_axis(::ORCA2) = false
+longitude_interfaces(::ORCA2) = (-180, 180)
+latitude_interfaces(::ORCA2) = (-80, 90)
```
simone-silvestri: do we need these methods?

test/test_orca_tripolar_grid.jl
```diff
@@ -0,0 +1,157 @@
...
+        @test underlying isa TripolarGrid
+        @test underlying.Nx == Nx0
+        @test underlying.Ny == Ny0
+        @test underlying.Nz == 5
+
+        # Coordinates span near-global domain
+        @test minimum(underlying.λᶜᶜᵃ.parent) < -179
+        @test maximum(underlying.λᶜᶜᵃ.parent) > 179
+        @test minimum(underlying.φᶜᶜᵃ.parent) < -70
+        @test maximum(underlying.φᶜᶜᵃ.parent) > 80
```
simone-silvestri:
```suggestion
        @test maximum(underlying.φᶜᶜᵃ.parent) > 87
```
for 2 degree :)

simone-silvestri:
Maybe we can also add the example to the literated examples? We would need a coarse resolution example.

src/Bathymetry/orca_tripolar_grid.jl
jagoosw: I think so, the caption for figure 8.4 on this page says ORCA2, 1/4, and 1/12 uses T point folds https://www.nemo-ocean.eu/doc/node59.html#Fig_North_Fold_T

src/DataWrangling/ORCA/ORCA.jl
jagoosw: Probably not, I will clean up

jagoosw:
> Maybe we can also add the example to the literated examples? We would need a coarse resolution example.

Should I do something similar to the 1 degree ocean-sea ice example or should we do something different?

src/Bathymetry/orca_tripolar_grid.jl
simone-silvestri: A `RightCenterFolded` is a fold on U points, if ORCA2 pivots on T-points (tracers) we don't have it.

simone-silvestri:
I think so. I think we also do not need WENO for momentum advection at 2 degrees, we can safely use `VectorInvariant` to reduce computational time.

jagoosw:
> I think so. I think we also do not need WENO for momentum advection at 2 degrees, we can safely use `VectorInvariant` to reduce computational time.

I've been using vector invariant with WENO3 vorticity and UP1 divergence, what do you think I should try for vorticity? Also what do you think I need for the sea-ice? I've been using WENO3 and I read its important that its more important that the scheme is conservative for ice?

simone-silvestri:
I think we can use just `VectorInvariant()`, WENO3 and UP1 are very very very diffusive, in practice I think it is better to use explicit diffusivity. 
For sea ice you can probably just use `dynamics=nothing`, I am not sure that they matter too much at coarse resolution. Sorry I closed by mistake, reopened it

jagoosw:
Can I add GeoMakie as a dependancy for the docs? I made this plot but perhaps its too messy +/- idk if we want to depend on geomakie too.


https://github.com/user-attachments/assets/77e783a1-bd5d-4f68-98cc-14ae6abb0592
(^ this version has sea ice dynamics on)

Updated:
image
image

jagoosw:
Updated with no GeoMakie dependency

https://github.com/user-attachments/assets/1a9982ac-a5af-49dd-a306-f424af66bf9a

src/DataWrangling/ORCA/ORCA.jl
jagoosw: Resolved now

jagoosw:
@simone-silvestri should I temporarily turn this example on to litarate so we can see what it looks like? Also, I saw that the examples on main take ~2.5hours to run: are the executed serially or together? If serially I should reduce how long this example runs for.

simone-silvestri:
very nice plots! I think you can have the ice surface temperature in the range -50 / 0 degrees. The examples are executed on 2 V100 GPUs, how long does this run for?

jagoosw:
> very nice plots! I think you can have the ice surface temperature in the range -50 / 0 degrees. The examples are executed on 2 V100 GPUs, how long does this run for?

Oh yeah thats much better 

https://github.com/user-attachments/assets/ba44e958-3b03-497d-af80-c326640d5659

It takes about 80minutes of running but ~2hours+ overall with the model construction on an A100

simone-silvestri:
Is this ready to review / merge? also #168 is changing the orca constructors. Maybe we want to wait for this PR to avoid too many conflicts?

jagoosw:
> Is this ready to review / merge? also #168 is changing the orca constructors. Maybe we want to wait for this PR to avoid too many conflicts?

Yeah I think this is ready now!

Happy to wait for #168 if you want to do them in that order

jagoosw:
This is a breaking change since it changes ORCAGrid to ORCATripolarGrid

taimoorsohail:
Yes I will merge #168  today if you can wait :)

src/Bathymetry/orca_tripolar_grid.jl
jagoosw: Sorry I missed this reply. Where does a `RightFaceFolded` fold I thought that was on a U point?

jagoosw:
I'm just sorting out the conflicts and everything is straight forward except the tests. I had modified the tests to do do all the same tests on the ORCA2 grid as the ORCA1 grid, but the incoming changes for ORCA12 don't test everything like metric consistency. Thinking about it now we probably don't need todo all the tests on all the grids because if the constructor works for one it should work for all of them? Should I make the tests just do basic tests on ORCA2 like ORCA12, or should I extend to include ORCA12 in all the tests, or do all the tests on 1 and 2 but not 12? @simone-silvestri @taimoorsohail

simone-silvestri:
I think they are quite simple tests so worth to add also ORCA12 to the more in-depth testing

jagoosw:
Okay I added them and ORCA12 is failing the metric consistency tests:
```julia
ORCATripolarGrid (ORCA12) metric consistency: Test Failed at /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102
  Expression: all((x->begin
                #= /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 =#
                x > 0
            end), interior) == true
   Evaluated: false == true

Stacktrace:
 [1] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 [inlined]
 [3] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] top-level scope
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:80
ORCATripolarGrid (ORCA12) metric consistency: Test Failed at /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102
  Expression: all((x->begin
                #= /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 =#
                x > 0
            end), interior) == true
   Evaluated: false == true

Stacktrace:
 [1] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 [inlined]
 [3] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] top-level scope
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:80
ORCATripolarGrid (ORCA12) metric consistency: Test Failed at /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102
  Expression: all((x->begin
                #= /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 =#
                x > 0
            end), interior) == true
   Evaluated: false == true

Stacktrace:
 [1] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 [inlined]
 [3] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] top-level scope
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:80
ORCATripolarGrid (ORCA12) metric consistency: Test Failed at /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102
  Expression: all((x->begin
                #= /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 =#
                x > 0
            end), interior) == true
   Evaluated: false == true

Stacktrace:
 [1] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 [inlined]
 [3] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] top-level scope
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:80
ORCATripolarGrid (ORCA12) metric consistency: Test Failed at /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102
  Expression: all((x->begin
                #= /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 =#
                x > 0
            end), interior) == true
   Evaluated: false == true

Stacktrace:
 [1] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 [inlined]
 [3] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] top-level scope
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:80
ORCATripolarGrid (ORCA12) metric consistency: Test Failed at /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102
  Expression: all((x->begin
                #= /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 =#
                x > 0
            end), interior) == true
   Evaluated: false == true

Stacktrace:
 [1] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 [inlined]
 [3] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] top-level scope
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:80
ORCATripolarGrid (ORCA12) metric consistency: Test Failed at /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102
  Expression: all((x->begin
                #= /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 =#
                x > 0
            end), interior) == true
   Evaluated: false == true

Stacktrace:
 [1] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 [inlined]
 [3] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] top-level scope
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:80
ORCATripolarGrid (ORCA12) metric consistency: Test Failed at /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102
  Expression: all((x->begin
                #= /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 =#
                x > 0
            end), interior) == true
   Evaluated: false == true

Stacktrace:
 [1] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 [inlined]
 [3] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] top-level scope
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:80
ORCATripolarGrid (ORCA12) metric consistency: Test Failed at /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102
  Expression: all((x->begin
                #= /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 =#
                x > 0
            end), interior) == true
   Evaluated: false == true

Stacktrace:
 [1] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 [inlined]
 [3] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] top-level scope
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:80
ORCATripolarGrid (ORCA12) metric consistency: Test Failed at /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102
  Expression: all((x->begin
                #= /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 =#
                x > 0
            end), interior) == true
   Evaluated: false == true

Stacktrace:
 [1] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 [inlined]
 [3] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] top-level scope
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:80
ORCATripolarGrid (ORCA12) metric consistency: Test Failed at /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102
  Expression: all((x->begin
                #= /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 =#
                x > 0
            end), interior) == true
   Evaluated: false == true

Stacktrace:
 [1] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 [inlined]
 [3] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] top-level scope
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:80
ORCATripolarGrid (ORCA12) metric consistency: Test Failed at /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102
  Expression: all((x->begin
                #= /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 =#
                x > 0
            end), interior) == true
   Evaluated: false == true

Stacktrace:
 [1] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:102 [inlined]
 [3] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] top-level scope
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:80
ORCATripolarGrid (ORCA12) metric consistency: Test Failed at /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:121
  Expression: n_west / Nx > 0.95
   Evaluated: 0.5002313743637204 > 0.95

Stacktrace:
 [1] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:121 [inlined]
 [3] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] top-level scope
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:80
ORCATripolarGrid (ORCA12) metric consistency: Test Failed at /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:129
  Expression: nsouth / length(φC) > 0.95
   Evaluated: 0.4337215751525236 > 0.95

Stacktrace:
 [1] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:129 [inlined]
 [3] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] top-level scope
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:80
ORCATripolarGrid (ORCA12) metric consistency: Test Failed at /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:140
  Expression: boundary_jump < 10interior_variation + 1.0e-10
   Evaluated: NaN < NaN

Stacktrace:
 [1] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:140 [inlined]
 [3] macro expansion
   @ ~/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] top-level scope
   @ ~/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:80
Test Summary:                                | Pass  Fail  Total   Time
ORCATripolarGrid (ORCA12) metric consistency |   20    15     35  17.0s
ERROR: LoadError: Some tests did not pass: 20 passed, 15 failed, 0 errored, 0 broken.
in expression starting at /Users/jago/Documents/Projects/NumericalEarth.jl/test/test_orca_tripolar_grid.jl:17
```

@taimoorsohail have you managed to run a model using the ORCA12 grid?

simone-silvestri:
Ok then I guess and issue that tracks the failure is more suited 😅 So we can merge this PR and deal with ORCA12 in another PR

jagoosw:
To summarise I think the tests are telling us that:
- ORCA12 has interior points with grid spacing or areas <= 0
- Lots of face points are the "wrong" side of the center points
- Theres something wrong with the periodic boundary

jagoosw:
> Ok then I guess and issue that tracks the failure is more suited 😅 So we can merge this PR and deal with ORCA12 in another PR

~okay I'll mark the tests as broken ~ I've commented out ORCA12 from the tests because it was easier than marking every test as broken

jagoosw:
I think this is ready for review now. @simone-silvestri do you think I should reduce the stop time for the example?

examples/orca2.jl
```diff
@@ -0,0 +1,338 @@
...
+radiation  = Radiation(arch)
+atmosphere = JRA55PrescribedAtmosphere(arch; backend=JRA55NetCDFBackend(),
+                                       include_rivers_and_icebergs = true)
+
+# ### Coupled simulation
+
+# Now we are ready to build the coupled ocean--sea ice model and bring everything
+# together into a `simulation`.
+
+# With Runge-Kutta 3rd order time-stepping we can safely use a timestep of 20 minutes.
```
simone-silvestri:
```suggestion
# With Runge-Kutta 3rd order time-stepping we can safely use a timestep of 90 minutes.
```

examples/orca2.jl
```diff
@@ -0,0 +1,339 @@
...
+
+# ### Coupled simulation
+
+# Now we are ready to build the coupled ocean--sea ice model and bring everything
+# together into a `simulation`.
+
+# With Runge-Kutta 3rd order time-stepping we can safely use a timestep of 90 minutes.
+
+coupled_model = OceanSeaIceModel(ocean, sea_ice; atmosphere, radiation)
+simulation = Simulation(coupled_model; Δt=90minutes, stop_time=5*365days)
```
simone-silvestri:
```suggestion
simulation = Simulation(coupled_model; Δt=90minutes, stop_time=2.5*365days)
```
Maybe we can do 2.5 years to reduce a bit the computational burden?

experiments/orca2.jl
```diff
@@ -0,0 +1,339 @@
...
+using Oceananigans.TurbulenceClosures: IsopycnalSkewSymmetricDiffusivity
+eddy_closure = IsopycnalSkewSymmetricDiffusivity(κ_skew=500, κ_symmetric=500)
+
+using Oceananigans.TurbulenceClosures.TKEBasedVerticalDiffusivities:
+    CATKEVerticalDiffusivity
+
+vertical_mixing = CATKEVerticalDiffusivity(VerticallyImplicitTimeDiscretization(); 
+                                           maximum_tracer_diffusivity = 1,
+                                           maximum_tke_diffusivity = 1,
+                                           maximum_viscosity = 1)
```
simone-silvestri, line 62: we cannot use the `default_ocean_closure`?

examples/orca2.jl
```diff
@@ -0,0 +1,339 @@
...
+
+using Oceananigans.TurbulenceClosures.TKEBasedVerticalDiffusivities:
+    CATKEVerticalDiffusivity
+
+vertical_mixing = CATKEVerticalDiffusivity(VerticallyImplicitTimeDiscretization(); 
+                                           maximum_tracer_diffusivity = 1,
+                                           maximum_tke_diffusivity = 1,
+                                           maximum_viscosity = 1)
+
+free_surface       = SplitExplicitFreeSurface(grid; substeps=20*3)
```
simone-silvestri:
```suggestion
free_surface       = SplitExplicitFreeSurface(grid; substeps=60)
```

examples/orca2.jl
```diff
@@ -0,0 +1,339 @@
...
+momentum_advection = WENOVectorInvariant(order=5)
+tracer_advection   = WENO(order=5)
+
+dates = DateTime(1993, 1, 1) : Month(1) : DateTime(1993, 11, 1)
+mask = LinearlyTaperedPolarMask(southern=(-80, -70), northern=(70, 90), z=(-100, 0))
```
simone-silvestri: do we need to restore under sea ice?

test/test_orca_tripolar_grid.jl
```diff
@@ -0,0 +1,157 @@
...
+
+        ds = Dataset(path)
+        name = 
+        bathy = ds[bathymetry_names[n]][:, :]
+        close(ds)
+
+        @test size(bathy) == (Nx0, Ny0)
+        @test maximum(bathy) > 5000  # Deep ocean
+    end
+end
```
simone-silvestri, line 192:
```suggestion
end

```

simone-silvestri (approved):

experiments/orca2.jl
jagoosw: I think I was getting instability when I didn't have it capped, but I'll rerun and check

examples/orca2.jl
jagoosw: Not sure, should I try without the mask?

jagoosw:
I think I've resolved the test conflicts but because I've renamed the test file it didn't show me the diffs automatically, is it just the new bit to predownload the datasets?

experiments/orca2.jl
glwagner: I don't think we should have any examples that don't use the default --- we should fix the default instead.

examples/orca2.jl
simone-silvestri: yeah I think so.

simone-silvestri:
I think the last bit before merging is to add these to the CATKE defaults:
```julia
maximum_tracer_diffusivity = 1
maximum_tke_diffusivity = 1
maximum_viscosity = 1
```
(If they are truly needed). I don't think it will damage the accuracy of the simulation, given that 1 $m^2/s$ is already quite a large diffusivity but I defer to @glwagner to assess if this is a correct move.

simone-silvestri:
>  I think I've resolved the test conflicts but because I've renamed the test file it didn't show me the diffs automatically, is it just the new bit to predownload the datasets?

you mean for `test_orca_grid`? I think so but you can confirm by diffing what it is main now

jagoosw:
I just checked and when I run without the limits I got NaNs. If I recall correctly, when I originally investigated this it was coming from a storm. I think maybe its a problem that is manifesting on coarse grids because if e.g. we interpolate the atmosphere fields and the coarse point happens to be near a very high wind speed point then that high wind speed is applied over the whole cell injecting a lot more momentum than on a more refined grid that resolves the variability in the atmosphere. I don't know if there are other strategies for downscaling the atmosphere some kind of averaging that people use?

glwagner:
My main concern is that the limiters are unphysical. A diffusivity that large can occur for winds of 33 m/s and a 1 m mixing length, which seems possible in extreme conditions for sure.

The question is why 1 m^2 / s lead to blow up.

jagoosw:
> My main concern is that the limiters are unphysical. A diffusivity that large can occur for winds of 33 m/s and a 1 m mixing length, which seems possible in extreme conditions for sure.
> 
> The question is why 1 m^2 / s lead to blow up.

1m^2/s was an arbitrary choice, I can try running with it higher to see what the stable maximum is. I'm not sure exactly how the instability is arising, I think there can be a feedback with the bulk formula fluxs if the fluxs become large which could happen if theres a deep mixing event and e.g. the SST suddenly drops?

PR #142, opened 2026-03-24
(0.3.0) Refactor bounding box to Metadata.region and add Column 

glwagner:
## Summary

- Introduces a `Column(longitude, latitude)` type as a new spatial region for `Metadata`, separate from `BoundingBox`
- When `region=Column(lon, lat)`, `native_grid` returns a column `RectilinearGrid` with `(Flat, Flat, Bounded)` topology and `location` automatically reduces horizontal dimensions to `Nothing`
- Renames `bounding_box` field → `region` on `Metadata` struct (accepts `Nothing`, `BoundingBox`, or `Column`)
- Renames per-dataset `location()` → `dataset_location()` with a generic `location(::Metadata)` wrapper that applies region-based dimension reduction
- Supports `Linear()` (default) and `Nearest()` interpolation for column data extraction

### Example usage
```julia
col = Column(35.1, 50.1)
T_meta = Metadatum(:temperature, dataset=GLORYSMonthly(), region=col)

native_grid(T_meta)  # → RectilinearGrid(Flat, Flat, Bounded)
location(T_meta)     # → (Nothing, Nothing, Center)
T_field = Field(T_meta)  # → column Field{Nothing, Nothing, Center}
```

### Breaking changes
- `Metadata.bounding_box` renamed to `Metadata.region`
- `bounding_box=` keyword argument renamed to `region=` in `Metadata()` and `Metadatum()` constructors
- Per-dataset `location()` methods renamed to `dataset_location()`

Closes #138

## Test plan
- [x] Verify existing tests pass after rename
- [x] Test `Column(lon, lat)` construction
- [x] Test `native_grid` returns `RectilinearGrid` for `Column` region
- [x] Test `location` returns `(Nothing, Nothing, Center)` for `Column` region
- [x] End-to-end test: ocean station papa with GLORYS column metadata
- [x] End-to-end test: RICO site with ERA5 column metadata

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #155, opened 2026-04-08
Surface restoring BC

simone-silvestri:
This PR introduces surface restoring as a boundary condition for the ocean. There are a couple of advantages in doing this rather than a 3D forcing:
- We can use datasets that have a smaller vertical size than the simulation grid (for example WOAMontly), which would lead to bounds error if we try using as a 3D forcing
- When #142 is ready, we can build a `DatasetRestoring` which is much smaller in memory (i.e. just the surface values), reducing memory usage and possible the need to read new data.


src/Oceans/flux_and_restoring.jl
```diff
@@ -0,0 +1,46 @@
...
+   returns a tendency in the top cell — typically a `DatasetRestoring`,
+   evaluating to `r * μ * (ψ_dataset - ψ)`. The tendency is converted to a
+   surface flux by multiplying by `-Δz` at the top cell, consistent with the
+   Oceananigans top-flux sign convention (top cell tendency contribution is
+   `-J / Δz`).
+
+This lets the coupled flux solver and a dataset restoring share the same top
+boundary condition without one clobbering the other.
+"""
+struct FluxAndRestoring{F, R} <: Function
```
glwagner: I'm confused by this name. Doesn't this return a flux? In that case what does "and restoring" mean?

Is this supposed to represent a flux that is the sum of a prescribed component and a component the represents restoring to some target?

src/Oceans/flux_and_restoring.jl
```diff
@@ -0,0 +1,46 @@
...
+    @inbounds J = fr.flux_field[i, j, 1]
+
+    # Restoring accessed as a tendency forcing (compatible with DatasetRestoring)
+    G = fr.restoring(i, j, Nz, grid, clock, fields)
+
+    # Top BC convention: tendency contribution = -J / Δz, so to inject
+    # `G` in the top cell the flux is `-G * Δz`.
+    Δz = Δzᶜᶜᶜ(i, j, Nz, grid)
+    return J - G * Δz
+end
```
glwagner: I think it would be better to use the `getbc` interface

src/Oceans/flux_and_restoring.jl
```diff
@@ -0,0 +1,46 @@
...
+Adapt.adapt_structure(to, fr::FluxAndRestoring) =
+    FluxAndRestoring(adapt(to, fr.flux_field),
+                     adapt(to, fr.restoring))
+
+@inline function (fr::FluxAndRestoring)(i, j, grid, clock, fields)
+    Nz = size(grid, 3)
+    @inbounds J = fr.flux_field[i, j, 1]
+
+    # Restoring accessed as a tendency forcing (compatible with DatasetRestoring)
+    G = fr.restoring(i, j, Nz, grid, clock, fields)
```
glwagner:
```suggestion
    G = getbc(fr.restoring, i, j, Nz, grid, clock, fields)
```

glwagner: this is more general if I understand correctly.

src/Oceans/ocean_simulation.jl
```diff
@@ -190,6 +196,7 @@ function ocean_simulation(grid;
                           gravitational_acceleration = default_gravitational_acceleration,
                           bottom_drag_coefficient = Default(0.003),
                           forcing = NamedTuple(),
+                          surface_restoring = NamedTuple(),
```
glwagner: The code you have written isn't really specific to surface restoring. The structure of what you have written just adds the ability to consider two components to the top boundary condition (rather than _only_ exchange fluxes).

I think it would be clearer to write the user interface to reflect this, which will make it clearer to users what is happening (and will also generalize to more cases, not just restoring)

src/Oceans/ocean_simulation.jl
```diff
@@ -273,8 +280,8 @@ function ocean_simulation(grid;
     # Construct ocean boundary conditions including surface forcing and bottom drag
     u_top_bc = FluxBoundaryCondition(τˣ)
     v_top_bc = FluxBoundaryCondition(τʸ)
-    T_top_bc = FluxBoundaryCondition(Jᵀ)
-    S_top_bc = FluxBoundaryCondition(Jˢ)
+    T_top_bc = build_top_tracer_bc(Jᵀ, get(surface_restoring, :T, nothing))
+    S_top_bc = build_top_tracer_bc(Jˢ, get(surface_restoring, :S, nothing))
```
glwagner: What is specific to tracer fluxes? I think what you are looking for is something like

```julia
# above
default_additional_fluxes = (u=nothing, v=nothing, T=nothing, S=nothing)
additional_fluxes = merge(default_additional_fluxes, user_additional_fluxes)

# below
u_top_bc = merge_boundary_conditions(τˣ, additional_fluxes.u)
```

etc

src/Oceans/flux_and_restoring.jl
glwagner: This should probably be `MultipleConditions`, analogous with `MultipleForcings` in Oceananigans.

glwagner (commented):
the implementation could be easier to understand and significantly more general with some small changes

simone-silvestri:
What would you suggest? I like the idea of having a `surface_restoring` kwarg because it is quite clear and understandable from the user side.

src/Oceans/ocean_simulation.jl
simone-silvestri: I am not sure about the complete generality of this, if a user wants custom bcs they can pass them to completely overload them, this should really just cover the case of surface restoring. Probably I can make this more clear and specific.

src/Oceans/ocean_simulation.jl
simone-silvestri: I would like it to be quite specific: a computed flux and a restoring. Maybe I can cut down the generality of this BC and make it much more specific. The problem is that `DatasetRestoring` already encodes a generic forcing and I wanted to use this with `DatasetRestoring` to avoid code duplication.

glwagner:
> What would you suggest? I like the idea of having a `surface_restoring` kwarg because it is quite clear and understandable from the user side.

I don't have a view of the entire API. But I would imagine something like this:

```julia
    additional_surface_fluxes = (; S = SurfaceRestoring(metadata)),
    other_kw...
)
```

obviously naming subject to change. 

It is semantically wasteful to have `surface_restoring = (; S=SurfaceRestoring))`, and I also think not as clear exactly how this works (ie when you add surface restoring, we are still going to use the computed fluxes --- this is key).

src/Oceans/flux_and_restoring.jl
navidcy: Yes, I presume it's either flux or restoring?

glwagner: i think its both flux and restoring

src/Oceans/ocean_simulation.jl
glwagner: I disagree with this philosophy because it leads to interfaces that are hard to understand and use

src/Oceans/ocean_simulation.jl
glwagner: The problem with that concept is that it does not reveal to the user the mechanism. It is more like a black box.

glwagner: > this should really just cover the case of surface restoring

Justify this? It's my understanding that surface restoring is only useful for long running climate simulations. It seems like a heavy price / big addition to pay for that small application, especially when a more general implementation is both simpler and easier to understand. What do we gain from the complexity?

simone-silvestri:
I can do `additional_surface_fluxes`!

simone-silvestri:
This should be ready to review

src/DataWrangling/restoring.jl
```diff
@@ -235,6 +236,47 @@ end
...
+restoring = DatasetRestoring(metadata, grid; rate = 1 / 30days)
+ocean = ocean_simulation(grid;
+    additional_surface_fluxes = (; S = SurfaceFluxRestoring(restoring)))
+```
+"""
+struct SurfaceFluxRestoring <: Function
+    dataset_restoring :: DatasetRestoring
+end
+
+Adapt.adapt_structure(sf::SurfaceFluxRestoring) = SurfaceFluxRestoring(Adapt.adapt(to, sfr.dataset_restoring))
```
simone-silvestri:
```suggestion
Adapt.adapt_structure(to, sf::SurfaceFluxRestoring) = SurfaceFluxRestoring(Adapt.adapt(to, sfr.dataset_restoring))
```

simone-silvestri:
This PR should be ready to merge

src/Diagnostics/interface_fluxes.jl
```diff
@@ -1,4 +1,8 @@
 
+@inline flux_field(condition) = condition
+@inline flux_field(bc::MultipleFluxes) = bc.flux_field
+@inline flux_field(bc::DiscreteBoundaryFunction) = flux_field(bc.func)
```
jagoosw, line 4: I initially didn't really understand why `MultipleFluxes` had to have a field as the first flux, but I gather that the intention is to only use this in coupled models in NumericalEarth where the atmosphere/ice is going to write into a field right? 

Is there any cases where `DiscreteBoundaryFunction` would be used here?

jagoosw (approved):
Took me a second to understand the `MultipleFluxs` situation but this seems like a good addition and makes more sense for surface relaxing than evaluating forcing across the whole field!

src/Diagnostics/interface_fluxes.jl
simone-silvestri: `DiscreteBoundaryFunction` here wraps around a `MultipleFluxes`, so when we use `MultipleFluxes` (which is internal), the resulting boundary condition is a `DiscreteBoundaryFunction` where the function is a `MultipleFluxes`

PR #157, opened 2026-04-09
OutputWriters on a ESM

simone-silvestri:
At the moment, once we write on a `Simulation{<:EarthSystemModel}`, we cannot really read the output with a `FieldTimeSeries` because the `EarthSystemModel` lacks a grid. 

This PR introduces the `exchange_grid` as the grid for the `EarthSystemModel` such that we can write and read outputs

### DIFF
--- a/src/EarthSystemModels/earth_system_model.jl
+++ b/src/EarthSystemModels/earth_system_model.jl
@@ -44,13 +44,20 @@ function Base.show(io::IO, cm::ESM)
     return nothing
 end
 
-# Assumption: We have an ocean!
+# Expose the exchange grid as `model.grid` so that JLD2Writer can save it.
+function Base.getproperty(model::ESM, name::Symbol)
+    if name === :grid
+        return getfield(model, :interfaces).exchanger.grid
+    end
+    return getfield(model, name)
+end
+
 architecture(model::ESM)           = model.architecture
 Base.eltype(model::ESM)            = Base.eltype(model.interfaces.exchanger.grid)
 prettytime(model::ESM)             = prettytime(model.clock.time)
 iteration(model::ESM)              = model.clock.iteration
 timestepper(::ESM)                 = nothing
-default_included_properties(::ESM) = tuple()
+default_included_properties(::ESM) = [:grid]
 prognostic_fields(cm::ESM)         = nothing
 fields(::ESM)                      = NamedTuple()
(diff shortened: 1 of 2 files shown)


src/EarthSystemModels/earth_system_model.jl
```diff
@@ -44,13 +44,20 @@ function Base.show(io::IO, cm::ESM)
...
 end
 
-# Assumption: We have an ocean!
+# Expose the exchange grid as `model.grid` so that JLD2Writer can save it.
+function Base.getproperty(model::ESM, name::Symbol)
+    if name === :grid
+        return getfield(model, :interfaces).exchanger.grid
+    end
+    return getfield(model, name)
+end
```
glwagner, line 53: It is better to have an explicit interface, rather than abusing `getproperty`, I think (Claude does like to do this)

I guess in this case, it would be `getgrid(model)`.

glwagner:
Should we relax the constraint that output writers need grids? If I recall this is a fairly minor change?

simone-silvestri:
I think they work, the problem I have is the compatibility with `FieldTimeSeries` that requires a grid. It would be possible to attach the output writer to the ocean to circumvent this problem, but I think it is cleaner to attach it to the esm when it comes to fluxes and interface variables.

glwagner:
> I think they work, the problem I have is the compatibility with `FieldTimeSeries` that requires a grid. It would be possible to attach the output writer to the ocean to circumvent this problem, but I think it is cleaner to attach it to the esm when it comes to fluxes and interface variables.

There is something weird though --- `FieldTimeSeries` should take grid _locally_ for each field, rather than "globally" for the whole file. Otherwise, we cannot save fields on different grids in a single file. And in my opinion one of the whole points of having an output writer on the coupled model is to allow such a capability.

simone-silvestri:
yeah, definitely, let's implement that!

src/EarthSystemModels/earth_system_model.jl
simone-silvestri: I think we can have a `grid()` accessor for fields (like we have for the other field properties) that can also be dispatched on the model. I am starting to implement it here https://github.com/CliMA/Oceananigans.jl/pull/5486

simone-silvestri:
I ll close this because this is dealt with in [Oceananigans](https://github.com/CliMA/Oceananigans.jl/pulls)

PR #158, opened 2026-04-09
Fix surface variables' range for Oceananigans' oceans

simone-silvestri:
After https://github.com/CliMA/Oceananigans.jl/pull/5408, we have switched `:xy` to mean `Nx, Ny+1, Nz` for a `RIghtFaceFolded` tripolar grid. For this reason, the `_assemble_net_ocean_fluxes!` which needs access to surface salinity fails because we pass the `interior` of the tracers.

This PR changes the surface variables to include the halos, so we are always safe.

### DIFF
--- a/src/Oceans/Oceans.jl
+++ b/src/Oceans/Oceans.jl
@@ -74,12 +74,12 @@ ocean_temperature(ocean::Simulation{<:HydrostaticFreeSurfaceModel}) = ocean.mode
 
 function ocean_surface_salinity(ocean::Simulation{<:HydrostaticFreeSurfaceModel})
     kᴺ = size(ocean.model.grid, 3)
-    return interior(ocean.model.tracers.S, :, :, kᴺ:kᴺ)
+    return view(ocean.model.tracers.S.data, :, :, kᴺ:kᴺ)
 end
 
 function ocean_surface_temperature(ocean::Simulation{<:HydrostaticFreeSurfaceModel})
     kᴺ = size(ocean.model.grid, 3)
-    return interior(ocean.model.tracers.T, :, :, kᴺ:kᴺ)
+    return view(ocean.model.tracers.T.data, :, :, kᴺ:kᴺ)
 end
 
 function ocean_surface_velocities(ocean::Simulation{<:HydrostaticFreeSurfaceModel})


src/Oceans/Oceans.jl
```diff
@@ -74,12 +74,12 @@ ocean_temperature(ocean::Simulation{<:HydrostaticFreeSurfaceModel}) = ocean.mode
...
     kᴺ = size(ocean.model.grid, 3)
-    return interior(ocean.model.tracers.S, :, :, kᴺ:kᴺ)
+    return view(ocean.model.tracers.S.data, :, :, kᴺ:kᴺ)
 end
 
 function ocean_surface_temperature(ocean::Simulation{<:HydrostaticFreeSurfaceModel})
     kᴺ = size(ocean.model.grid, 3)
-    return interior(ocean.model.tracers.T, :, :, kᴺ:kᴺ)
+    return view(ocean.model.tracers.T.data, :, :, kᴺ:kᴺ)
 end
```
glwagner, line 83: where are these used? This is not the right form if this is user facing, which arguably it should be. It would be better to have a more general implementation that can be used both externally and internally.

simone-silvestri: ah no, these are internal to the model, and used to compute fluxes, not user facing.

glwagner: yeah, that makes sense (which I gathered from `interior`) but I believe we will need user facing diagnostics which have exactly the same name as these

simone-silvestri: My conundrum is this one: 
I could use `view(field, :, :, Nz)` which would return me a `Field` type, which would be better, but then I need to access this as `field[i, j, Nz]` instead of `field[i, j, 1]` which does not make sense because it carries information of the ocean instead of being general. So what I would really like is a `Field` which can be accessed with `field[i, j, 1]` and does not copy the data around. Maybe I can do it by constructing a 
`Field{Center, Center, Nothing}(grid, data=view(f.data, :, :, Nz))` I will try like this.

simone-silvestri: It was actually not possible, this motivated creating a new feature: https://github.com/CliMA/Oceananigans.jl/pull/5487

glwagner: I think that you can have something like:

```julia
ocean_surface_temperature(ocean, field=true)
```

for `field=false` return the view of offset array

simone-silvestri:
This should become a `slice` once https://github.com/CliMA/Oceananigans.jl/pull/5487 is merged

PR #162, opened 2026-04-15
Add LargeYeagerTransferCoefficients bulk flux formulation

simone-silvestri:
## Summary

Refactor `CoefficientBasedFluxes` to cleanly support the Large & Yeager (2004, 2009)
stability-corrected bulk algorithm alongside constant and polynomial coefficients.

The previous API used separate keyword arguments (`drag_coefficient`, `heat_transfer_coefficient`,
`vapor_flux_coefficient`), which made it awkward to pass coupled coefficient formulations like
the L&Y algorithm — where all three coefficients are computed together from a single set of
parameters. To resolve this, the three separate fields are replaced by a single
`transfer_coefficients` field that accepts either a `SimilarityScales` tuple (for independent
constant or callable coefficients) or a `LargeYeagerTransferCoefficients` object (for the
coupled L&Y formulation). The `von_karman_constant` is moved from `CoefficientBasedFluxes`
into `LargeYeagerTransferCoefficients`, where it is actually used.

### Changes

- Replace `drag_coefficient`, `heat_transfer_coefficient`, `vapor_flux_coefficient` with a single `transfer_coefficients` keyword
- Add `LargeYeagerTransferCoefficients` implementing the full L&Y 2004/2009 algorithm with Monin-Obukhov stability corrections (OMIP-2 protocol)
- Add `PolynomialNeutralDragCoefficient` for wind-speed-dependent neutral drag
- Add `LinearStableStabilityFunction` and `large_yeager_stability_functions` for L&Y-style stability corrections
- Move `von_karman_constant` from `CoefficientBasedFluxes` into `LargeYeagerTransferCoefficients`
- Export `SimilarityScales` as the public container for transfer coefficients

## New API

```julia
# Constant coefficients
CoefficientBasedFluxes(transfer_coefficients = SimilarityScales(2e-3, 1e-3, 1e-3))

# Polynomial drag with constant scalar coefficients
CoefficientBasedFluxes(transfer_coefficients = SimilarityScales(PolynomialNeutralDragCoefficient(), 1e-3, 1e-3))

# Full Large & Yeager stability-corrected algorithm
CoefficientBasedFluxes(transfer_coefficients = LargeYeagerTransferCoefficients(),
                       solver_stop_criteria = FixedIterations(5))
```

## References

- Large, W.G. & Yeager, S.G. (2004): NCAR/TN-460+STR
- Large, W.G. & Yeager, S.G. (2009): Climate Dynamics, 33, 341-364

## Test plan

- [x] Unit tests for `PolynomialNeutralDragCoefficient` (bounds, cap, monotonicity, Float32)
- [x] Unit tests for `LinearStableStabilityFunction` and `large_yeager_stability_functions`
- [x] Constructor tests for `LargeYeagerTransferCoefficients`
- [x] Integration tests: constant, polynomial, and L&Y coefficients produce finite fluxes
- [x] CI passes


src/EarthSystemModels/InterfaceComputations/transfer_coefficient_fluxes.jl
```diff
@@ -0,0 +1,243 @@
...
+
+Neutral 10-m drag coefficient from the Large & Yeager (2004) polynomial
+(Table 1):
+
+    CdN = (a/U + b + c·U - d·U⁶) × 10⁻³   for U < high_wind_speed_threshold
+    CdN = high_wind_drag_coefficient                              otherwise
+
+Called as a functor: `CdN = drag(U)`.
+"""
+struct WindDependentNeutralDrag{FT}
```
glwagner: the name is too general

src/EarthSystemModels/InterfaceComputations/transfer_coefficient_fluxes.jl
```diff
@@ -0,0 +1,243 @@
...
+- Stability corrections via reusable `SimilarityScales` (default: Paulson unstable + linear stable)
+- 5 fixed iterations (no convergence check)
+- Wind speed floor at 0.5 m/s
+
+References:
+- Large, W.G. & Yeager, S.G. (2004): NCAR/TN-460+STR
+- Large, W.G. & Yeager, S.G. (2009): Climate Dynamics, 33, 341-364
+- AeroBulk reference: `mod_blk_ncar.f90` (Brodeau et al.)
+"""
+struct TransferCoefficientFluxes{FT, D, SF, S}
```
glwagner: we have `CoefficientBasedFluxes`. this name does not really distinguish between the two

src/EarthSystemModels/InterfaceComputations/transfer_coefficient_fluxes.jl
glwagner: `PolynomialNeutralDragCoefficient`

simone-silvestri:
Initially everything was prefixed by `NCAR` but I think it is not very nice to have it in a name. I ll think a bit more at the names

glwagner:
There is a simpler, non-iterative way to implement this. See here: https://github.com/NumericalEarth/Breeze.jl/blob/main/src/BoundaryConditions/polynomial_bulk_coefficient.jl

simone-silvestri:
Interesting. This is a regression to approximate the iterations looks like? We can also add this formulation to the model, that fits very cleanly with the `CoefficientBasedFluxes`, just need to make the coefficients polynomial. 

For OMIP compliance we need the iterative fluxes unfortunately to be exactly equal to the OMIP2 procedure. I would like to make sure the biases are not because of different forcing. I think I can add this whole machinery by just making some changes to generalize the `CoefficientBasedFluxes` which becomes cleaner.

simone-silvestri:
Ok, this might be better now.

glwagner:
> Interesting. This is a regression to approximate the iterations looks like? We can also add this formulation to the model, that fits very cleanly with the `CoefficientBasedFluxes`, just need to make the coefficients polynomial.

It's a regression to the bulk Ri and stability parameter. If the stability parameter is known then the fluxes can be evaluated directly with no iteration.

I guess it is simply an approximation solution to the flux inversion. It's fairly accurate as demonstrated by that paper.

The question is basically how many flux models we need? If we want something that uses polynomials approximations, it might make sense to go all the way to avoiding iteration. But I understand if you really need LY09 exactly.

simone-silvestri:
I agree that maybe that formulation is probably better since it is a single sweep. I can test the difference and see if we can just keep the simple one-shot formulation. I have restructured the PR to use `CoefficientBasedFluxes` so removing the L&Y formulation should be a matter of only deleting one method (`evaluate_coefficients(::LargeYeagerTransferCoefficients`)).

glwagner:
> I agree that maybe that formulation is probably better since it is a single sweep. I can test the difference and see if we can just keep the simple one-shot formulation. I have restructured the PR to use `CoefficientBasedFluxes` so removing the L&Y formulation should be a matter of only deleting one method (`evaluate_coefficients(::LargeYeagerTransferCoefficients`)).

oh neat

simone-silvestri:
These new fluxes can be seen in action here https://numericalearth.github.io/NumericalEarthDocumentation/previews/PR162/interface_fluxes/
In particular:
image

docs/src/interface_fluxes.md
```diff
@@ -137,9 +137,31 @@ A comprehensive example is given below, but we note briefly here that
...
+```@example interface_fluxes
+using NumericalEarth.EarthSystemModels.InterfaceComputations: PolynomialNeutralDragCoefficient
+
+poly_drag = PolynomialNeutralDragCoefficient()
+poly_fluxes = CoefficientBasedFluxes(transfer_coefficients = (poly_drag, 1e-3, 1e-3))
+```
+
+For the full Large & Yeager (2004) bulk algorithm with stability corrections,
+use `LargeYeagerTransferCoefficients`. This computes all three transfer coefficients
+(drag, Stanton, Dalton) from the neutral drag polynomial with Monin-Obukhov
```
glwagner:
```suggestion
(drag, temperature, moisture) from the neutral drag polynomial with Monin-Obukhov
```

is that right? I've never heard of the "Stanton" or "Dalton" coefficient, but I don't think we need jargon

src/EarthSystemModels/InterfaceComputations/coefficient_based_turbulent_fluxes.jl
```diff
@@ -1,71 +1,210 @@
...
+- Large, W.G. & Yeager, S.G. (2009): Climate Dynamics, 33, 341-364
+"""
+struct LargeYeagerTransferCoefficients{FT, D, SF}
+    von_karman_constant :: FT
+    neutral_drag_coefficient :: D
+    stability_functions :: SF
+    reference_height :: FT
+    stanton_stable :: FT
+    stanton_unstable :: FT
+    dalton :: FT
```
glwagner: lets use normal words not "stanton" and "dalton" whatever those mean!

glwagner (approved):
This looks good except that I don't like "Stanton" and "Dalton". Better to call these things with normal names. We originally had "heat flux coefficient" and "vapor flux coefficient". Actually, I would amend "vapor flux" to "moisture flux" (the latter is more general).

glwagner:
> These new fluxes can be seen in action here https://numericalearth.github.io/NumericalEarthDocumentation/previews/PR162/interface_fluxes/ In particular: image

What's the difference between LY04 and LY09?

Project.toml
```diff
@@ -13,6 +13,8 @@ DataDeps = "124859b0-ceae-595e-8997-d05f6a7a8dfe"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
+Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+DocumenterCitations = "daee34ce-89f3-4625-b898-19384cb65244"
```
navidcy: are we sure we need Documenter and DocumenterCitations as deps of the main package?

PR #164, opened 2026-04-15
Integrate ClimaSeaIce snow model

simone-silvestri:
## Summary
- `sea_ice_simulation`: add `with_snow`, `snow_conductivity
- Update imports for ClimaSeaIce snow-model branch (`SlabThermodynamics`, etc.)
- `default_ai_temperature` dispatches on `IceSnowConductiveFlux` when snow present
- New `flux_balance_temperature` dispatch using combined resistance R = hs/ks + hi/ki
- Refactored into shared `conductive_flux_balance_temperature` (DRY)
- Snowfall routing: interpolate JRA55 snow separately → `Jˢⁿ` field → sea ice `top_fluxes.snowfall`
- Sea ice exchanger state: rename `h` → `hi`, add `hs`

Requires ClimaSeaIce >= 0.5.0 (snow-model branch merged)

## Test plan
- [x] Run OMIP with `with_snow=true`, verify snow accumulates
- [x] Verify conductive flux reduces with snow (insulation effect)
- [x] Check surface temperature solve converges with IceSnowConductiveFlux

PR #165, opened 2026-04-15
Partition ocean freshwater: rain through ice, snow to open water

simone-silvestri:
## Summary
Split the ocean freshwater budget:
- **Rain** reaches the ocean everywhere (runs through cracks in ice)
- **Snow** only reaches the ocean through the open-water fraction (1-ℵ);
  snow on ice is routed to the sea ice model as snowfall
- **Evaporation** is from the open-water fraction (1-ℵ)

The total freshwater flux `ΣFao` now has ice-partitioning built in,
so `Jˢao = -S × ΣFao` no longer needs additional `(1-ℵ)` scaling.

Depends on #164

## Test plan
- [ ] Verify freshwater budget closes (rain + snow + evap + ice-ocean = total)
- [ ] Check salinity tendencies have correct sign for rain/evap/melt


simone-silvestri:
I will close this PR as it is absorbed in #164 that already splits rain from snow fluxes

PR #167, opened 2026-04-16
Adjust kwargs and imports in `sea_ice_simulation` to align with `ClimaSeaIce` changes

taimoorsohail:
As title suggests, this is to ensure compatibility with ClimaSeaIce#main

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -52,7 +52,7 @@ NumericalEarthWOAExt = "WorldOceanAtlasTools"
 
 [compat]
 Adapt = "4"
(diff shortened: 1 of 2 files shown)


simone-silvestri:
Ok this looks good, then we can integrate with the snow model in #164

simone-silvestri (approved):

simone-silvestri:
Forget the test errors, they will be solved after breeze bumps the version https://github.com/NumericalEarth/Breeze.jl/pull/632

PR #169, opened 2026-04-17
(0.4.2) Unify JRA55 onto the generic dataset backend, extend region support across all paths

### LINKED ISSUE #18: Unify JRA55 with all the other datasets
at the moment JRA55 datasets are a bit "special" because they have their own `FieldTimeSeries` and `JRA55NetCDFBackend` methods which are different from all the other datasets.
I think we can unify the datasets to use only one `FieldTimeSeries` method (there should be no difference between JRA55 and the other methods).

simone-silvestri:
## Summary

Two threads, woven together:

1. **JRA55 now uses the same code path as every other dataset.** The bespoke `JRA55NetCDFBackend` and `JRA55FieldTimeSeries` are gone. JRA55 atmospheric variables flow through `Field(::Metadatum)` and `FieldTimeSeries(::Metadata)` like ECCO, EN4, WOA, GLORYS, ERA5, ORCA.

2. **`BoundingBox` and `Column` regions now work everywhere.** These types existed on `main`, but Column extraction was a special-cased branch in `Field(::Metadatum)`, regions weren't supported on `FieldTimeSeries` at all, and JRA55 ignored them entirely. After this PR, regions work uniformly across `Field`, `FieldTimeSeries`, every dataset including JRA55, on CPU and GPU.

Closes #18.

## What changes for users

Old JRA55 entry points are gone:

```julia
# Before
ua = JRA55FieldTimeSeries(:eastward_velocity, arch; backend=JRA55NetCDFBackend(40))
atmosphere = JRA55PrescribedAtmosphere(arch; backend=JRA55NetCDFBackend(40))
```

New, same shape as every other dataset:

```julia
ua = FieldTimeSeries(Metadata(:eastward_velocity; dataset=RepeatYearJRA55()), arch;
                     time_indices_in_memory=40)
atmosphere = JRA55PrescribedAtmosphere(arch; time_indices_in_memory=40)
```

Regions on any dataset, including JRA55:

```julia
# Clip JRA55 to a region
bbox = BoundingBox(longitude=(-30, 30), latitude=(-60, 60))
ua_clipped = FieldTimeSeries(Metadata(:eastward_velocity; dataset=RepeatYearJRA55(),
                                       region=bbox))

# Extract a single Column from ECCO with bilinear interpolation
col = Column(12.0, -50.0; interpolation=Linear())
md  = Metadatum(:temperature; dataset=ECCO4Monthly(), date=..., region=col)
field = Field(md)   # 1×1×Nz Flat/Flat/Bounded
```

The `InMemory()` backend for JRA55 is dropped; if you want the whole series in memory, pass `time_indices_in_memory = length(metadata)`.

## What was already on main, what we added

**Already there:**
- `BoundingBox` and `Column` region types (`src/DataWrangling/metadata.jl`).
- `construct_native_grid` dispatches for both region types.
- `Field(::Metadatum)` accepting Column metadata via a separate `column_field_from_file` branch.
- ECCO velocity-axis "mangling" (`ShiftSouth` / `AverageNorthSouth`) for staggered files.

**What this PR adds:**
- A single GPU-friendly kernel `_set_region_kernel!` (`src/DataWrangling/set_region_data.jl`) that all read paths funnel through. It handles, in one pass:
  - `BoundingBox` offset (straight indexed read into the clipped target),
  - `Column` cyclic-aware bilinear/nearest blending,
  - NaN-aware corner handling so `_FillValue`/land cells don't poison the average,
  - unit conversion,
  - the staggered-velocity mangling (previously a separate path).
- `Field(::Metadatum)` rewritten on top of this kernel — the special `column_field_from_file` branch is gone; both regions go through `set_region_data!`.
- `FieldTimeSeries(::Metadata)` and the JRA55 `set!` paths also call `set_region_data!`, so regions work for time series and for JRA55 (neither did before).
- Cyclic-aware coordinate bracketing (`bracket_with_weight` with optional `period`) so a Column at 359.5° wraps correctly across the periodic seam.

## Smaller cleanups along the way

- Dead code removed from the bounding-box helpers (a half-dozen unused dispatches that simplified down to one each after the refactor).
- Double native-grid rebuild in `FieldTimeSeries(::Metadata, grid)` collapsed into a single architecture-aware comparison.
- JRA55 multi-year calendar handling: verified by inspecting a real file that the time axis uses the standard Gregorian calendar (leap days included), so the previous `DateTimeNoLeap` workaround was unnecessary; the bespoke component-matching helper is gone, replaced with a plain `findfirst`. Function name and surrounding comments updated to match reality.
- New tests: `test_column_field.jl`, `test_dataset_region.jl`, `test_mangling.jl`, `test_jra55_region.jl`, plus additions to `test_metadata.jl`.

## Test plan

- [x] `test/test_jra55.jl`, `test/test_jra55_region.jl` — JRA55 on the new path, with and without regions
- [x] `test/test_column_field.jl` — Column extraction (Linear/Nearest, cyclic wrap, NaN-aware blend)
- [x] `test/test_dataset_region.jl` — BoundingBox across datasets
- [x] `test/test_mangling.jl` — ECCO `v_velocity` mangling end-to-end
- [x] `test/test_ocean_only_model.jl`, `test/test_ocean_sea_ice_model.jl` — migrated to the new atmosphere signature
- [x] `test/test_checkpointer.jl` — JRA55 atmosphere survives JLD2 round-trip on the new path


src/DataWrangling/prefetching_backend.jl
```diff
@@ -0,0 +1,123 @@
...
+# the spawn cooperatively-scheduled and the optimisation a no-op.
+
+using Oceananigans.OutputReaders: AbstractInMemoryBackend, FlavorOfFTS, FieldTimeSeries, time_index
+using Oceananigans.Fields: location
+
+import Oceananigans.OutputReaders: new_backend
+import Oceananigans.Fields: set!
+
+mutable struct PrefetchingBackend{B<:DatasetBackend, F<:FieldTimeSeries} <: AbstractInMemoryBackend{Int}
+    inner :: B
```
glwagner:
```suggestion
    backend :: B
```

this is the backend right?

src/DataWrangling/prefetching_backend.jl
```diff
@@ -0,0 +1,123 @@
...
+    PrefetchingBackend{typeof(inner), typeof(buffer_fts)}(inner, nothing, buffer_fts, 0)
+
+# `:buffer_fts` deliberately omitted — see race invariant in preamble.
+function Base.getproperty(p::PrefetchingBackend, name::Symbol)
+    if name in (:inner, :pending, :next_start)
+        return getfield(p, name)
+    else
+        return getproperty(getfield(p, :inner), name)
+    end
+end
```
glwagner: can this be avoided?

glwagner:
Heh, nice idea!

Just to explore designs, would an alternative implementation be to add a `prefetching` property to `FieldTimeSeries`?

This would avoid the wrapper on backend and could allow us to make this a default without materialization perhaps.

src/DataWrangling/prefetching_backend.jl
simone-silvestri: Happy to rename it

simone-silvestri:
> Just to explore designs, would an alternative implementation be to add a `prefetching` property to `FieldTimeSeries`?

Nice idea. We could implement a PR in Oceananigans, FieldTimeSeries would need to carry the prefetch state that would be hooked in into `update_field_time_series!`. Shall we land this here so we test it and I add a PR in Oceananigans? We probably need some guards since this is a feature that users should not tamper with in postprocessing and when analyzing outputs.

src/DataWrangling/prefetching_backend.jl
simone-silvestri: we could do it by having accessors functions in oceananigans (`start(backend)`, `length(backend)` and so on)

simone-silvestri:
As a part of this PR I will temporarily disable the `single_column_os_papa_simulation.jl` since it builds a JRA55PrescribedAtmosphere on a single column given latitude and longitude. This feature is discontinued in this PR, the way to do it is to prescribe a region to the metadata after #142

glwagner:
> > Just to explore designs, would an alternative implementation be to add a `prefetching` property to `FieldTimeSeries`?
> 
> Nice idea. We could implement a PR in Oceananigans, FieldTimeSeries would need to carry the prefetch state that would be hooked in into `update_field_time_series!`. Shall we land this here so we test it and I add a PR in Oceananigans? We probably need some guards since this is a feature that users should not tamper with in postprocessing and when analyzing outputs.

Why would we land this PR, if we plan to revert it later and implement a `prefetching` property in `FieldTimeSeries`?

simone-silvestri:
I think it is a good idea to include it here. Note it is a very small part of this PR, just a file and a test. I think we are not still sure how to put it in `FieldTimeSeries` and exposing it to users, so it is something we might want to check and discuss. However, we are sure we want it here (hidden by users) and it is a very simple implementation that can be reverted in a very simple way once the design lands in Oceananigans.

I found out that it always has been very useful to start with a very simple and focused prototype that can be tested to have a better idea of the usefulness and design of a more general implementation. It might be in the end that we don't find it helpful to have in Oceananigans in the long run.

simone-silvestri:
Let me split it from this PR so we can figure out what to do with it

glwagner:
I don't think the implementation is any different, it just avoids the wrapper, right?

simone-silvestri:
Let me give it a shot. I think we can make it similar yes. This would work only for `PartlyInMemory` field time series, so we could add the property to the backend in some way, not to have hanging properties in FieldTimeSeries.

simone-silvestri:
I think what I want to achieve with this design in terms of UI (in Oceananigans) is:
- generalizability (what works for `InMemory` should also work out of the box for `DatasetBackend` and other `AbstractInMemoryBackend`s without code changes)
- no changes to `FieldTimeSeries` itself. I think the `prefetch=true` should probably go in `InMemory(; prefetch=true)` since it would be a throw-away kwarg for `FieldTimeSeries` if we use backends other than `AbstractInMemoryBackend`.

So actually, the wrapper might be the best solution to have both requirements satisfied. I am trying other avenues but I couldn't find other that are as clean.

simone-silvestri:
Ok I have split out the prefetch in https://github.com/CliMA/Oceananigans.jl/pull/5536

simone-silvestri:
Ok, I think this PR is ready for another review, it just unifies JRA55 with the other datasets the prefetch ability has been moved to an Oceananigans PR

simone-silvestri:
@glwagner, this might be ready to merge

src/DataWrangling/ECCO/ECCO_atmosphere.jl
```diff
@@ -31,26 +31,19 @@ function ECCOPrescribedAtmosphere(architecture = CPU(), FT = Float32;
...
-    va = FieldTimeSeries(va_meta, architecture; kw...)
-    Ta = FieldTimeSeries(Ta_meta, architecture; kw...)
-    qa = FieldTimeSeries(qa_meta, architecture; kw...)
-    pa = FieldTimeSeries(pa_meta, architecture; kw...)
-    ℐꜜˡʷ = FieldTimeSeries(ℐꜜˡʷ_meta, architecture; kw...)
-    ℐꜜˢʷ = FieldTimeSeries(ℐꜜˢʷ_meta, architecture; kw...)
-    Fr = FieldTimeSeries(Fr_meta, architecture; kw...)
+    ecco_fts(name) = FieldTimeSeries(Metadata(name; dataset, start_date, end_date, dir), architecture; kw...)
+
+    ua   = ecco_fts(:eastward_wind)
```
glwagner: shouldn't it be possible to define `ECCOFieldTimeSeries` with a proper alias? It would read so much better.

src/DataWrangling/JRA55/JRA55_prescribed_atmosphere.jl
```diff
@@ -1,56 +1,61 @@
...
-    # Remember that rivers and icebergs are on a different grid and have
-    # a different frequency than the rest of the JRA55 data. We use `PrescribedAtmospheres`
-    # "auxiliary_freshwater_flux" feature to represent them.
+    # Rivers and icebergs are on a different grid and have a different
+    # frequency than the rest of the JRA55 data. We use the
+    # PrescribedAtmosphere `auxiliary_freshwater_flux` feature for them.
     if include_rivers_and_icebergs
-        Fri = JRA55FieldTimeSeries(:river_freshwater_flux, architecture;   kw...)
-        Fic = JRA55FieldTimeSeries(:iceberg_freshwater_flux, architecture; kw...)
+        Fri = jra55_fts(:river_freshwater_flux)
```
glwagner: can we use `JRA55FieldTimeSeries` here? It reads to much better. I don't like the idea of coming up with a new name creatively for every dataset. I also don't like using "fts" abbreviation in a constructor.

glwagner (commented):
A few complaints...

- constructors like `blah_blah_fts` don't read well in my opinion. We could use `blah_blah_field_time_series`. But much better would be `BlahBlahFieldTimeSeries`. I think this is much more obvious

- `time_indices_in_memory` is a pretty awkward kwarg but maybe we are stuck with it

src/DataWrangling/JRA55/JRA55_prescribed_atmosphere.jl
simone-silvestri: done

src/DataWrangling/ECCO/ECCO_atmosphere.jl
simone-silvestri: done

simone-silvestri:
@glwagner this is ready to merge

simone-silvestri:
This PR is undergoing a lot of conflicts, I would need a review.
If it is accepted I resolve the conflicts correctly and merge

examples/generate_surface_fluxes.jl
```diff
@@ -45,9 +45,9 @@ save("ECCO_continents.png", fig)
 # - downwelling longwave radiation
 #
 # We load in memory only the first two time indices, corresponding to January 1st
-# (at 00:00 AM and 03:00 AM), by using `JRA55NetCDFBackend(2)`.
+# (at 00:00 AM and 03:00 AM), by using `time_indices_in_memory = 2`.
 
-atmosphere = JRA55PrescribedAtmosphere(; backend = JRA55NetCDFBackend(2))
+atmosphere = JRA55PrescribedAtmosphere(; time_indices_in_memory = 2)
```
glwagner, line 50: why this change?

src/DataWrangling/ECCO/ECCO.jl
```diff
@@ -113,6 +114,13 @@ longitude_interfaces(::ECCODataset) = (0, 360)
 longitude_interfaces(::ECCO4Monthly) = (-180, 180)
 latitude_interfaces(::ECCODataset) = (-90, 90)
 
+longitude_name(::Metadata{<:ECCODataset})        = "LONGITUDE_T"
+latitude_name(::Metadata{<:ECCODataset})         = "LATITUDE_T"
+longitude_name(::Metadata{<:ECCO4Monthly})       = "longitude"
+latitude_name(::Metadata{<:ECCO4Monthly})        = "latitude"
+longitude_name(::Metadata{<:ECCO4DarwinMonthly}) = "longitude"
+latitude_name(::Metadata{<:ECCO4DarwinMonthly})  = "latitude"
```
glwagner, line 121: what is this needed for?

src/DataWrangling/ECCO/ECCO_atmosphere.jl
```diff
@@ -31,26 +31,19 @@ function ECCOPrescribedAtmosphere(architecture = CPU(), FT = Float32;
...
 
-    ua = FieldTimeSeries(ua_meta, architecture; kw...)
-    va = FieldTimeSeries(va_meta, architecture; kw...)
-    Ta = FieldTimeSeries(Ta_meta, architecture; kw...)
-    qa = FieldTimeSeries(qa_meta, architecture; kw...)
-    pa = FieldTimeSeries(pa_meta, architecture; kw...)
-    ℐꜜˡʷ = FieldTimeSeries(ℐꜜˡʷ_meta, architecture; kw...)
-    ℐꜜˢʷ = FieldTimeSeries(ℐꜜˢʷ_meta, architecture; kw...)
-    Fr = FieldTimeSeries(Fr_meta, architecture; kw...)
+    ECCOFieldTimeSeries(name) = FieldTimeSeries(Metadata(name; dataset, start_date, end_date, dir), architecture; kw...)
```
glwagner: `ECCOFieldTimeSeries` needs to be an alias... is that possible? It shouldn't be a closure defined inside this prescribed atmosphere constructor

src/DataWrangling/JRA55/JRA55_field_time_series.jl
```diff
@@ -1,467 +1,123 @@
...
-    #
-    # then for example, we should find that (iᵢ, i₂) = (1, 5).
-    # So we want to reduce the first index by one, and limit them
-    # both by the available data. There could be some mismatch due
-    # to the use of different coordinate systems (ie whether λ ∈ (0, 360)
-    # which we may also need to handle separately.
-    i₁ = searchsortedfirst(hc, h₁)
-    i₂ = searchsortedfirst(hc, h₂)
-    i₁ = max(1, i₁ - 1)
-    i₂ = min(Nh, i₂)
```
glwagner, line 49: why delete this comment?

src/DataWrangling/JRA55/JRA55_field_time_series.jl
```diff
@@ -1,467 +1,123 @@
...
-        ds = Dataset(filepath)
-        data = ds[shortname][i₁:i₂, j₁:j₂, time_indices_in_memory]
-        close(ds)
-
-        copyto!(interior(fts, :, :, 1, :), data)
-        fill_halo_regions!(fts)
-
-        return fts
-    end
-end
```
glwagner, line 467: so do we lose the alias `JRA55FieldTimeSeries` ? I don't see where it is defined.

src/DataWrangling/JRA55/JRA55_prescribed_atmosphere.jl
```diff
@@ -1,44 +1,50 @@
...
-Return a [`PrescribedAtmosphere`](@ref) representing JRA55 reanalysis data.
-The atmospheric data will be held in `JRA55FieldTimeSeries` objects containing.
-For a detailed description of the keyword arguments, see the [`JRA55FieldTimeSeries`](@ref) constructor.
+Return a [`PrescribedAtmosphere`](@ref) representing JRA55 reanalysis data. Each atmospheric field is constructed via 
+`FieldTimeSeries(::JRA55Metadata)`, which uses a `DatasetBackend` parameterised by JRA55 metadata so that the JRA55-specific 
+`set!` (chunked-yearly NetCDF) is dispatched.
+The `region` keyword restricts the atmosphere to a sub-domain of the global JRA55 grid.
 """
-function JRA55PrescribedAtmosphere(architecture = CPU(), FT = Float32;
+function JRA55PrescribedAtmosphere(architecture = CPU();
```
glwagner, line 27: how do you control the `FT` of the data? Is it possible?

simone-silvestri: this PR is aligning JRA55 with the other datasets, so it is something we should implement at the `metadata_field_timeseries` level. It is probably missing

src/DataWrangling/JRA55/JRA55_field_time_series.jl
simone-silvestri: yeah, we use the `FieldTimeSeries(::metadata)` like for the other datasets

src/DataWrangling/ECCO/ECCO_atmosphere.jl
simone-silvestri: Should we have that? If we implement it we should have it for all the datasets. I would rather call `FieldTimeSeries(::Metadata)`

src/DataWrangling/ECCO/ECCO.jl
simone-silvestri: This is required to go fetch the coordinates of the data inside the dataset. Not all datasets have the same name for coordinates, so a getter allows us to generalize the constructor of the native grid.

examples/generate_surface_fluxes.jl
simone-silvestri: to align it with the other field time series datasets (and removed support for backends other than `DatasetBackend`)

src/DataWrangling/JRA55/JRA55_field_time_series.jl
glwagner: Does this preclude the definition of type aliases for specific `FieldTimeSeries`? This is possible if eg the metadata is a property of the backend

src/DataWrangling/ECCO/ECCO_atmosphere.jl
glwagner: if we don't have an alias, then we cannot use the name `ECCOFieldTimeSeries`.

examples/generate_surface_fluxes.jl
glwagner: ah ok. I feel that `time_indices_in_memory` is rather awkward 😂 but perhaps a problem for another day...

src/DataWrangling/dataset_backend.jl
```diff
@@ -0,0 +1,100 @@
...
+
+Type parameters `N` (on-native-grid) and `C` (cache-inpainted-data) are
+flags hoisted into the type so that dispatch and `Adapt.adapt_structure`
+can act on them without allocating.
+"""
+struct DatasetBackend{N, C, I, M} <: AbstractInMemoryBackend{Int}
+    start :: Int
+    length :: Int
+    inpainting :: I
+    metadata :: M
```
glwagner, line 32: since metadata is here, I believe that type aliases are possible. This makes sense for user-facing constructors.

src/DataWrangling/set_region_data.jl
```diff
@@ -0,0 +1,219 @@
...
+    h₂ = @allowscalar maximum(hg)
+    return h₁, h₂
+end
+
+# `ε` forgives Float32 to Float64 promotion noise so the slice doesn't lose a
+# cell at each end when grid centers are compared against file centers.
+function compute_bounding_indices(bounds::Tuple, hc)
+    h₁, h₂ = bounds
+    Nh = length(hc)
+    ε  = eps(Float32) * max(one(eltype(hc)), abs(h₁), abs(h₂))
```
glwagner, line 24:
```suggestion
    ε  = eps(Float32) * max(one(hc), abs(h₁), abs(h₂))
```

src/DataWrangling/set_region_data.jl
```diff
@@ -0,0 +1,219 @@
...
+
+#####
+##### Region-aware filling for Fields and FieldTimeSeries via a single kernel.
+#####
+##### `read_data(data, i, j, k, region_info, mangling)` is the only access point: it composes the
+##### file-axis offset (region) with the lat-axis remap (mangling). All region/mangling combinations
+##### go through one kernel that handles NaN + unit conversion in the same pass.
+#####
+
+struct BBoxOffset
```
glwagner: BoundingBoxOffset? (not sure if "BBox" is the same as "BoundingBox" here)

src/DataWrangling/set_region_data.jl
```diff
@@ -0,0 +1,219 @@
...
+end
+
+struct ColumnInfo{F, I}
+    i⁻ :: Int
+    i⁺ :: Int
+    j⁻ :: Int
+    j⁺ :: Int
+    wx :: F
+    wy :: F
+    ℑ  :: I
```
glwagner, line 81: comments about what these mean might help

src/DataWrangling/set_region_data.jl
```diff
@@ -0,0 +1,219 @@
...
+function infer_longitudinal_period(λc)
+    length(λc) < 2 && return nothing
+    Δ = λc[2] - λc[1]
+    span = λc[end] - λc[1] + Δ
+    return span ≈ 360 ? 360 : nothing
+end
+
+# Cyclic-aware bracketing. With `period`, the cell between `coords[end]` and `coords[1] + period` is the wrap cell: 
+# returns `(n, 1, w)` so the blend reads `data[n, …]` and `data[1, …]`.
+function bracket_with_weight(coords, x; period = nothing)
```
glwagner, line 124: does performance matter in this function?

src/DataWrangling/set_region_data.jl
```diff
@@ -0,0 +1,219 @@
...
+    w11 =      c.wx  *      c.wy  * !isnan(d11)
+    Σw  = w00 + w10 + w01 + w11
+    Σw == 0 && return convert(FT, NaN)
+    return (w00 * ifelse(isnan(d00), zero(FT), d00) +
+            w10 * ifelse(isnan(d10), zero(FT), d10) +
+            w01 * ifelse(isnan(d01), zero(FT), d01) +
+            w11 * ifelse(isnan(d11), zero(FT), d11)) / Σw
+end
+
+@inline function blend(data, c, k, mangling, ::Nearest, FT)
```
glwagner:
```suggestion
@inline function blend(::Nearest, data, c, k, mangling, FT)
```

same as above, per convention for dispatch.

src/DataWrangling/set_region_data.jl
```diff
@@ -0,0 +1,219 @@
...
+
+# `read_data(data, i, j, k, region, mangling, FT)` returns the file value at
+# the grid's (i, j, k) as `FT`, with `Missing` converted to NaN.
+@inline read_data(data, i, j, k, ::Nothing,     mangling, FT) = nan_convert_missing(FT, mangle(i, j, k, data, mangling))
+@inline read_data(data, i, j, k, b::BBoxOffset, mangling, FT) = nan_convert_missing(FT, mangle(i + b.di, j + b.dj, k, data, mangling))
+@inline read_data(data, _, _, k, c::ColumnInfo, mangling, FT) = blend(data, c, k, mangling, c.ℑ, FT)
+
+# NaN-aware bilinear blend: drop NaN corners and renormalise weights. 
+# Returns NaN only when all four corners are land.
+@inline function blend(data, c, k, mangling, ::Linear, FT)
```
glwagner: I'm not grasping why this stuff is being added here. Perhaps a comment on how this is used.

simone-silvestri:
Commit https://github.com/NumericalEarth/NumericalEarth.jl/pull/169/commits/93119b0f0b473458b0fe27613a2b85fb08b18fe1 should have address most of the code-changes comments:
  - `examples/inspect_JRA55_data.jl`: renamed `rht/rhn/axrh` to `qat/qan/axqa` and label to "Specific humidity (kg kg⁻¹)" so the plot matches the actual variable.
  - `src/DataWrangling/set_region_data.jl`: renamed `BBoxOffset` to `BoundingBoxOffset`.
  - `src/DataWrangling/set_region_data.jl`: documented each field of ColumnInfo inline (`i⁻/i⁺/j⁻/j⁺` indices in the file grid, `wx/wy` blend weights, `ℑ` interpolation kind) in a docstring.
  - `src/DataWrangling/set_region_data.jl`: moved the dispatch tag to the first argument in blend and updated the call site in `read_data` and the `Nearest` to `Linear` fallback.
  - `src/DataWrangling/set_region_data.jl`: expanded the comment above `blend(::Linear, …)` to explain why the NaN-aware logic exists.

I say we forget about the aliases for this PR such that the remaining items for the cleanup after this PR, which apply to all datasets are:
- find a better name than `time_indices_in_memory`
- create aliases for FTS for each dataset (and maybe also Fields)?
- control the floating type of the dataset in FTS and Field (maybe an additional type parameter to `Metadata`)

I can open an issue to track these further cleanup items.

src/DataWrangling/JRA55/JRA55_field_time_series.jl
simone-silvestri: I am ok with defining type aliases, but I think we should do it for all datasets if we decide to do it.

src/DataWrangling/ECCO/ECCO_atmosphere.jl
simone-silvestri: I ll change the name to not have type closures, I just liked how clean it looked.

examples/generate_surface_fluxes.jl
simone-silvestri: Yeah, it is probably not the best name, what I like though is that we have only one backend. Maybe we can find a better naming :)

src/DataWrangling/set_region_data.jl
simone-silvestri: should not be a problem, all the operations are rather simple, and this is a CPU-only function so there should be no problem of branch divergence with the if conditions as well

simone-silvestri:
Commit https://github.com/NumericalEarth/NumericalEarth.jl/pull/169/commits/b8d91615f786c4dd6cc9da1de50b59b91e6876ce removed the closures from the `*PrescribedAtmosphere` constructors and also removed the dead `FT` positional arguments from the `ECCOPrescribedAtmosphere` constructor as per 

> control the floating type of the dataset in FTS and Field (maybe an additional type parameter to Metadata)

glwagner:
Happy to forget about aliases, just make sure not to define functions like `ECCOFieldTimeSeries` in that case

simone-silvestri:
@glwagner @navidcy this PR should be ready to merge

examples/inspect_JRA55_data.jl
```diff
@@ -4,8 +4,11 @@ using Oceananigans
 using Oceananigans.Units
 using Printf
 
-Qswt = NumericalEarth.JRA55.JRA55FieldTimeSeries(:downwelling_shortwave_radiation)
-rht  = NumericalEarth.JRA55.JRA55FieldTimeSeries(:relative_humidity)
+using NumericalEarth.DataWrangling: Metadata
+using NumericalEarth.JRA55: RepeatYearJRA55
+
+Qswt = FieldTimeSeries(Metadata(:downwelling_shortwave_radiation; dataset=RepeatYearJRA55()); time_indices_in_memory=8)
+qat  = FieldTimeSeries(Metadata(:specific_humidity;               dataset=RepeatYearJRA55()); time_indices_in_memory=8)
```
glwagner, line 11: use new notation?

glwagner: actually forget it... there is a lot of notation that needs to be lceaned up, better for a dedicated PR I think

glwagner (approved):
what is the status of the JRA55FieldTimeSeries or jra55_field_time_series after this?

simone-silvestri:
it does not exist anymore, we need to call `FieldTimeSeries(metadatum)`. Initially I thought we could just have an alias, but I guess it is better to open a new PR where we can define all the aliases, because we will need to make a choice on which one is the "default" dataset associated with the type aliases (RepeatYear or MultiYear for example)

PR #171, opened 2026-04-20
Update Oceananigans version range in Project.toml

taimoorsohail:
Version bump to support new Oceananigans#main


simone-silvestri (approved):

simone-silvestri:
same as #170 :)

PR #172, opened 2026-04-20
Add distributed checkpoint reproducibility test

taimoorsohail:
## Summary
- add a new distributed checkpoint/restart reproducibility test in `test/test_distributed_utils.jl`
- mirror the serial checkpointer workflow: run to iteration 3, checkpoint, restore into a fresh simulation, then continue to iteration 6
- assert field agreement using the same tolerances as the serial test (`T,S,h`: `rtol=1e-13`; `u,v,ui,vi`: `rtol=1e-10`), and assert exact clock time/iteration equality

## Why this is needed in addition to serial tests
- serial checkpoint tests validate restart consistency for single-process execution, but they do not exercise MPI domain decomposition, collective reductions, or halo-exchange ordering
- distributed runs can introduce additional floating-point drift modes that do not appear in serial, so we need explicit multi-rank coverage to keep reproducibility-within-tolerance guarantees from regressing

## Implementation notes
- test uses an MPI-aware `Distributed(CPU(), partition=Partition(mpi_ranks, 1))`
- local checkpoint files are verified and cleaned up per rank
- test is skipped with a warning when only one MPI rank is active

with Codex


simone-silvestri:
I think to make this work we would need to uncomment the distributed tests in the CI.yml pipeline

taimoorsohail:
Ah! Didn't know they were commented. Is there a reason for that? No resources or too slow or something?

simone-silvestri:
When we have set up the pipeline we delayed those tests because they need rewriting to fix the GHA way of writing distributed tests. In practice we need to write down strings with the test script and run them with 
```
run($mpiexec() ...
```
Something like this https://github.com/CliMA/Oceananigans.jl/blob/main/test/test_mpi_tripolar.jl

simone-silvestri (approved):

test/runtests.jl
```diff
@@ -109,8 +111,10 @@ function __init__()
...
-runtests(NumericalEarth, args; testsuite)
-
-delete_inpainted_files(@get_scratch!("."))
+if group == "distributed_mpi"
+    include("test_distributed_utils.jl")
+else
+    __init__()
+    runtests(NumericalEarth, args; testsuite)
+    delete_inpainted_files(@get_scratch!("."))
+end
```
simone-silvestri:
```suggestion
__init__()
runtests(NumericalEarth, args; testsuite)
delete_inpainted_files(@get_scratch!("."))
```
We do not need to do this workaround. You can look at how the `test_downloading.jl`, which is run separately, works to have an independent test running

PR #178, opened 2026-04-21
Fix problems with GLORYS' `free_surface`

simone-silvestri:
free surface with GLORYS is a little buggy because it is assumed to be 3D.
This PR fixes it.

Found by @aklocker42


aklocker42 (approved):

simone-silvestri:
The GLORYS tests pass. The GPU test erroring is because of #181 in test_bathymetry unrelated to this PR.

Issue #181, opened 2026-04-22
GPU tests failing with OOM errors


simone-silvestri:
since a week or so, the GPU tests start failing randomly with OOM errors. Examples of this failure:
https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/24768246993/job/72467804322
https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/24667361659/job/72128408807

Maybe we are at the limit of what the GPU worker can handle? Should we increase the cache size or clean the cache before starting a job?

cc @awiteck @giordano

giordano:
Not sure how the cache would help. As I mentioned in https://github.com/NumericalEarth/NumericalEarth.jl/pull/179#issuecomment-4294777482, a possible strategy is to use fewer concurrent jobs, 7 instead of 8 should already help a lot, we had a similar situation in Breeze.jl

simone-silvestri:
Ah stupid me I thought that comment was about reading the crash output 😅 
I ll open a PR to reduce the number of jobs

PR #182, opened 2026-04-22
Reduce number of parallel jobs in GPU tests

simone-silvestri:
to clear issue #181

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,6 +131,12 @@ jobs:
       - name: df after cleanup
         run: |
           df -hT
+      - name: "Set test arguments"
+        shell: bash
+        run: |
+          # Run tests in verbose mode
+          TEST_ARGS=(--verbose)
+          TEST_ARGS+=(--jobs=$(($(nproc) - 1)))
       - name: Update registry
         shell: julia --color=yes {0}
         run: |
@@ -145,7 +151,7 @@ jobs:
               using Pkg;
               Pkg.test(; coverage=true,
                        julia_args=["--check-bounds=yes", "--compiled-modules=yes", "-O0"],
-                       test_args=["--verbose", "--jobs=$(($(nproc) - 1))"])
+                       test_args=`${{ env.runtest_test_args }}`])
           '
       - uses: julia-actions/julia-processcoverage@v1
       - uses: codecov/codecov-action@v5
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,6 +137,7 @@ jobs:
           # Run tests in verbose mode
           TEST_ARGS=(--verbose)
           TEST_ARGS+=(--jobs=$(($(nproc) - 1)))
+          echo "runtest_test_args=${TEST_ARGS[@]}" | tee -a "${GITHUB_ENV}"
       - name: Update registry
         shell: julia --color=yes {0}
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -145,7 +145,7 @@ jobs:
               using Pkg;
               Pkg.test(; coverage=true,
                        julia_args=["--check-bounds=yes", "--compiled-modules=yes", "-O0"],
-                       test_args=["--verbose"])
+                       test_args=["--verbose", "--jobs=$(($(nproc) - 1))"])
```
giordano: I suggest reducing the number of jobs only for the GPU tests, like in Breeze: they have many more CPUs than the others, and more at risk of oversubscribing the machine.

simone-silvestri: will this change affect also the CPU pipeline?

giordano: As it is right now, yes. As we do in Breeze where we decrease the number of jobs conditionally on the name of the "os", no.

simone-silvestri: looks like the CPU pipeline is still running with 4 workers though. Probably because here we have two different pipelines for CPU and GPU while in Breeze you have only one pipeline?

giordano: I don't think this is working correctly: https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/24772443470/job/72482256244?pr=182#step:12:901
```
Running 19 tests using 8 parallel jobs. If this is too many concurrent jobs, specify the `--jobs=N` argument to the tests, or set the `JULIA_CPU_THREADS` environment variable.
```
The GPU one is still running with 8 parallel jobs

.github/workflows/ci.yml
```diff
@@ -131,6 +131,12 @@ jobs:
       - name: df after cleanup
         run: |
           df -hT
+      - name: "Set test arguments"
+        shell: bash
+        run: |
+          # Run tests in verbose mode
+          TEST_ARGS=(--verbose)
+          TEST_ARGS+=(--jobs=$(($(nproc) - 1)))
```
giordano, line 139: [This](https://github.com/NumericalEarth/Breeze.jl/blob/30dc2d02a15e59c7567e2a22dc8f0111dd8d1761/.github/workflows/CI.yml#L118) is needed to actually make save the environment variable for the next steps
```suggestion
          TEST_ARGS=(--verbose)
          TEST_ARGS+=(--jobs=$(($(nproc) - 1)))
          echo "runtest_test_args=${TEST_ARGS[@]}" | tee -a "${GITHUB_ENV}"
```

But again, I'd recommend something like
```suggestion
          TEST_ARGS=(--verbose)
          if [[ ${{ matrix.os }} == *linux-nvidia-gpu* ]]; then
              # On the GPU runners we sometimes run out-of-memory because we have several
              # concurrent jobs, some of them reaching a large memory usage.  We reduce the
              # number of jobs to avoid memory issues.
              TEST_ARGS+=(--jobs=$(($(nproc) - 1)))
          fi
          echo "runtest_test_args=${TEST_ARGS[@]}" | tee -a "${GITHUB_ENV}"
```

simone-silvestri: maybe I am not understanding, but there is no `matrix.os` in the GPU jobs, there is only one in the CPU jobs:
https://github.com/NumericalEarth/NumericalEarth.jl/blob/9110798147a3e78bbbf07593abc3a9c09545e27e/.github/workflows/ci.yml#L96-L160

giordano: Aaaaah, right, these are different jobs!

simone-silvestri:
Ok this seems to have worked!
image

giordano:
From the screenshot you can also see what's the issue: with ~26 GB of total memory available, 8 concurrent workers have on average of 3.25 GB of memory each, with 7 concurrent workers the share raises to 3.7 GB, for likely little overall time change.

giordano (approved):

Issue #193, opened 2026-04-28
CPU CI takes over an hour and seems to overlap with data downloading


glwagner:
Currently the CPU CI takes a _very_ long time to run, eg:

https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/25079084434/job/73485468142?pr=149

It also seems to test data downloading (?) which overlaps with the data downloading GHA jobs.

Do we have an idea why GPU tests are much faster? Usually GPU takes longer because of longer compile; however in this case the CPU for CPU tests and CPU for GPU tests are very different (since compile happens on CPU).

We could consider reducing the CPU testing while keeping GPU more complete. Or maybe there are other strategies to speed up the CPU tests.

simone-silvestri:
I think one of the culprit of the CPU being slower is that we have 4 workers vs 8 workers on the GPU. It might be a good idea to remove some tests from the CPU given that GPU tests are typically more stringent. Maybe we can remove the slowest ones. Given these times
```
                                     │   Test   │   Init   │ Compile │ ──────────────── CPU ──────────────── │
Test                        (Worker) │ time (s) │ time (s) │   (%)   │ GC (s) │ GC % │ Alloc (MB) │ RSS (MB) │
test_diagnostics_2               (1) │        started at 2026-04-28T22:49:11.667
test_speedy_coupling             (3) │        started at 2026-04-28T22:49:12.781
test_sea_ice_ocean_heat_fluxes   (2) │        started at 2026-04-28T22:49:12.781
test_flux_and_restoring          (4) │        started at 2026-04-28T22:49:12.883
test_diagnostics_2               (1) │   130.97 │     4.59 │   93.71 │   4.93 │  3.8 │    6624.32 │  1813.54 │
test_ocean_only_model            (1) │        started at 2026-04-28T22:51:24.972
test_flux_and_restoring          (4) │   225.87 │     5.63 │   96.35 │   9.29 │  4.1 │   10039.14 │  1817.42 │
test_ecco4_en4                   (4) │        started at 2026-04-28T22:53:00.762
test_ecco4_en4                   (4) │     1.47 │     1.57 │   89.51 │   0.00 │  0.0 │      55.95 │  1817.42 │
test_surface_fluxes              (4) │        started at 2026-04-28T22:53:04.476
test_surface_fluxes              (4) │   223.46 │     1.09 │   97.09 │   4.58 │  2.1 │    9847.25 │  1850.53 │
[...]

PR #195, opened 2026-04-29
Add hdf5 version constraint in `install_veros` function

simone-silvestri:
Pin hdf5 version to prevent dependency issues. Apparently new versions of hdf5 are incompatible with veros. Sometimes th e tests pick up hdf5 v>=2 leading to test flakyness. Hopefully pinning to a version < 2 will solve test flakyness.


simone-silvestri:
@glwagner should we prune CPU tests as per #193 in this PR?

glwagner (approved):

glwagner:
> @glwagner should we prune CPU tests as per #193 in this PR?

sorry just saw this. lets use a different PR.

PR #197, opened 2026-04-30
Add support for downloading ERA5 pressure levels data product (new)

### LINKED ISSUE #88: ERA5 data wranglers only set up for single levels
To support work with Breeze, I would like to utilize these tools to retrieve 3D gridded data at pressure levels. Data are available between 1000 hPa and 1 hPa over 37 levels: https://cds.climate.copernicus.eu/datasets/reanalysis-era5-pressure-levels?tab=overview

ewquon:
This replaces https://github.com/NumericalEarth/NumericalEarth.jl/pull/93. All previous comments and suggestions have already been addressed — the difference is that this new PR originates from this repo rather than my fork. The reason for opening a new PR is that the CI from my fork was failing due to empty credentials.

Closes https://github.com/NumericalEarth/NumericalEarth.jl/issues/88.


examples/ERA5_bounded_pressure_level_data.jl
```diff
@@ -0,0 +1,121 @@
...
+using Oceananigans
+
+using Statistics
+using CairoMakie
+
+selected_levels = filter(≥(250hPa), ERA5_all_pressure_levels) # select all levels below 250 hPa
+dataset = ERA5HourlyPressureLevels(selected_levels)
+
+dates = DateTime(2004, 12, 16):Hour(1):DateTime(2005, 01, 09)  # van Zanten et al 2011 study period
+#dates = DateTime(2004, 12, 16):Hour(1):DateTime(2004, 12, 23)  # shorter time range for demo
```
glwagner: remove commented code for merge

examples/ERA5_bounded_pressure_level_data.jl
```diff
@@ -0,0 +1,121 @@
...
+
+# Horizontal-mean profile for each snapshot → (Nz, Nt) arrays
+function horizontal_mean_profiles(series)
+    profiles = zeros(Nz, Nt)
+    for n in 1:Nt
+        data = interior(series[n], :, :, :)
+        profiles[:, n] = dropdims(mean(data, dims=(1, 2)), dims=(1, 2))
+    end
+    return profiles
+end
```
glwagner, line 72: What happens if we write

```
profile = mean(data, dims=(1, 2))
```

?

examples/ERA5_bounded_pressure_level_data.jl
ewquon: 🫡

examples/ERA5_winds_and_stokes_drift.jl
```diff
@@ -42,10 +42,11 @@ v_wind_meta   = Metadatum(:northward_velocity;     dataset, date)
 # We build a single `LatitudeLongitudeGrid` and use `set!` to download
 # and interpolate all four variables onto it.
 
-grid = LatitudeLongitudeGrid(size = (1440, 721, 1),
+grid = LatitudeLongitudeGrid(size = (1440, 720, 1),
                              longitude = (0, 360),
                              latitude = (-90, 90),
                              z = (0, 1))
+@show grid
```
glwagner: remove show from example for merge

glwagner (commented):
Should we combine the ERA5 data demos into a single example? And add to docs? Generally we should not have examples that aren't in the docs (I believe there are still some orphaned right now, but we need to clean that up)

src/DataWrangling/DataWrangling.jl
```diff
@@ -242,4 +242,17 @@ using .WOA
 using .JRA55
 using .OSPapa
 
+# Fallback: if no download extension is loaded, check that all files already exist
+function download_dataset(metadata::Metadata)
```
glwagner, line 246: Isn't it better to throw an error?

I think if someone introduces `Metadata` which cannot be downloaded, they should still extend `download_dataset`.

(Arguably, we should not have Metadata in public repos that cannot be downloaded -- it's sort of defeats the point of using it since downloading / fetching data is typically the hardest step)

src/DataWrangling/metadata_field.jl
```diff
@@ -389,9 +409,12 @@ function set_metadata_field!(field, data, metadatum)
     arch = architecture(grid)
 
     Nx, Ny, Nz = size(metadatum)
+
     mangling = if size(data, 2) == Ny-1
+        @info "Shifting field southward"
         ShiftSouth()
     elseif size(data, 2) == Ny+1
+        @info "Averaging field in north-south dir"
```
glwagner: Let's either remove these or turn into debug. We try to have all info statements behind "verbose" flags, generally its poor form to be overly chatty.

ewquon: Personally, I like chatty — it lets me know the code is doing something! And it makes my log files more useful. Will change `@info` -> `@debug`

examples/ERA5_bounded_pressure_level_data.jl
ewquon: The expected result... silly Claude 😅

src/DataWrangling/DataWrangling.jl
ewquon: This does throw an error and is slightly more user-friendly.
```julia
using NumericalEarth
using NumericalEarth.DataWrangling: download_dataset
using NumericalEarth.DataWrangling.ERA5
using Dates

meta = Metadatum(:temperature;
                 dataset = ERA5HourlySingleLevel(),
                 region = Column(-61.5, 17.5),
                 date = DateTime(2004, 12, 16))
download_dataset(meta)

# ERROR: LoadError: No download method is available (is the backend package loaded?) and 1 data file(s) are missing. First missing: /Users/eliot/.julia/scratchspaces/904d977b-046a-4731-8b86-9235c0d1ef02/ERA5/2m_temperature_ERA5HourlySingleLevel_2004-12-16T00_-61.5_-61.5_17.5_17.5.nc
```

If the user loaded `using CDSAPI`, then the download proceeds as expected.

glwagner: I'm worried that it does not error if the data is present, but maybe I am reading it wrong

glwagner: This allows you to build scripts that work on one machine but not another, which we want to avoid.

src/DataWrangling/DataWrangling.jl
```diff
@@ -242,4 +242,17 @@ using .WOA
...
+    paths = metadata_path(metadata)
+    paths isa AbstractString && (paths = [paths])
+    missing_files = filter(!isfile, paths)
+    if !isempty(missing_files)
+        n = length(missing_files)
+        error("No download method is available (is the backend package loaded?) " *
+              "and $n data file(s) are missing. First missing: $(first(missing_files))")
+    end
+    return paths
```
glwagner:
```suggestion
    error("No download method for $metadata is available (is the backend package loaded?)")
```

src/DataWrangling/DataWrangling.jl
ewquon: Not erroring with the data present was actually the design.

glwagner (approved):

PR #198, opened 2026-04-30
Reduce burden on the CPU testing suite

simone-silvestri:
The heaviest tests blow up the testing time on the CPU job.
These tests, however are also repeated on (faster) GPU workers. To speed up the testing infrastructure we keep the heaviest tests only on the GPU which is a more stringent testing hardware


simone-silvestri:
tests go from 1hr16 to 47m now sure if it is a great improvement

glwagner (approved):
I think this is a necessary evil. But perhaps should go further. Are there any tests that can be moved to data downloading?

glwagner:
I think this closes an issue right?

simone-silvestri:
deleted also `test_orca_grid` which was the largest after the already removed ones.

PR #199, opened 2026-04-30
Restores ECCO Darwin and allows `DatasetRestoring` to act on fields by different name

### LINKED ISSUE #61: ECCO Darwin units
I don't think that this is correct:
https://github.com/NumericalEarth/NumericalEarth.jl/blob/4e27347c2cfc96efca94cfeb419e99e3434aa92b/src/DataWrangling/ECCO/ECCO_darwin.jl#L78
because the average nitrate is 0.019, I think they're actually in mmol/L because they should be about 1000x larger?

Image

jagoosw:
This PR:
- fixes ECCO Darwin data loading (broken in the last few releases),
- moves `variable_name` to be a `kwarg` in `DatasetRestoring` to allow users to override the default naming.,
- changes the units of the Darwin tracers as they were documented incorrectly

Closes #61 and #324


simone-silvestri:
I don't think we can pass a varname that is different than the one in metadata, while this keyword argument would allow it. I would probably drop the variable_name kwarg, what you want to achieve can be done just by passing `field_name` am I right?

src/DataWrangling/restoring.jl
```diff
@@ -197,7 +197,9 @@ function DatasetRestoring(metadata::Metadata,
                           time_indices_in_memory = default_time_indices_in_memory(metadata),
                           time_indexing = Cyclical(),
                           inpainting = NearestNeighborInpainting(Inf),
-                          cache_inpainted_data = true)
+                          cache_inpainted_data = true,
+						  variable_name = metadata.name,
+						  field_name = oceananigans_fieldnames[variable_name])
```
simone-silvestri:
```suggestion
                          cache_inpainted_data = true,
						  field_name = oceananigans_fieldnames[metadata.name])
```

jagoosw:
> I don't think we can pass a varname that is different than the one in metadata, while this keyword argument would allow it. I would probably drop the variable_name kwarg, what you want to achieve can be done just by passing `field_name` am I right?

Oh yes thats true

simone-silvestri (approved):

jagoosw:
Could I add a scale factor to the dataset? Do you think it would be best to do it in the constructor, or to pass a scale factor to the forcing function?

simone-silvestri:
there is already the `rate` and `mask` kwargs. Would the scale be different?

jagoosw:
> there is already the `rate` and `mask` kwargs. Would the scale be different?

I mean if the dataset is different (e.g. in different units) to the field its restoring, so it would go in as s in $r  (s\psi_r - \psi)$

simone-silvestri:
We have a `conversion_units` and a `convert_units` to convert data units to the model required units. We could do it through there?

jagoosw:
> We have a `conversion_units` and a `convert_units` to convert data units to the model required units. We could do it through there?

Doesn't that do a fixed conversion depending on the dataset? Or is the a user way to change the conversion?

jagoosw:
Circling back to this issue: I think it would be good if users could pass a "conversion unit" to `Metadata` incase their model etc. is using different units. Its also a bit confusing what the "native" unit being converted to is so it might be more clear to e.g. write `MolPerLiterToMilliMolPerLiter` etc. even though its longer?

simone-silvestri:
The units system is a bit rough around the edges. @bgroenks96 proposed refactoring it using https://github.com/JuliaPhysics/Unitful.jl.
 
> I think it would be good if users could pass a "conversion unit" to Metadata incase their model etc. is using different units.

For physics quantities we probably do not want to change the units because the interface fluxes enforce a certain convention (temperature in kelvin, salinity in psu, velocities in m/s and so on). I can imagine that for tracers each model might want to have its own convention. We will need to enforce some type of consistency with what the flux solver sees though.

jagoosw:
> The units system is a bit rough around the edges. @bgroenks96 proposed refactoring it using https://github.com/JuliaPhysics/Unitful.jl.

I saw this which seems like a good idea but a lot of work.

> For physics quantities we probably do not want to change the units because the interface fluxes enforce a certain convention (temperature in kelvin, salinity in psu, velocities in m/s and so on). I can imagine that for tracers each model might want to have its own convention. We will need to enforce some type of consistency with what the flux solver sees though.

Yeah that makes sense r.e. physical quantities. People use different units for different BGC models (e.g. mol C vs mol N) so I think theres going to frequently be cases where you might want to change units for restoring, but thats going to make a challenge to enforce correct units when necessary for fluxes. Could I merge this PR and open a new one to think about that?

simone-silvestri (approved):

jagoosw:
I will open an issue to figure out why the tests didn't catch the ECCO Darwin issue but merging this for now

jagoosw:
This is working now, will merge when tests pass

src/DataWrangling/ECCO/ECCO_darwin.jl
```diff
@@ -108,6 +112,9 @@ DataWrangling.longitude_interfaces(::ECCO4DarwinMonthly) = (-180, 180)
     retrieve_data(metadata::Metadatum{<:ECCO4DarwinMonthly})
 
 Read a ECCO4DarwinMonthly data file and regrid using MeshArrays on to regular lat-lon grid
+
+Warning: interpolation from the LLC grid interpolates zeros so the values are incorrect
+near the boundary. In the future we should work out how to inpaint before.
```
simone-silvestri:
```suggestion
```

PR #205, opened 2026-05-01
Update ERA5 examples

ewquon:
Consolidate ERA5 examples with region subsetting into a single, coherent example that exercises `ERA5HourlySingleLevel` and `ERA5HourlyPressureLevels` datasets, as well as `BoundingBox` and `Column` regions.


ewquon:
To facilitate review, I've literated a preview HTML page attached here.

[NumericalEarth - Update ERA5 examples - DRAFT.zip](https://github.com/user-attachments/files/27281693/NumericalEarth.-.Update.ERA5.examples.-.DRAFT.zip)

ewquon:
Hmm, doesn't look like `examples/ERA5_atmosphere.jl` is being built — not sure why.

docs/make.jl
```diff
@@ -35,7 +35,8 @@ examples = [
     Example("Global climate simulation", "global_climate_simulation", false),
     Example("Veros ocean simulation", "veros_ocean_forced_simulation", false),
     Example("Breeze over two oceans", "breeze_over_two_oceans", false),
-    Example("ERA5 winds and Stokes drift", "ERA5_winds_and_stokes_drift", true),
+    Example("ERA5 surface winds and Stokes drift", "ERA5_winds_and_stokes_drift", true),
+    Example("ERA5 atmosphere", "ERA5_atmosphere", false),
```
glwagner: can we combine these two? What do you think @ewquon ? Just a "Using ERA5 data" tutorial?

glwagner:
@ewquon the preview should get generated if all goes well. If we want to generate the preview quickly we can comment out the heavy examples while we work on this PR. Also, ERA5 example could possibly be built every commit. Then we can avoid the tag "build all examples"

docs/make.jl
ewquon: Sure I could just stitch them together -- or just leave out the original example

PR #207, opened 2026-05-01
DataWrangling cleanup: standardize download_dataset returns, drop ERA5 dead code

### LINKED ISSUE #201: Inconsistent return types across `download_dataset` implementations
## Summary

`download_dataset` has ~18 method definitions across `src/DataWrangling/` and `ext/`, and they don't agree on a return value. Three different return types are in use, and there isn't a clear pattern (e.g. by single-date vs multi-date) that callers can rely on.

Surfaced while reviewing the API in #93.

### `String` (single file path) — 5 methods
- `src/DataWrangling/ETOPO/ETOPO.jl:58` — `download_dataset(::ETOPOMetadatum)`
- `src/DataWrangling/ORCA/ORCA.jl:110` — `download_dataset(::ORCAMetadatum)`
- `src/DataWrangling/OSPapa/OSPapa_ocean_observations.jl:69` — `download_dataset(::OSPapaMetadata)`
- `ext/NumericalEarthCopernicusMarineExt.jl:24` — `download_dataset(::GLORYSMetadatum; …)`
- `ext/NumericalEarthCDSAPIExt.jl:105` — `download_dataset(::ERA5Metadatum; …)`

### `Vector{String}` (multiple file paths) — 3 methods
- `src/DataWrangling/DataWrangling.jl:246` — generic fallback `download_dataset(::Metadata)`
- `ext/NumericalEarthCopernicusMarineExt.jl:16` — `download_dataset(::GLORYSMetadata; …)`
- `ext/NumericalEarthCDSAPIExt.jl:227` — `download_dataset(names, ::ERA5PressureMetadatum; …)`

### `nothing` — 8+ methods
- `src/DataWrangling/EN4/EN4.jl:207` — `download_dataset(::Metadata{<:EN4Monthly})`
- `src/DataWrangling/JRA55/JRA55_metadata.jl:192` — `download_dataset(::JRA55Metadata)`
- `src/DataWrangling/ECCO/ECCO.jl:304` — `download_dataset(::ECCOMetadata)`
- `src/DataWrangling/OSPapa/OSPapa_flux_observations.jl:97` — `download_dataset(::OSPapaFluxMetadata)`
- `ext/NumericalEarthWOAExt.jl:38` — `download_dataset(::Metadata{<:WOAClimatology}; …)`
- `ext/NumericalEarthCDSAPIExt.jl:138` — `download_dataset(::ERA5Metadata; …)` (implicit `nothing`)
- `ext/NumericalEarthCDSAPIExt.jl:214` — `download_dataset(names, ::ERA5PressureMetadata; …)`
- `ext/NumericalEarthCDSAPIExt.jl:309` — `download_dataset(names, ::ERA5Dataset, datetimes; …)`

### Delegating overloads (inherit one of the above)
- `ext/NumericalEarthCDSAPIExt.jl:291, 298, 327`

## Why it matters

1. **Within a single module.** `ext/NumericalEarthCDSAPIExt.jl` returns `String`, `Vector{String}`, or `nothing` depending on which overload you hit — even for the same dataset family. A caller has to read the source to know which one applies.

2. **Composability.** A consistent return of paths would let callers chain into `NCDataset`, `metadata_path` validators, etc. without re-deriving the path from the metadata.

## Suggested resolution

Standardize on **\"return the path(s) on disk\"**:
- Single-file methods → `String`
- Multi-file methods → `Vector{String}`
- Drop the `nothing` returns; replace with `return metadata_path(metadata)` (which already gives `String` for `Metadatum` and `Vector{String}` for `Metadata`).

This is mechanical (~8 lines, one per offending method), cheap, and makes the dispatch tree predictable. It's a behavior change in a public-ish API, so it should land as its own small PR (separate from feature work) and call out the change in the next release notes.

ewquon:
## Summary

- Standardize `download_dataset` return types across all ~17 methods: `String` for single-file `Metadatum`, `Vector{String}` for multi-file `Metadata`. Drops the previous mix of `String` / `Vector{String}` / `nothing` returns. Closes #201.
- Refactor ERA5 helpers (`download_era5_day`, `download_era5_multivar_day`) to return their already-computed path lists instead of `nothing`, and to construct paths via `joinpath(dir, metadata_filename(...))` directly rather than instantiating throwaway `Metadatum` objects.
- Remove dead code in the ERA5 module: unused `ERA5_single_level_accumulated_variables` Set, unused `Day` import, redundant `metadata_path` import, and 1-arg `metadata_prefix` overloads that duplicated the 4-arg form (tests updated to call 4-arg form directly).

## Test plan

- [ ] Existing CPU test suite passes
- [ ] `test_cds_downloading.jl` still exercises the ERA5 surface + pressure-level paths
- [ ] Spot-check that `download_dataset(...)` callers receive paths (previously discarded)

### DIFF
--- a/ext/NumericalEarthCDSAPIExt.jl
+++ b/ext/NumericalEarthCDSAPIExt.jl
@@ -139,12 +139,15 @@ function download_dataset(metadata::ERA5Metadata; skip_existing=true, cleanup=tr
     dates = metadata.dates isa AbstractVector ? metadata.dates : [metadata.dates]
     grouped = _group_by_calendar_day(dates)
 
+    paths = String[]
     for day in sort(collect(keys(grouped)))
-        download_era5_day(metadata.name, metadata.dataset, grouped[day];
-                           region = metadata.region,
-                           dir = metadata.dir,
-                           skip_existing, cleanup)
+        append!(paths, download_era5_day(metadata.name, metadata.dataset, grouped[day];
+                                         region = metadata.region,
+                                         dir = metadata.dir,
+                                         skip_existing, cleanup))
     end
+
+    return paths
 end
 
 """
@@ -160,16 +163,15 @@ function _group_by_calendar_day(datetimes)
 end
 
 function download_era5_day(name, dataset, day_dates;
-                            region, dir, skip_existing, cleanup)
+                           region, dir, skip_existing, cleanup)
 
-    MDatum    = NumericalEarth.DataWrangling.Metadatum
-    meta_path = NumericalEarth.DataWrangling.metadata_path
+    meta_filename = NumericalEarth.DataWrangling.metadata_filename
 
-    all_pairs = [(dt, meta_path(MDatum(name; dataset, date=dt, region, dir)))
+    all_pairs = [(dt, joinpath(dir, meta_filename(dataset, name, dt, region)))
                  for dt in day_dates]
 
     pending = skip_existing ? filter(((_, p),) -> !isfile(p), all_pairs) : all_pairs
-    isempty(pending) && return nothing
+    isempty(pending) && return [p for (_, p) in all_pairs]
 
     sorted_dts = sort(unique([dt for (dt, _) in pending]))
     hours_str  = [lpad(string(Dates.hour(dt)), 2, '0') * ":00" for dt in sorted_dts]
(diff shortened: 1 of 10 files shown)


ewquon:
Oops, accidentally opened a PR from my fork again.

PR #209, opened 2026-05-04
Combine all ERA5 examples

ewquon:
The new example covers:
* Hourly global vs regional data retrieval, with `BoundingBox` and `Column` region types
* `ERA5HourlySingleLevel` and `ERA5HourlyPressureLevels` datasets


ewquon:
https://numericalearth.github.io/NumericalEarthDocumentation/previews/PR209/literated/ERA5_hourly_data/

simone-silvestri:
Looks good, maybe the CDS infos are a bit too verbose to render in the example, so if there is a way to suppress them it would be nice (maybe using Suppressor.jl?). Otherwise looks ok.

simone-silvestri (approved):

ewquon:
Thanks, @simone-silvestri, will look into Suppressor.

ewquon:
CI seems stuck

simone-silvestri:
CI does not fire if you haven't changed the src. Only the docs. However, the docs are waiting for the build on main to finish since we are using only one worker.

PR #210, opened 2026-05-04
Added wind stress keys to ECCO module and removed dead immersed grid export

GodotMisogi:
- The keys `zonal_wind_stress` and `meridional_wind_stress` have been added as retrieval keys from ECCO metadata.
- The undefined `ECCO_immersed_grid` function has been removed as an export.


simone-silvestri (approved):

simone-silvestri:
The test failures are not connected to this PR but to the fact that forks do not have access to the github secrets of this repository, used to provide credentials for downloading copernicus data. I will merge anyway.

PR #211, opened 2026-05-05
Fix ESM reset_clock! component clocks and add test

taimoorsohail:
I found a bug where `reset!(simulation)` for `Simulation{:EarthSystemModel}` only resets `simulation.model.clock` and not the clock of the individual components of the `ESM`. 

This means a workflow like:

```Julia
coupled_model = OceanSeaIceModel(ocean, sea_ice; atmosphere, radiation)
simulation = Simulation(coupled_model)

run!(simulation)
reset!(simulation)
run!(simulation)
```

Hits a segmentation fault. This is because `reset!` moves the ocean clock to 0, but the atmosphere clock doesn't change, then the `Initial Time Step` `interpolation` fails. 

I have attempted to fix this bug, and added a test that should fail on main which recreates this issue.


src/EarthSystemModels/earth_system_model.jl
```diff
@@ -55,10 +54,45 @@ timestepper(::ESM)                 = nothing
...
+    return nothing
+end
+
+function reset_clock!(model::ESM)
+    reset!(model.clock)
+
+    reset_component_clock!(model.ocean)
+    reset_component_clock!(model.sea_ice)
+    reset_component_clock!(model.atmosphere)
+    reset_component_clock!(model.land)
```
navidcy: Why not new methods, e.g., 

```julia
reset_clock!(ocean::TheAppropriateType) = reset!(ocean.clock)
reset_clock!(sea_ice::TheAppropriateType) = reset!(sea_ice.clock)
reset_clock!(atmosphere::TheAppropriateType) = reset!(atmosphere.clock)
reset_clock!(land::TheAppropriateType) = reset!(land.clock)
```

and then

```suggestion
    reset_clock!(model.ocean)
    reset_clock!(model.sea_ice)
    reset_clock!(model.atmosphere)
    reset_clock!(model.land)
```

Isn't this cleaner?

navidcy:
@taimoorsohail have a look at the changes I did. I think the functionality didn't change, meaning the changes I made didn't alter your original functionality and the tests still passes.

src/EarthSystemModels/earth_system_model.jl
```diff
@@ -57,13 +56,42 @@ timestepper(::ESM)                 = nothing
...
+
+reset_clock!(component::Simulation) = reset_clock!(component.model)
+
+function reset_clock!(component)
+    if hasproperty(component, :clock)
+        reset!(getproperty(component, :clock))
+    end
+
+    return nothing
+end
```
navidcy: Questions:

1. is there a component that doesn't have a clock? Why not

```suggestion
reset_clock!(component) = reset!(getproperty(component, :clock))
```

2. We sure we want to reset the clock for the Prescribed components (e.g. reanalysis atmosphere/land)?

src/EarthSystemModels/earth_system_model.jl
```diff
@@ -57,13 +56,42 @@ timestepper(::ESM)                 = nothing
...
+    reset_clock!(model.land)
+
+    # Keep prescribed-component cached state synchronized with the rewound clock.
+    if applicable(update_state!, model.atmosphere)
+        update_state!(model.atmosphere)
+    end
+
+    if applicable(update_state!, model.land)
+        update_state!(model.land)
+    end
```
navidcy: This seems cleaner:

```suggestion
    for component in (model.atmosphere, model.land)
        if applicable(update_state!, component)
            update_state!(component)
        end
    end
```

But also, it seems we are single-ing out atmosphere and land. Why is that. Does this come from the fact that we are used to running an ocean-sea ice model or is this more generic? What if we were running an AMIP simulation?

navidcy:
I think we need to have a cleaner way to find which model components are Prescribed and which are not.

Perhaps a `ModelComponent` type with `PrescribedModelComponent` and `DiagnosticModelComponent`?

```julia
::PrescribedModelComponent <: ModelComponent
::DiagnosticModelComponent <: ModelComponent
```

cc @simone-silvestri, @glwagner what do you think?

Then the above could be:

```Julia
    components = (model.radiation, model.atmosphere, model.land, model.ocean, model.sea_ice)
    prescribed_components = Tuple(c for c in components if c isa PrescribedModelComponent)

    for component in prescribed_components
        update_state!(component)
    end
```

simone-silvestri:
I think all the components define an `update_state!` and the `update_state!` should be idempotent so that we can safely call it any time. (we can define fallbacks like `update_state!(::Component) = nothing` for components that do not need updating). So the code above could simplify to
```julia
for component in components
   update_state!(component)
end
```

src/EarthSystemModels/earth_system_model.jl
```diff
@@ -57,13 +56,42 @@ timestepper(::ESM)                 = nothing
...
+        reset!(getproperty(component, :clock))
+    end
+
+    return nothing
+end
+
+function reset_clock!(model::ESM)
+    reset!(model.clock)
+
+    reset_clock!(model.ocean)
```
glwagner: this suggests we should define a function `components(esm)` so we can write

```julia
for component in components(esm)
    reset_clock!(component)
end
```

navidcy: Done via https://github.com/NumericalEarth/NumericalEarth.jl/pull/211/commits/77c76940e9a42671f816a9c5069d309961eb9685

src/EarthSystemModels/earth_system_model.jl
navidcy: talking with @taimoorsohail now seems the answer is yes

navidcy (approved):

src/EarthSystemModels/earth_system_model.jl
```diff
@@ -48,22 +48,34 @@ function Base.show(io::IO, cm::ESM)
...
-prognostic_fields(cm::ESM)         = nothing
+prognostic_fields(::ESM)           = nothing
 fields(::ESM)                      = NamedTuple()
-default_clock(TT)                   = Oceananigans.TimeSteppers.Clock{TT}(0, 0, 1)
+default_clock(TT)                  = Oceananigans.TimeSteppers.Clock{TT}(0, 0, 1)
+
+reset_clock!(::Nothing) = nothing
+
+reset_clock!(component::Simulation) = reset_clock!(component.model)
+reset_clock!(component) = reset!(getproperty(component, :clock))
```
simone-silvestri, line 64: probably we need a specialized `reset_clock!` for the non-Oceananigans components, like veros and SpeedyWeather.
I can give it a go.

ext/NumericalEarthVerosExt/veros_ocean_simulation.jl
```diff
@@ -300,3 +301,9 @@ function set!(ocean::VerosOceanSimulation, v, x; path = :variables)
         error("path must be either :variable or :settings.")
     end
 end
+
+function Simualtions.reset_clock!(ocean::VerosOceanSimulation)
```
simone-silvestri:
```suggestion
function Simulations.reset_clock!(ocean::VerosOceanSimulation)
```

simone-silvestri (approved):

taimoorsohail:
@navidcy happy for me to merge once relevant checks pass?

test/test_checkpointer.jl
```diff
@@ -2,6 +2,24 @@ include("runtests_setup.jl")
 
 using Glob
 using Oceananigans.OutputWriters: Checkpointer
+using Oceananigans.TimeSteppers: reset!
+
+function make_coupled_model(grid, arch)
```
navidcy: I suggest we read the arch from grid

navidcy: Regardless; good catch!

navidcy: https://github.com/NumericalEarth/NumericalEarth.jl/pull/211/commits/ce357d7dd046a0e51bd166d66b29698034c6182d

navidcy:
definitely merge if CI passes!

taimoorsohail:
I think the tests are failing for unrelated reasons

simone-silvestri:
the download tests are failing because of https://github.com/NumericalEarth/NumericalEarth.jl/issues/225#issuecomment-4429231929. We can still merge even if those tests fail, as soon as the ECCO server is back online I will add the missing files in the artifacts

navidcy:
hm...

Even after #225, here still we get:

```Julia
┌ Error: Your NVIDIA TITAN V GPU (compute capability 7.0) is not supported on CUDA 13+.
│ Please use a device with at least capability 7.5, or downgrade your NVIDIA driver to below v580.
└ @ CUDACore /storage5/github-action-runners/julia-depot/packages/CUDACore/gtlJx/lib/cudadrv/state.jl:225
```

https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/25728491231/job/75547348664#step:7:25

simone-silvestri:
weird, in https://github.com/CliMA/Oceananigans.jl/pull/5577 that was enough to fix the tartarus errors

glwagner:
> I think we need to have a cleaner way to find which model components are Prescribed and which are not.
> 
> Perhaps a `ModelComponent` type with `PrescribedModelComponent` and `DiagnosticModelComponent`?
> 
> ```julia
> ::PrescribedModelComponent <: ModelComponent
> ::DiagnosticModelComponent <: ModelComponent
> ```

Do you mean an abstract type, like `AbstractPrescribedComponent`? 

What is an example of a "diagnostic" component vs a prescribed component?

navidcy:
> > I think we need to have a cleaner way to find which model components are Prescribed and which are not.
> 
> > 
> 
> > Perhaps a `ModelComponent` type with `PrescribedModelComponent` and `DiagnosticModelComponent`?
> 
> > 
> 
> > ```julia
> 
> > ::PrescribedModelComponent <: ModelComponent
> 
> > ::DiagnosticModelComponent <: ModelComponent
> 
> > ```
> 
> 
> 
> Do you mean an abstract type, like `AbstractPrescribedComponent`? 
> 
> 
> 
> What is an example of a "diagnostic" component vs a prescribed component?
> 
> 

I meant to say prognostic, not diagnostic. 

A reanalysis atmosphere is prescribed while a SpeedyWeather atmosphere is prognostic.

glwagner:
We can't generically label components as prognostic with an abstract type because we don't own them. For example this would require `Simulation <: AbstractPrognosticComponent`. Or same for Speedy. But would you be ok if we just use the abstract type for prescribed components?

navidcy:
Good point. 

Yes, just a classification for the prescribed ones would be as useful.  We can still say things like 

```julia
component isa PrescribedModelComponent

!(component isa PrescribedModelComponent)

```

navidcy:
I'll make a new PR

PR #214, opened 2026-05-06
CI: aggregator job as single required check; drop trigger-level paths filter

glwagner:
## What

Restructure `CI` and `Documentation` so a single aggregator job (`test-summary` / `docs-summary`) is the only required status check, and the workflow *always* runs (so the aggregator always reports).

## Why

Both workflows had trigger-level `paths:` filters. When a PR touched no in-path files, the whole workflow was skipped. GitHub branch protection treats "never reported" as "still pending" — not as "passed" — so the merge button stayed disabled for non-admins even though the conversation view said "all checks passed". PR #209 hit this; only an admin force-merge got it through.

A required status check that **never reports** is not the same as one that **passed**. Skipped/not-triggered jobs do not satisfy the requirement.

## How

- **Drop trigger-level `paths:`** from both workflows so the workflow always starts.
- **Add a `changes` job** using `dorny/paths-filter@v3` that exposes a boolean output for the same path set the trigger filter used to gate.
- **Gate each previously-conditional job** with `if: github.event_name != 'pull_request' || needs.changes.outputs.<x> == 'true'`, so out-of-path PRs skip the heavy jobs (preserving prior behavior) but the workflow itself still runs.
- **Add an `if: always()` aggregator** (`test-summary` in `ci.yml`, `docs-summary` in `docs.yml`) that passes iff every needed job is `success` or `skipped`. Real failures still block the merge; legitimate skips don't.

The matrix in `cpu_tests` reports `skipped` only if the entire matrix was skipped — that's the desired behavior, no special handling needed.

## ⚠️ Branch protection follow-up (UI, not code)

This part has to be done in the GitHub UI — the workflow change alone is not enough. In **Settings → Rules → Rulesets** (or legacy Branch protection rules for `main`):

**Remove** these from "Require status checks to pass":
- `(CPU) - Julia 1.12.5`
- `(GPU)`
- `Data Downloading - …` (each matrix variant)
- `Reactant extension - …`
- `Build documentation`
- `deploy-docs`
- the old `CI` aggregator (if listed)
- any per-job names from the docs workflow

**Add** only:
- `test-summary`
- `docs-summary`

Leaving any of the old job names in the required-checks list will reproduce the bug regardless of these workflow changes — that's the load-bearing step.

## Out of scope

- Changing *what* tests run when. The conditional logic is preserved; only how completion is reported has changed.
- Migrating from branch protection rules to rulesets (orthogonal — same fix applies to both).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,15 +1,15 @@
 name: CI
 
+# Path-based gating is implemented *inside* this workflow (via the `changes`
+# job and per-job `if:` conditions) rather than via a trigger-level `paths:`
+# filter. With a trigger-level filter, the entire workflow — including the
+# `test-summary` aggregator — would not run on out-of-path PRs, leaving any
+# required status checks pending forever and blocking the merge button for
+# non-admins. By dropping `paths:` here, the workflow always starts and
+# `test-summary` always reports a status to branch protection.
 on:
   pull_request:
-    paths: &paths
-      - ".github/workflows/ci.yml"
-      - "ext/**"
-      - "src/**"
-      - "test/**"
-      - "Project.toml"
   push:
-    paths: *paths
     branches:
       - main
     tags: "*"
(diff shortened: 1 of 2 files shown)


simone-silvestri:
Cool, so in theory the CI does not run here if we don't change the src but there is also no "pending" final check?

simone-silvestri (approved):

glwagner:
@giordano does this make sense or is there a better way?

.github/workflows/ci.yml
```diff
@@ -312,17 +341,26 @@ jobs:
...
       - cds_downloading
       - reactant
     runs-on: ubuntu-latest
-    if: ${{ success() }}
     steps:
-      - run: echo "All CI jobs passed successfully."
+      - name: Check that all needed jobs passed or were skipped
+        run: |
+          echo '${{ toJSON(needs) }}'
+          jq -e 'to_entries | all(.value.result == "success" or .value.result == "skipped")' <<< '${{ toJSON(needs) }}'
```
giordano, line 366: What's this for?

giordano (commented):
> @giordano does this make sense or is there a better way?

Yeah, I've seen this approach before, I just find it annoying that GitHub marks as pending jobs which are skipped, and I wish they fixed that instead of forcing everybody to work around it.

PR #220, opened 2026-05-08
Add composable SlabLand and atmosphere–land coupling

kaiyuan-cheng:
## Summary

Adds a composable land-surface component to `EarthSystemModel` and the atmosphere–land coupling that lets Breeze drive it through similarity-theory turbulent fluxes. The default configuration — `SlabEnergy + BucketHydrology + ConstantSurfaceProperties` — is the classic Manabe-bucket slab in a few hundred lines of core numerics.

## Architecture — `SlabLand`, three closure axes

A `SlabLand` is a 2D land model on an Oceananigans `(Nx, Ny, Flat)` grid composed of three pluggable closures along independent axes:

- `energy    :: AbstractEnergyBalance`     — skin temperature
- `hydrology :: AbstractHydrology`         — water budget + moisture availability β
- `surface   :: AbstractSurfaceProperties` — albedo / emissivity / roughness

Each closure declares its prognostic state (`prognostic_variables`) and the flux/forcing fields it consumes (`flux_variables`); the container assembles the union as a `NamedTuple`, so a pure thermal slab carries `state = (T,)` only and the bucket adds the water-budget fields it actually needs. Atmosphere-facing accessors (`surface_temperature`, `surface_wetness`, `albedo`, `emissivity`, `momentum_roughness_length`, `scalar_roughness_length`) dispatch through the closures.

Per step the container calls `step!(energy, …)` then `step!(hydrology, …)`, refreshes diagnostics via `update_diagnostics!` in `hydrology → surface → energy` order, and fills state halos. Closures read each other's state through the shared `NamedTuple` but never call each other directly.

## Default closures

- `SlabEnergy` — single-temperature slab energy balance with areal heat capacity `ρ c H` (J m⁻² K⁻¹).
- `BucketHydrology` — Manabe (1969) bucket with field capacity `W_max` and critical wetness `W_crit / W_max`; produces the moisture-availability factor β consumed by the latent-heat BC.
- `ConstantSurfaceProperties` — spatially uniform albedo / emissivity / momentum and scalar roughness lengths.
- `DryLand` (β ≡ 0) and `SaturatedSurface` (β ≡ 1) — stub hydrologies for tests and idealized runs.

## Atmosphere–land coupling

`AtmosphereLandModel` mirrors `AtmosphereOceanModel`. A new `atmosphere_land_interface` runs the similarity-theory turbulent-flux pipeline:

- reads the slab's `T`, `moisture_availability` (β), and (optionally) `roughness_length` through the `ComponentExchanger`;
- computes sensible / latent / water-vapor / momentum fluxes plus `u★, θ★, q★`;
- writes back into the slab's flux accumulators — `net_energy_flux = −(𝒬ᵀ + 𝒬ᵛ)` (positive into the slab), and the positive/negative parts of the vapor flux into `precipitation` / `evaporation`.

Atmosphere flux drivers now dispatch on the interface type, so adding new surface components doesn't require touching the ocean/sea-ice paths. A 3D Breeze + `SlabLand` example over a homogeneous slab is included (`examples/breeze_over_slab_land.jl`).

## Known limitations (follow-ups)

- Land radiation (albedo / emissivity → SW / LW) is not yet plumbed through the `Radiation` component.
- When both ocean and land interfaces are present, the Breeze bottom-BC assembler overwrites rather than tile-fraction blending.
- Direct-Breeze-BC integration.
- Heterogeneous land-use / vegetation-class registry to return alongside richer closures.


examples/breeze_over_ruc_slab_land.jl
```diff
@@ -0,0 +1,307 @@
...
+# turbulent exchange between sensible and latent heat differs across strips.
+
+land_grid = RectilinearGrid(grid.architecture,
+                            size = (grid.Nx, grid.Ny),
+                            halo = (grid.Hx, grid.Hy),
+                            x = (-10kilometers, 10kilometers),
+                            y = (-10kilometers, 10kilometers),
+                            topology = (Periodic, Periodic, Flat))
+
+slab_land = RucSlabLand(land_grid;
```
glwagner:
```suggestion
slab_land = SlabLand(land_grid;
```

examples/breeze_over_ruc_slab_land.jl
```diff
@@ -0,0 +1,307 @@
...
+                        parameters = RucSlabLandParameters(eltype(land_grid);
```
glwagner: Let's avoid properties called `parameters` and struct names that end in `*Parameters`. Instead, choose names that explicity describe the property, like `hydrology`, `energy`, etc.

src/EarthSystemModels/InterfaceComputations/atmosphere_ocean_fluxes.jl
```diff
@@ -2,6 +2,8 @@ using Oceananigans.Operators: intrinsic_vector
 using Oceananigans.Grids: inactive_node
 
 function compute_atmosphere_ocean_fluxes!(coupled_model)
+    isnothing(coupled_model.interfaces.atmosphere_ocean_interface) && return nothing
```
glwagner: It could be better to use dispatch for this

src/Lands/energy_balance/ruc_energy.jl
```diff
@@ -0,0 +1,99 @@
...
+"""
+    RucEnergy(parameters::RucSlabLandParameters)
+    RucEnergy(FT = Float64; depth, density, heat_capacity, canopy_heat_capacity)
+
+The RUC slab energy balance. `(depth, density, heat_capacity)` define
+the ground areal heat capacity `(ρ c H)_g = depth · density ·
+heat_capacity` in `J m⁻² K⁻¹`. `canopy_heat_capacity` is the canopy
+areal `(ρ c H)_c` directly.
+"""
+struct RucEnergy{FT} <: AbstractEnergyBalance
```
glwagner: this looks like a slab model. Is it really different from `SlabEnergy`?

src/Lands/energy_balance/ruc_energy.jl
```diff
@@ -0,0 +1,99 @@
...
+    depth                :: FT   # H_g [m]
+    density              :: FT   # ρ_g [kg m⁻³]
+    heat_capacity        :: FT   # c_g [J kg⁻¹ K⁻¹]
+    canopy_heat_capacity :: FT   # (ρ c H)_c [J m⁻² K⁻¹]
```
glwagner: is the canopy part of the interface or part of the land model? I think there is a tricky ambiguity here. We may want to avoid this now so that we don't have to undo a lot of things in the future

src/Lands/surface_properties/constant_surface_properties.jl
```diff
@@ -0,0 +1,43 @@
...
+                              z0_m = 0.1,
+                              z0_h = 0.01)
+
+Spatially uniform surface optical and aerodynamic properties.
+"""
+struct ConstantSurfaceProperties{FT} <: AbstractSurfaceProperties
+    albedo     :: FT
+    emissivity :: FT
+    z0_m       :: FT
+    z0_h       :: FT
```
glwagner: We don't use this notation for roughness lenghs. It could be best to use `momentum_roughness_length` here. Also I believe the scalar doesn't only apply to heat (also applies to moisture) so `scalar_roughness_length` is more appropriate.

src/Lands/hydrology/manabe_bucket.jl
```diff
@@ -0,0 +1,59 @@
...
+    ManabeBucket(FT = Float64; field_capacity = 150, critical_wetness = 0.75)
+
+Manabe (1969) bucket. `field_capacity` is `W_max` in `kg m⁻²` (≈ 15 cm
+of liquid water). `critical_wetness` is `W_crit / W_max` — the
+fractional wetness above which `β` saturates to one.
+"""
+struct ManabeBucket{FT, B} <: AbstractHydrology
+    field_capacity   :: FT
+    critical_wetness :: FT
+    β                :: B   # cached wetness Field; allocated by initial_state
```
glwagner: Let's remove "Manabe" from the name of this, and come up with a name for "beta"

src/Lands/surface_properties/ruc_surface_properties.jl
```diff
@@ -0,0 +1,191 @@
...
+        α_blend = max(keep * α_sn,
+                      min(α_free + (α_sn - α_free) * f_sn, α_sn))
+        if α_blend < FT(0.4) || keep == one(FT)
+            albedo_field[i, j, 1] = α_blend
+        else
+            T_C = T[i, j, 1] - FT(273.15)
+            albedo_field[i, j, 1] = min(α_blend,
+                               max(α_blend - FT(0.1) * (T_C + FT(10)) /
+                                             FT(10) * α_blend,
+                                   α_blend - FT(0.05)))
```
glwagner: Introduce an interface for computing albedo, rather than hard coding a parameterization in this way that is monolithic with the rest of the kernel. In addition, the parameters of the parameterization should be changeable rather than hard coded here. Also, I think we should keep all temperature computations in Kelvin for simplicity. There is a function `convert_to_celsius` if we need to convert (don't hard code constants)

src/Lands/surface_properties/ruc_surface_properties.jl
```diff
@@ -0,0 +1,191 @@
...
+        end
+        roughness_length_field[i, j, 1] = z₀_new
+
+        # Shortwave-albedo blend (Robinson and Kukla 1985), with the
+        # warm-T attenuation branch below 0.4 disabled when fresh snow has
+        # latched `keep_snow_albedo = 1`.
+        α_blend = max(keep * α_sn,
+                      min(α_free + (α_sn - α_free) * f_sn, α_sn))
+        if α_blend < FT(0.4) || keep == one(FT)
+            albedo_field[i, j, 1] = α_blend
```
glwagner: don't use short-circuiting logic. Use types, and don't hard code 0.4

src/Lands/surface_properties/ruc_surface_properties.jl
```diff
@@ -0,0 +1,191 @@
...
+        z₀_veg_ij = roughness_length_vegetation_field[i, j, 1]
+        α_free  = (one(FT) - vf) * albedo_bare   + vf * α_veg_ij
+        ε_free  = (one(FT) - vf) * emissivity_bare + vf * ε_veg_ij
+        z₀_free = (one(FT) - vf) * roughness_length_bare    + vf * z₀_veg_ij
+
+        # Roughness: default to the snow-free composite. RUC's
+        # snow-blending branch only applies when the surface is short
+        # (z₀_free ≤ 0.2 m, i.e. grass / shrub / cropland) and there is
+        # no fresh snow this step; tall canopies (forests with
+        # z₀_free > 0.2 m) keep the canopy roughness.
```
glwagner: Introduce an abstraction for "blended" roughness lengths, and put the free parameters in it so they can be changed (don't hard code parameters)

src/Lands/surface_properties/ruc_surface_properties.jl
```diff
@@ -0,0 +1,191 @@
...
+        z₀_new = z₀_free
```
glwagner: Roughness length is denoted as $\ell$. The symbol `z` denotes "height above the surface", so it cannot be used for a length scale.

src/Lands/ruc_slab_land.jl
```diff
@@ -0,0 +1,254 @@
...
+                           T = nothing,
+                           Tc = nothing,
+                           θ = nothing,
+                           θ_ice = nothing,
+                           snwe = nothing,
+                           snhei = nothing,
+                           rhosn = nothing,
+                           swl = nothing,
+                           vegfrac = nothing,
+                           lai = nothing)
```
glwagner: We need proper math symbols for many of these

src/Lands/ruc_slab_land.jl
```diff
@@ -0,0 +1,254 @@
...
+    surface = land.surface
+
+    !isnothing(T)              && haskey(state, :T)   && set!(state.T,   T)
+    !isnothing(Tc)             && haskey(state, :Tc)  && set!(state.Tc,  Tc)
+    isnothing(Tc) && !isnothing(T) && haskey(state, :Tc) && set!(state.Tc, T)
+    !isnothing(θ)              && haskey(state, :θ)     && set!(state.θ,     θ)
+    !isnothing(θ_ice)          && haskey(state, :θ_ice) && set!(state.θ_ice, θ_ice)
+    !isnothing(snwe)           && haskey(state, :snwe)  && set!(state.snwe,  snwe)
+    !isnothing(snhei)          && haskey(state, :snhei) && set!(state.snhei, snhei)
+    !isnothing(rhosn)          && haskey(state, :rhosn) && set!(state.rhosn, rhosn)
```
glwagner: Claude likes `haskey` checks, but I think this is bad practice in general. I suggest use explicit dispatch for this.

src/Lands/ruc_slab_land_parameters.jl
```diff
@@ -0,0 +1,204 @@
...
+
+# Phase change
+- `latent_heat_fusion = 3.337e5`  : J kg⁻¹
+
+# Stomatal resistance (Jarvis-Stewart)
+- `stomatal_resistance_min = 100.0`, `stomatal_resistance_max = 5000.0`,
+  `solar_radiation_limit = 100.0`, `vapor_pressure_deficit_limit = 4.0e-3`,
+  `temperature_optimum = 298.0`
+"""
+struct RucSlabLandParameters{FT}
```
glwagner: There are too many different concepts lumped into a single struct here: canopy, soil, snow, vegetation. Can we use a more hierarchical design, and have dedicated structs for each concept? That will allow different components to be swapped in and out.

Note, radiation properties do not go into components. Radiation parameterizations should be added to the radiation component.

src/Lands/usgs_land_classification.jl
```diff
@@ -0,0 +1,180 @@
...
+##### | 15 | Mixed Forest                  | 0.80   | 5.50 | 0.80   |0.13|0.94 | 125 |
+##### | 16 | Water                         | 0.00   | 0.01 | 0.0001 |0.08|0.98 | 100 |
+##### | 17 | Herbaceous Wetland            | 0.60   | 5.65 | 0.20   |0.14|0.95 | 40  |
+##### | 18 | Wooded Wetland                | 0.60   | 5.80 | 0.40   |0.14|0.95 | 100 |
+##### | 19 | Barren or Sparsely Vegetated  | 0.01   | 0.75 | 0.05   |0.25|0.85 | 999 |
+##### | 20 | Herbaceous Tundra             | 0.60   | 3.35 | 0.10   |0.15|0.92 | 150 |
+##### | 21 | Wooded Tundra                 | 0.60   | 3.35 | 0.15   |0.15|0.93 | 150 |
+##### | 22 | Mixed Tundra                  | 0.60   | 3.35 | 0.10   |0.15|0.92 | 150 |
+##### | 23 | Bare Ground Tundra            | 0.30   | 3.35 | 0.065  |0.25|0.85 | 200 |
+##### | 24 | Snow or Ice                   | 0.00   | 0.01 | 0.011  |0.55|0.98 | 999 |
```
glwagner: Since we are using Julia, I don't think we need to use a look up table with numerical classifications. I think it would be better to use ordinary structs with names.

src/Lands/usgs_land_classification.jl
```diff
@@ -0,0 +1,180 @@
...
+- `albedo`         : snow-free shortwave albedo.
+- `emissivity`     : snow-free longwave emissivity.
+- `r_smin`         : minimum stomatal resistance [s m⁻¹].
+- `root_depth`     : nominal root-zone depth [m] (informational).
+
+Use `usgs_land_classifications(FT)` for the default 24-class registry.
+"""
+struct LandClassification{FT}
+    id :: Int
+    name :: Symbol
```
glwagner: note, symbols are not isbits and do not compile on GPU

src/Lands/usgs_land_classification.jl
```diff
@@ -0,0 +1,180 @@
...
+    vegfrac :: FT
+    lai :: FT
+    z0 :: FT
```
glwagner:
```suggestion
    roughness_length :: FT
```

src/Lands/usgs_land_classification.jl
```diff
@@ -0,0 +1,180 @@
...
+    albedo :: FT
+    emissivity :: FT
+    r_smin :: FT
```
glwagner: Use an english name?

src/Lands/usgs_land_classification.jl
```diff
@@ -0,0 +1,180 @@
...
+- `lai`            : peak leaf area index [m² m⁻²].
+- `z0`             : roughness length [m].
+- `albedo`         : snow-free shortwave albedo.
+- `emissivity`     : snow-free longwave emissivity.
+- `r_smin`         : minimum stomatal resistance [s m⁻¹].
+- `root_depth`     : nominal root-zone depth [m] (informational).
+
+Use `usgs_land_classifications(FT)` for the default 24-class registry.
+"""
+struct LandClassification{FT}
```
glwagner:
```suggestion
struct VegatationProperties{FT}
```

and then remove `name`

docs/src/appendix/notation.md
```diff
@@ -109,12 +109,26 @@ denote the _component_ or _location_.
...
 
+## Land state variables
+
+| Math | Code | Property | Description |
+|:----:|:----:|:---------|:------------|
+| ``W`` | `W` | water storage | Layer-integrated liquid water storage (kg m⁻²) |
+| ``β`` | `β` | moisture availability | Moisture availability factor (–) |
+| ``Tᶜ`` | `Tᶜ` | deep climatological temperature | Prescribed deep/climatological target temperature (K) |
+| ``Tᵈ`` | `Tᵈ` | deep temperature state | Prognostic deep-layer soil temperature (K) |
+| ``zʳ`` | `zʳ` | effective root depth | Root-zone scaling of moisture capacity (m) |
```
glwagner: let's come up with a good symbol for this ("z" as a "Depth" would have to be negative). Maybe just "d"

docs/src/appendix/notation.md
```diff
@@ -109,12 +109,26 @@ denote the _component_ or _location_.
...
+| ``\Lambda`` | `Λ` | leaf area index | Leaf area index (–) |
+| ``τˢ`` | `τˢ` | surface-to-deep coupling timescale | Surface/deep temperature relaxation timescale (s) |
+| ``τᵈ`` | `τᵈ` | deep-to-climate coupling timescale | Deep temperature climatological relaxation timescale (s) |
```
glwagner: we use \tau for surface stresses, so maybe a mathcal T here

docs/src/appendix/notation.md
kaiyuan-cheng: "d" is good.

docs/src/appendix/notation.md
kaiyuan-cheng: This one is tricky. τ is also used as timescale in Breeze, but I think it is ok to use \mathcal{T} in Breeze.

examples/era5_forced_slab_land.jl
```diff
@@ -0,0 +1,332 @@
...
+# bundled into a single [`MetadataSet`](@ref) and loaded onto the 1 km
+# land grid in one call — the dataset backend bilinearly downscales
+# from ERA5 native to 1 km at load time.
+#
+# ## Elevation downscaling
+#
+# `SlabLand` itself has no terrain knowledge, and ERA5's T₂ₘ is at its
+# own ~28 km grid-cell mean elevation (~2 km in this domain). To make
+# the 1 km grid show elevation-driven temperature contrasts we apply a
+# dry-environmental lapse-rate correction:
```
kaiyuan-cheng:
```suggestion
# moist-environmental lapse-rate correction:
```

docs/src/appendix/notation.md
glwagner: true, we may want to change the notation in Breeze too

examples/era5_forced_slab_land.jl
```diff
@@ -0,0 +1,332 @@
...
+#
+# `SlabLand` itself has no terrain knowledge, and ERA5's T₂ₘ is at its
+# own ~28 km grid-cell mean elevation (~2 km in this domain). To make
+# the 1 km grid show elevation-driven temperature contrasts we apply a
+# moist-environmental lapse-rate correction:
+#
+#     T_local(λ, φ) = T_ERA5(λ, φ) − Γ · (z_ETOPO(λ, φ) − z_ERA5_eff(λ, φ))
+#
+# with Γ = 6.5 K km⁻¹. `z_ERA5_eff` is ETOPO box-averaged onto the ERA5
+# native grid (≈ what ERA5 thinks the surface elevation is in each
```
ewquon: Worth noting that this procedure is inconsistent with the construction of mean orography in IFS.

From section 11.3.2 in https://www.ecmwf.int/en/elibrary/79697-ifs-documentation-cy41r2-part-iv-physical-processes:
> Orography, or geopotential height, is derived from the 30′′data in a few steps:
>
> 1. The smoothing operator with a Pre-Filter-Scale of 5 km is applied to the 30′′orography.
> 2. The smoothed 30′′data is interpolated to a 5 km grid (reduced Gaussian grid with NGL= 4000). For target resolutions with 4000 grid points or more between the south and north poles (NGL≥4000), the pre-smoothing and interpolation scale of 5 km is reduced to 2 km.
> 3. The 5 km (or 2 km) resolution field is smoothed with the Filter-Scale listed in Table 11.2 dependent on the selected target resolution, e.g. at T1279, the linear grid has NGL= 1280 and a Filter-Scale of 16000 m.
> 4. The resulting field is interpolated to target resolution, e.g. NGL= 1280 for a T1279 linear grid.
> 5. The grid-point orography is transformed to spectral space at target resolution, e.g. T1279.
> 6. The orographic spectrum is tapered by multiplication with 1/[1 + 4{l(l+ 1)}8{lmax(lmax + 1)}−8], where l is the wave number index and lmax is the maximum wave number of the target resolution (e.g. 1279). This is equivalent to smoothing of the grid point field with a 5∇16 operator. The exponent determines the sharpness of the tapering and the factor 5 determines the filtering strength.
> 7. The spectral orography is transformed back to grid point space.

glwagner: how would you recommend running a land simulation forced by ERA5, but where the land has a higher resolution than ERA5 thus resolving more topography?

ewquon: You can smooth the higher-res topography to match the ERA5 mean topography within a fringe region. This blending region would coincide with the boundary relaxation zone.

glwagner: Is that relevant for this case which is a land simulation only? I think what you're suggesting might be something we need to do if we have a prognostic atmosphere embedded in ERA5 solution (but maybe I am not completely understanding the suggestion)

ewquon: My original comment was regarding the simple average of the underlying high-res terrain to the ERA5 grid. Just pointing out that in practice ERA5 also starts with higher-resolution elevation data but does something quite a bit more sophisticated. So many just add a comment that we're demonstrating a simplified approach.

navidcy:
@kaiyuan-cheng please add your details in the CITATION.cff file in this PR so it's updated when merged!

kaiyuan-cheng:
> @kaiyuan-cheng please add your details in the CITATION.cff file in this PR so it's updated when merged!

Done. Thanks!

examples/era5_forced_slab_land.jl
glwagner: Ah ok I see what you're saying. I wish we could access the raw ERA5 data instead of the interpolated data. I guess the "effective topographic height" of the interpolated data is not known? Or maybe the right thing is to somehow get the true ERA5 elevation from somewhere. I will look into this and adjust the comment either way.

docs/src/appendix/notation.md
```diff
@@ -109,11 +110,30 @@ denote the _component_ or _location_.
...
+| ``M^{\mathrm{la}}`` | `water_storage` | land water | Prognostic land water mass per area (kg m⁻²) |
+| ``M^{\mathrm{la}\!+}`` | `maximum_water_storage` | maximum land water | Bucket capacity; soil-science "field capacity" (kg m⁻²) |
+| ``𝒮`` | `moisture_availability` | surface saturation | Continuous land surface saturation ``\mathrm{clamp}(Mˡᵃ/Mˡᵃ⁺, 0, 1)``; the interface humidity models derive their availability ``β`` from it (–) |
+| ``𝒮ᶜ`` | `critical_saturation` | critical saturation | Saturation above which the surface evaporates at full efficiency, for `CriticalWetness` (–) |
+| ``T^{\mathrm{deep}}`` | `deep_temperature` | deep climatological temperature | Prescribed deep/climatological target temperature for force-restore (K) |
+| ``τ^{\mathrm{deep}}`` | `deep_time_scale` | deep-restore time scale | Time scale of surface relaxation toward ``T^{\mathrm{deep}}`` (s) |
+| ``zʳ`` | `root_depth` | effective root depth | Root-zone scaling of storage capacity (m) |
+| ``d`` | `surface_thickness` | saturation depth | Depth of the dry surface layer through which soil vapor diffuses, for `SkinHumidity` (m) |
```
ewquon: Any reason we didn't make the variable name and property name more consistent here?

docs/src/appendix/notation.md
```diff
@@ -155,6 +175,8 @@ denote the _component_ or _location_.
 | ``\psi`` | `ψ` | stability function | Integrated stability correction (–) |
 | ``\zeta`` | `ζ` | stability parameter | ``z / L_\star`` (–) |
 | ``\ell`` | `ℓ` | roughness length | Aerodynamic roughness length (m) |
+| ``\ell^\mathrm{m}`` | `ℓᵐ` | momentum roughness length | Aerodynamic momentum roughness length (m) |
```
ewquon, line 174: 🪦 z₀

glwagner: good riddance! 😂

docs/src/appendix/notation.md
glwagner: bah, z^r for root depth is obv wrong (even worse than roughness bc z is negative for land!)

glwagner: also I don't think we have a root model in this PR so we should just remove it

ext/NumericalEarthCDSAPIExt.jl
```diff
@@ -661,6 +661,27 @@ const COL  = NumericalEarth.DataWrangling.Column
...
+# Columns and unbounded regions: the area is a pure function of the region.
+era5_request_area(region, dataset, name, date) = build_era5_area(region)
+
+# Bounding box: the native grid is built by center-bracketing `restrict`, which
+# can reach one cell past a boundary-aligned edge. Fetch two native cells of
+# margin (in the bbox's own longitude convention) so the downloaded file always
+# covers the grid the data is interpolated onto — otherwise downscaling leaves
+# NaNs at the domain edges. Over-fetching is harmless: `restrict` selects the
+# exact cells from the larger file.
+function era5_request_area(bbox::BBOX, dataset, name, date)
```
ewquon: I don't see why `date` is needed as an argument

glwagner: fixing this

glwagner: Right — `date` was carried in the signature but never consulted inside the body (the area depends only on the bbox + the dataset/variable's native resolution). Dropped in 9ae61e71 from both the `BBOX` method and the generic fallback, and from the `build_era5_request` call site.

src/EarthSystemModels/InterfaceComputations/interface_states.jl
```diff
@@ -123,13 +275,19 @@ struct WindVelocity end
...
 @inline function velocity_difference(::RelativeVelocity, 𝒰₁, 𝒰₀)
-    Δu = 𝒰₁.u - 𝒰₀.u
-    Δv = 𝒰₁.v - 𝒰₀.v
+    Δu = x_velocity(𝒰₁) - x_velocity(𝒰₀)
+    Δv = y_velocity(𝒰₁) - y_velocity(𝒰₀)
     return Δu, Δv
 end
 
-@inline velocity_difference(::WindVelocity, 𝒰₁, 𝒰₀) = 𝒰₁.u, 𝒰₁.v
+@inline velocity_difference(::WindVelocity, 𝒰₁, 𝒰₀) = x_velocity(𝒰₁), y_velocity(𝒰₁)
```
ewquon, line 290: Can we call this `flux_velocity` or `exchange_velocity`? `velocity_difference` is a misnomer when called like this.

glwagner: I think the purpose of using the term "velocity difference" is that this is fundamentally correct in the context of MO theory (and Galilean invariance). When people pretend the surface isn't moving, it's an approximation, not a different MO theory.

glwagner: So quite literally this property determines "how do we compute the velocity difference between the atmosphere and the surface"?

It is not otherwise a parameterization for a special velocity field with a bespoke name. If that makes sense.

glwagner: Or perhaps put another way, the output of this function very much IS the velocity difference. It's just that some people choose to pretend that surfaces are not moving, when they actually are!

src/EarthSystemModels/InterfaceComputations/interface_states.jl
```diff
@@ -371,44 +529,211 @@ end
...
+temperature), and `specific_humidity` (`qˢ`) — and differ only in the surface
+property each interface needs: `salinity` for air–sea, the land `hydrology` /
+`energy` state for air–land.
+"""
+abstract type AbstractInterfaceState{FT} end
+
+Base.eltype(::AbstractInterfaceState{FT}) where FT = FT
+
+# Interface velocity components (see `velocity_difference`).
+@inline x_velocity(Ψ::AbstractInterfaceState) = Ψ.velocities.u
```
ewquon, line 650: If we're using `Ψ` throughout to represent interface states, it might be worth a note in the notations doc. Not sure if there's a precedent for this — personally I'd prefer `S` or even `𝒮` but I know those are "taken" by salinity and saturation.

examples/breeze_over_slab_land.jl
```diff
@@ -0,0 +1,363 @@
...
+# evaporates strongly during the day so that incoming radiation is partitioned
+# into latent heat. The dry edges (`𝒮 = 0`) cannot evaporate, so all the net
+# radiation goes into sensible heat — producing strong surface heating
+# and a vigorous dry convective boundary layer. At the wet/dry boundary
+# the contrast drives a low-level "sea breeze"-like circulation.
+#
+# Coupling lives entirely in the `EarthSystemModel`:
+#   * `AtmosphereLandModel(atmos, slab_land; radiation = rtm)` wires
+#     turbulent surface fluxes (sensible, latent, momentum) through
+#     Monin–Obukhov similarity theory and hands the RRTMGP
```
ewquon: We need to implement different stability functions over land. Right now, I think it's defaulting to Edson

ewquon: > _"A default_atmosphere_land_fluxes that silently inherits atmosphere_ocean_stability_functions is a footgun."_
> – _Claude_

docs/src/appendix/notation.md
```diff
@@ -15,16 +15,16 @@ radiative intensity, `J` for mass flux, and `τ` for kinematic momentum flux.
 **Superscripts** refine the meaning in several ways:
 
 - _Phase or species_: `ᵛ` (vapor), `ˡ` (liquid), `ⁱ` (ice), `ᶜ` (condensate)
-- _Component_: `ᵃᵗ` (atmosphere), `ᵒᶜ` (ocean), `ˢⁱ` (sea ice), `ˡᵈ` (land)
+- _Component_: `ᵃᵗ` (atmosphere), `ᵒᶜ` (ocean), `ˢⁱ` (sea ice), `ˡᵃ` (land)
 - _Interface pair_: `ᵃᵒ` (atm–ocean), `ᵃⁱ` (atm–ice), `ⁱᵒ` (ice–ocean)
```
kaiyuan-cheng: no `al` needed?

examples/breeze_over_slab_land.jl
```diff
@@ -0,0 +1,363 @@
...
+    T_xz .- mean(T_xz, dims = 1)
+end
+qˡn   = @lift view(interior(qˡ_ts[$n]), :, 1, :)
+Tˡᵃ_n = @lift vec(interior(Tˡᵃ_ts[$n], :, 1, 1))
+M_n   = @lift vec(interior(M_ts[$n],   :, 1, 1))
+𝒮_n   = @lift vec(interior(𝒮_ts[$n],   :, 1, 1))
+
+heatmap!(ax_w,   x_atmos, z_face,   wn;    colormap = :balance, colorrange = (-wlim, wlim))
+heatmap!(ax_Tᵃᵗ, x_atmos, z_center, Tᵃᵗ_n; colormap = :balance, colorrange = (-2, 2))
+heatmap!(ax_qˡ,  x_atmos, z_center, qˡn;   colormap = :dense,   colorrange = (0, qˡlim))
```
ewquon: I can't see any cloud liquid water in the preview -- perhaps the upper bound `qˡlim` should be reduced?

Also, some colorbars for the heatmaps would be nice.

examples/breeze_over_slab_land.jl
kaiyuan-cheng: Nice catch!

src/Lands/hydrology/dry_land.jl
```diff
@@ -0,0 +1,14 @@
...
+#####
+##### Useful for desert idealisations and unit-test isolation of the
+##### energy path.
+#####
+
+struct DryLand <: AbstractHydrology end
+
+flux_variables(::DryLand) = ()
+
+saturation(::DryLand, land) = 0.0
```
glwagner:
```suggestion
saturation(::DryLand, land) = 0
```

glwagner (approved):

Project.toml
```diff
@@ -5,8 +5,12 @@ version = "0.5.3"
 authors = ["NumericalEarth contributors"]
 
 [deps]
+Accessors = "7d9f7c33-5ae7-4f3b-8dc6-eff91059b697"
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
+Breeze = "660aa2fb-d4c8-4359-a52c-9c057bc511da"
 CFTime = "179af706-886a-5703-950a-314cd64e0468"
+CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
+CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
```
navidcy, line 13: were CUDA and CairoMakie added here intentionally? because I don't see where they are used in the source code

if they were here for debugging purposed let's delete them?
if they were intentional and we are keeping them they need a compat entry; registering v0.5.4 fails at the moment because of the lack of a compat entry for these two; see https://github.com/JuliaRegistries/General/pull/156755#issuecomment-4587597597

PR #223, opened 2026-05-11
Pressure-based formula for surface specific humidity

simone-silvestri:
This PR  replaces the density-extrapolation form of `surface_specific_humidity` with the direct pressure-based formula from COARE 3.6 / Edson (2013): $q_s = \gamma · e_s / (p − (1 − \gamma) e_s),   \gamma = R^d / R_v$
where $e_s = \chi_{H_2O}  p^\star (T_s)` is the (Raoult-corrected) saturation vapor pressure at the surface temperature.

The code on main first extrapolated air density to the surface temperature via an adiabatic ideal-gas transformation, then computed $q_s$ from $(\rho_s, p^\star)$ via `q_vap_from_p_vap`. The new form evaluates $q_s$ at the actual atmospheric pressure directly with no density extrapolation steps. The underlying assumption behind this formulation is that all moisture is vapor at saturation.


src/EarthSystemModels/InterfaceComputations/interface_states.jl
```diff
@@ -51,31 +51,22 @@ ImpureSaturationSpecificHumidity(phase) = ImpureSaturationSpecificHumidity(phase
...
+    p  = convert(CT, pᵃᵗ)
+    p★ = AtmosphericThermodynamics.saturation_vapor_pressure(ℂᵃᵗ, T, formulation.phase)
 
-    # Compute saturation specific humidity according to Raoult's law
+    # Raoult's law on the saturation vapor pressure.
     χ_H₂O = compute_water_mole_fraction(formulation.water_mole_fraction, Sₛ)
-    qₛ = χ_H₂O * q★
+    eₛ    = χ_H₂O * p★
+    ε     = 1 / AtmosphericThermodynamics.Parameters.Rv_over_Rd(ℂᵃᵗ)
+    qₛ    = ε * eₛ / (p - (1 - ε) * eₛ)
```
glwagner: The notation doesn't follow what's written in `notation.md`

simone-silvestri:
do you mean `e` and `p\bigstar`? I think we do not have a notation for saturation pressure, I can change `e_s` to `p_s` but I think it can be misunderstood for "surface pressure". Probably better to update the notation.md

glwagner:
> do you mean `e` and `p\bigstar`? I think we do not have a notation for saturation pressure, I can change `e_s` to `p_s` but I think it can be misunderstood for "surface pressure". Probably better to update the notation.md

If there is notation missing from NumericalEarth, can we follow Breeze's (and add to NumericalEarth notation docs)? Breeze notation is defined here:

https://numericalearth.github.io/BreezeDocumentation/dev/appendix/notation/

Saturation vapor pressure is denoted $p^{v+}$, and saturation specific humidity is $q^{v+}$. Optionally, we can also denote the surface over which the saturation specific humidity is defined (liquid, ice, or a mixed surface).

More generally I think it is bad practice to introduce new symbols for a concept, eg we use $p$ for pressure; therefore saturation vapor pressure should also use $p$ (not $e$ -- this is historical, but confusing).

In terms of $\star$, we use that for the interface state (eg fluxes), so cannot use for saturation vapor pressure / humidity (though this is how the code was originally written, I think we should try to be more systematic in the future).

simone-silvestri:
Resolved and added the notation

docs/src/appendix/notation.md
```diff
@@ -201,6 +203,7 @@ with a superscript band.
 | ``c^{pm}`` | `cᵖᵐ` | moist air heat capacity | Moist isobaric specific heat (J kg⁻¹ K⁻¹) |
 | ``c^{pd}`` | `cᵖᵈ` | dry air heat capacity | Dry-air isobaric specific heat (J kg⁻¹ K⁻¹) |
 | ``\rho^{\mathrm{at}}`` | `ρᵃᵗ` | air density | Atmospheric air density (kg m⁻³) |
+| ``\varepsilon`` | ``ε`` | specific gas constant ratio (``Rd / Rv`` ) | Ratio of dry air and vapor gas constants (-) |
```
glwagner: Can you add a label $\epsilon^{dv}$ here? $\epsilon$ is too broad (we use it for many things). Breeze also uses ordinary epsilon, but we can chnage to varepsilon if you prefer it (I think varepsilon is ugly 😂 )

glwagner:
> Resolved and added the notation

I don't think the notation for saturation vapor pressure is correct yet (I just see `ps`)

simone-silvestri:
no more `ps`

docs/src/appendix/notation.md
simone-silvestri: added the `dv`. I kind of wanted to keep it separate to emissivity. For me epsilon is automatically emissivity, so at least having a separate symbol helps. Maybe we can change it entirely?

glwagner (approved):
note the description has outdated notation still

PR #224, opened 2026-05-12
Adjust sea ice model constants to align with ClimaSeaIce.jl#main

taimoorsohail:
Currently, NumericalEarth.jl#main and ClimaSeaIce.jl#main cannot be run together due to a mismatch in constant naming and function definitions in `src/EarthSystemModels/InterfaceComputations`. 

I have attempted to change the relevant files so that `OceanSeaIceModel()` again works with the #main branches of both projects. This was the minimal fix I could find to ensure the code ran -- not necessarily an exhaustive fix for existing bugs in the #main branches of either repo.


simone-silvestri:
This should be fixed in PR #164, can you review that one? Should be rather easy a part the change to the temperature interface computation which is reformulated a bit, the rest should be trivial (split freshwater fluxes in rain and snow and accumulate snow in the sea ice fluxes).

taimoorsohail:
Thanks @simone-silvestri -- I had a look and it looks good. However, I think there are some specific changes here that are not in #164 :

1) Swap OceanSeaIceModel() input order so it is (ocean, sea_ice; kwargs...) as before
2) Adjust the ThreeEquationState default in `ComponentInterfaces` to be `sea_ice_ocean_heat_flux = ThreeEquationHeatFlux(eltype(exchange_grid)),`
3) Add a guard so that ocean.model is not accidentally passed to sea_ice specific components:

```julia
sea_ice_model(sea_ice::Simulation) = sea_ice_model(sea_ice.model)
sea_ice_model(model::SeaIceModel) = model

sea_ice_model(model) =
    throw(ArgumentError("Expected a ClimaSeaIce.SeaIceModel or sea-ice Simulation, got $(typeof(model))"))
```
    
(3) is more descriptive in case users accidentally swap the order of `ocean` and `sea_ice` in `OceanSeaIceModel`

simone-silvestri:
Ok, I ll merge that one then we can also add what implemented in this PR

src/EarthSystemModels/InterfaceComputations/component_interfaces.jl
```diff
@@ -358,7 +358,7 @@ function ComponentInterfaces(atmosphere, ocean, sea_ice=nothing;
                              freshwater_density = default_freshwater_density,
                              atmosphere_ocean_fluxes = SimilarityTheoryFluxes(eltype(exchange_grid)),
                              atmosphere_sea_ice_fluxes = atmosphere_sea_ice_similarity_theory(eltype(exchange_grid)),
-                             sea_ice_ocean_heat_flux = ThreeEquationHeatFlux(sea_ice),
+                             sea_ice_ocean_heat_flux = ThreeEquationHeatFlux(eltype(exchange_grid)),
```
simone-silvestri, line 366: are you sure about this change? `ThreeEquationHeatFlux` is specialized for a `ClimaSeaIce` simulation to include the internal heat flux in the balance of fluxes required by the interfacial temperature. If we use `eltype(exchange_grid)` we loose the internal conductive flux

src/SeaIces/sea_ice_simulation.jl
```diff
@@ -41,6 +41,10 @@ function sea_ice_simulation(grid, ocean=nothing;
                             internal_heat_flux = ConductiveFlux(; conductivity),
                             snow_thermodynamics = default_snow_thermodynamics(grid))
 
+    if !isnothing(ice_density)
+        sea_ice_density = ice_density
+    end
+
```
simone-silvestri:
```suggestion
```

src/SeaIces/sea_ice_simulation.jl
```diff
@@ -139,25 +143,32 @@ end
...
+sea_ice_model(sea_ice::Simulation) = sea_ice_model(sea_ice.model)
+sea_ice_model(model::SeaIceModel) = model
+
+sea_ice_model(model) = throw(ArgumentError("Expected a ClimaSeaIce.SeaIceModel or sea-ice Simulation, got $(typeof(model))"))
+
+sea_ice_thickness(sea_ice) = sea_ice_model(sea_ice).ice_thickness
+sea_ice_concentration(sea_ice) = sea_ice_model(sea_ice).ice_concentration
+
+heat_capacity(sea_ice) = sea_ice_model(sea_ice).phase_transitions.heat_capacity
+reference_density(sea_ice) = sea_ice_model(sea_ice).phase_transitions.density
```
simone-silvestri, line 151: maybe instead of having a `sea_ice_model` guard we can specialize these functions for 
```julia
heat_capacity(sea_ice::Simulation{<:SeaIceModel})
```
which is the more "julia" way

src/SeaIces/sea_ice_simulation.jl
```diff
@@ -173,14 +184,15 @@ function default_ai_temperature(sea_ice::Simulation{<:SeaIceModel})
...
-function ThreeEquationHeatFlux(sea_ice::Simulation{<:SeaIceModel}, FT::DataType = Oceananigans.defaults.FloatType;
+# Constructor that accepts the sea-ice model or a sea-ice Simulation.
+function ThreeEquationHeatFlux(sea_ice, FT::DataType = Oceananigans.defaults.FloatType;
                                heat_transfer_coefficient = 0.0095,
                                salt_transfer_coefficient = heat_transfer_coefficient / 35,
                                friction_velocity = convert(FT, 0.002))
 
-    conductive_flux = sea_ice.model.ice_thermodynamics.internal_heat_flux
-    ice_temperature = sea_ice.model.ice_thermodynamics.top_surface_temperature
+    model = sea_ice_model(sea_ice)
```
simone-silvestri, line 189: same as above, we probably should just specialize `ThreeEquationHeatFlux` to dispatch on 
```julia
ThreeEquationHeatFlux(sea_ice::Simulation{<:SeaIceModel}, ....)
```

taimoorsohail:
Thanks @simone-silvestri -- I think you are right about the `ThreeEquation` fix here. I will make the changes now.

taimoorsohail:
Actually @simone-silvestri I am closing this PR now. I think the only meaningful addition was the more verbose sea ice model error, and that can be added upstream to warn the user that the input order in `OceanSeaIceModel` is wrong. It doesn't need to happen in the `InterfaceComputations`. The choice of doing `OceanSeaIceModel(sea_ice, ocean; kwargs...)` vs `OceanSeaIceModel(ocean, sea_ice; kwargs...)` is probably better discussed in an issue anyway.

taimoorsohail:
Closed in favour of #230

PR #225, opened 2026-05-12
Set JULIA_CUDA_USE_COMPAT=false in test / docs

simone-silvestri:
This PR adds `JULIA_CUDA_USE_COMPAT=false` to both test and docs to get around the compatibility issues between CUDA drivers and older GPUs like Titan V GPUs

Edit: this PR also fixes a bug in `interpolate_bathymetry` exposed by https://github.com/CliMA/Oceananigans.jl/pull/5489 where we were using a folded topology for a `LatitudeLongitudeGrid` in `regrid_bathymetry`


.github/workflows/ci.yml
```diff
@@ -32,6 +32,7 @@ env:
   CDSAPI_KEY: ${{ secrets.CDSAPI_KEY }}
   DATADEPS_ALWAYS_ACCEPT: true
   JULIA_PKG_SERVER_REGISTRY_PREFERENCE: 'eager'
+  JULIA_CUDA_USE_COMPAT: false
```
giordano: It's not needed here

simone-silvestri:
```suggestion
```

simone-silvestri:
The downloading errors are independent from this PR, we are missing some data in NumericalEarthArtifacts following https://github.com/NumericalEarth/NumericalEarth.jl/pull/210 and the ECCO server is down. This PR should be ready to merge, then when the ECCO server is back up I ll add the necessary files to NumericalEarthArtifacts

giordano (approved):

src/Bathymetry/regrid_bathymetry.jl
```diff
@@ -369,7 +369,8 @@ function interpolate_bathymetry_in_passes(native_z, target_grid;
     Nφ = Int[Nφ..., Nφt]
 
     old_z  = native_z
-    TX, TY = topology(target_grid)
+    TXt, _, _  = topology(target_grid)
+    _, TYn, _  = topology(native_z.grid)
```
navidcy:
```suggestion
    TXt, _, _ = topology(target_grid)
    _, TYn, _ = topology(native_z.grid)
```

navidcy (approved):

PR #226, opened 2026-05-12
Avoid computing frazil fluxes when sea ice model does not exist

taimoorsohail:
This PR replaces frazil fluxes with zeros when no sea ice component exists to the Earth System Model. I have also added to the diagnostic tests to compute heat fluxes in an ocean only model to test for this type of issue in the future.


src/Diagnostics/interface_fluxes.jl
```diff
@@ -2,6 +2,13 @@
 @inline flux_field(condition) = condition
 @inline flux_field(bc::MultipleFluxes) = bc.flux_field
 @inline flux_field(bc::DiscreteBoundaryFunction) = flux_field(bc.func)
+@inline has_sea_ice_ocean_interface(esm::EarthSystemModel) = !isnothing(esm.interfaces.sea_ice_ocean_interface)
+
+function zero_flux(esm::EarthSystemModel)
+    zero_flux = Field{Center, Center, Nothing}(esm.interfaces.exchanger.grid)
+    fill!(zero_flux, zero(eltype(esm)))
+    return zero_flux
+end
```
simone-silvestri:
```suggestion
```

src/Diagnostics/interface_fluxes.jl
```diff
@@ -13,6 +20,8 @@
 Return the two-dimensional frazil temperature flux (K m s⁻¹) in a coupled `esm`.
 """
 function frazil_temperature_flux(esm::EarthSystemModel)
+    has_sea_ice_ocean_interface(esm) || return zero_flux(esm)
```
simone-silvestri:
```suggestion
    has_sea_ice_ocean_interface(esm) || return ZeroField()
```

src/Diagnostics/interface_fluxes.jl
```diff
@@ -68,6 +82,8 @@ end
 Return the two-dimensional frazil heat flux (W m⁻²) in a coupled `esm`.
 """
 function frazil_heat_flux(esm::EarthSystemModel)
+    has_sea_ice_ocean_interface(esm) || return zero_flux(esm)
```
simone-silvestri:
```suggestion
    has_sea_ice_ocean_interface(esm) || return ZeroField()
```

src/Diagnostics/interface_fluxes.jl
```diff
@@ -91,8 +107,13 @@ Return the sea ice-ocean heat flux (W m⁻²) at the sea ice-ocean interface
...
 function sea_ice_ocean_heat_flux(esm::EarthSystemModel)
-    sea_ice_ocean_heat_flux =
-        esm.interfaces.sea_ice_ocean_interface.fluxes.interface_heat + frazil_heat_flux(esm)
+    sea_ice_ocean_heat_flux = if has_sea_ice_ocean_interface(esm)
+        sea_ice_ocean_fluxes = esm.interfaces.sea_ice_ocean_interface.fluxes
+        sea_ice_ocean_fluxes.interface_heat + frazil_heat_flux(esm)
+    else
+        zero_flux(esm)
+    end
+
```
simone-silvestri:
```suggestion
    sea_ice_ocean_heat_flux = sea_ice_ocean_fluxes.interface_heat + frazil_heat_flux(esm)
```
I think this is ok if we return a `ZeroField` from frazil_heat_flux?

src/Diagnostics/interface_fluxes.jl
```diff
@@ -132,6 +153,8 @@ Return the sea ice-ocean salinity flux (g/kg m s⁻¹) at the sea ice-ocean inte
 in a coupled `esm`.
 """
 function sea_ice_ocean_salinity_flux(esm::EarthSystemModel)
+    has_sea_ice_ocean_interface(esm) || return zero_flux(esm)
```
simone-silvestri:
```suggestion
    has_sea_ice_ocean_interface(esm) || return ZeroField()
```

navidcy:
Does it make sense to use

https://github.com/NumericalEarth/NumericalEarth.jl/blob/ebffee2bd035e3d610352a4bc618c9d2c182da80/src/EarthSystemModels/EarthSystemModels.jl#L76-L84

instead of defining new method `has_sea_ice_ocean_interface`?

taimoorsohail:
Good, done, have a look @navidcy

navidcy:
CI errors:

```
  UndefVarError: `ZeroField` not defined in `NumericalEarth.Diagnostics`
```

You might need to import `ZeroField` from `Oceananigans.Fields`?

src/Diagnostics/Diagnostics.jl
```diff
@@ -16,7 +16,10 @@ using Oceananigans.Fields: FieldStatus
 using Oceananigans.Utils: launch!
 using KernelAbstractions: @index, @kernel
 using Oceananigans.BoundaryConditions: DiscreteBoundaryFunction
-using NumericalEarth.EarthSystemModels: EarthSystemModel
+using NumericalEarth.EarthSystemModels: EarthSystemModel,
+                                          NoSeaIceInterfaceModel,
+                                          NoOceanInterfaceModel,
+                                          NoInterfaceModel
```
navidcy:
```suggestion
using NumericalEarth.EarthSystemModels: EarthSystemModel,
                                        NoSeaIceInterfaceModel,
                                        NoOceanInterfaceModel,
                                        NoInterfaceModel
```

src/Diagnostics/interface_fluxes.jl
```diff
@@ -178,6 +191,9 @@ function sea_ice_ocean_freshwater_flux(esm::EarthSystemModel; reference_salinity
     return Field(sea_ice_ocean_freshwater_flux)
 end
 
+sea_ice_ocean_freshwater_flux(esm::NoSeaIceOceanInterfaceModel;
+                              reference_salinity = 35) = ZeroField()
```
navidcy:
```suggestion
sea_ice_ocean_freshwater_flux(esm::NoSeaIceOceanInterfaceModel; kwargs...) = ZeroField()
```

test/test_diagnostics_2.jl
```diff
@@ -92,5 +92,71 @@ for arch in test_architectures
             @test atmosphere_ocean_freshwater[1, 1, 1] ≈ - ρᵒᶜ / S₀ * (S_flux_value - sea_ice_ocean_salt_flux_value)
             @test net_ocean_freshwater[1, 1, 1] ≈ atmosphere_ocean_freshwater[1, 1, 1] + sea_ice_ocean_freshwater[1, 1, 1]
         end
+
+        ocean_only_ocean = ocean_simulation(grid;
```
navidcy: this is a weird name.... `ocean_only_ocean`. That's tautology?
Also that's the same as the `ocean` configured above... That was only an ocean that was only an ocean...

The "ocean-only" part comes at `ocean_only_esm = OceanOnlyModel(...)` so let's not create this whole sea of names that means nothing?

taimoorsohail: Thanks and sorry. Have refactored to loop over possible configs to avoid sea of variables

src/Diagnostics/interface_fluxes.jl
```diff
@@ -3,6 +3,15 @@
 @inline flux_field(bc::MultipleFluxes) = bc.flux_field
 @inline flux_field(bc::DiscreteBoundaryFunction) = flux_field(bc.func)
 
+const NoSeaIceOceanInterfaceModel = Union{NoSeaIceInterfaceModel,
+                                          NoOceanInterfaceModel,
+                                          NoInterfaceModel}
+
+function zero_surface_flux(esm::EarthSystemModel)
+    zero_flux = Field{Center, Center, Nothing}(esm.interfaces.exchanger.grid)
+    return set!(zero_flux, ZeroField())
```
navidcy: What does this:
```julia
set!(zero_flux, ZeroField())
```
do?

Why not

```suggestion
    return set!(zero_flux, 0)
```

But also, `zero_flux` is zero since it's just created... I don't understand.

taimoorsohail: I was unable to reference `ZeroField()` directly as suggested by @simone-silvestri because it is of location `{Nothing, Nothing, Nothing}` so setting the frazil fluxes directly with `ZeroField()` didn't seem to work. So I thought initialising an empty field and setting with `ZeroField()` would at least preserve that...

src/Diagnostics/interface_fluxes.jl
taimoorsohail: I was unable to reference ZeroField() directly because it is of location {Nothing, Nothing, Nothing} -- I think we need some sort of function that creates a zero-flux field

src/Diagnostics/interface_fluxes.jl
simone-silvestri: I think we don't need this function, any "zero" flux field should just be a [ZeroField](https://github.com/CliMA/Oceananigans.jl/blob/9ae1eb3e8f62ddf8f519f2f8a3cfa66e6b82a8ad/src/Fields/constant_field.jl#L4)

src/Diagnostics/interface_fluxes.jl
taimoorsohail: I was unable to reference ZeroField() directly because it is of location {Nothing, Nothing, Nothing} -- I think we need some sort of function that creates a zero-flux field

navidcy (approved):

PR #228, opened 2026-05-12
Add AbstractPrescribedComponent abstract type

navidcy:
This PR adds a PrescribedModelComponent abstract type and ensures that all PrescribedModelComponents are subtypes of this type.

This follows discussions in https://github.com/NumericalEarth/NumericalEarth.jl/pull/211#issuecomment-4433904092


src/Atmospheres/Atmospheres.jl
```diff
@@ -13,6 +13,7 @@ using Oceananigans.Utils
 using Adapt
 using Thermodynamics.Parameters: AbstractThermodynamicsParameters
 using KernelAbstractions: @kernel, @index
+using NumericalEarth.EarthSystemModels: PrescribedModelComponent
```
glwagner: I vote to use `AbstractPrescribedComponent`. The reason is that "model" is ambiguous here. Do you mean AtmosphereModel? EarthSystemModel?

Also, we agreed long ago that abstract types start with `Abstract***`.

So I propose writing AbstractPrescribedComponent.

navidcy: done

glwagner (approved):

PR #200, opened 2026-04-30
(0.4.0) Add radiation as a top-level EarthSystemModel component

glwagner:
## Summary

Closes #30. Wholesale refactor that promotes radiation from a `ComponentInterfaces` field into its own top-level `EarthSystemModel` component, mirroring the structure of #149 (land).

- New `src/Radiations/` module with `PrescribedRadiation`, `SurfaceRadiationProperties`, `InterfaceRadiationFlux`, the moved `LatitudeDependentAlbedo` / `TabulatedAlbedo` modules, and the radiation kernels.
- `EarthSystemModel` struct: type-param order `{R, A, L, I, O, F, C, Arch}`; `radiation` field placed above `atmosphere` ("from the sun"). Default `radiation = nothing` ⇒ radiatively decoupled (no upwelling LW, no SW/LW absorption).
- `StateExchanger` gains a `radiation` slot. `ComponentInterfaces` constructor takes `radiation` as a kwarg used only to construct the radiation exchanger.
- **Kernel split**: `_assemble_net_ocean_fluxes!` and `_assemble_net_sea_ice_fluxes!` now do only turbulent + sea-ice contributions. New `apply_air_sea_radiative_fluxes!` and `apply_air_sea_ice_radiative_fluxes!` (in `Radiations/`) add radiative contributions on top, dispatch wholesale on `coupled_model.radiation` type, and write to per-surface `InterfaceRadiationFlux` diagnostics.
- `SkinTemperature` consumes radiation via the new `air_sea_interface_radiation_state(rk, ex_state, i, j, k, grid, time)` getter (and sea-ice variant). When radiation is off the getter returns zeros, so `SkinTemperature` degrades cleanly to a turbulent-only flux balance.
- `PrescribedAtmosphere` loses `downwelling_radiation`; `TwoBandDownwellingRadiation` is removed entirely (committing to two-band).
- New data-wrangling constructors: `JRA55PrescribedRadiation`, `ECCOPrescribedRadiation`, `OSPapaPrescribedRadiation`.
- All tests, examples, experiments, docs, and the Reactant ext updated.

## Migration notes (breaking changes for v0.3)

- `Radiation()` is gone. To enable radiation, build a `PrescribedRadiation` (or one of the data-wrangling helpers) and pass it via `radiation = ...` to `OceanOnlyModel` / `OceanSeaIceModel` / `EarthSystemModel`.
[...]

PR #230, opened 2026-05-13
(0.5.0) Flip OceanSeaIceModel arg order to (ocean, sea_ice) and validate constructor args

taimoorsohail:
## Summary

Two changes to the `EarthSystemModel` convenience constructors:

1. **Breaking**: `OceanSeaIceModel` positional argument order is now `(ocean, sea_ice)` (previously `(sea_ice, ocean)`), so the order matches the constructor name. All in-repo callers — src/, examples/, experiments/, tests/, and docs/ — are updated.
2. **Argument validation** on `OceanOnlyModel`, `OceanSeaIceModel`, and `AtmosphereOceanModel`. If an obviously-wrong component is supplied in a positional slot — e.g. a sea-ice simulation where an ocean is expected — the constructor throws an `ArgumentError` pointing at the offending position, instead of erroring obscurely deep inside `InterfaceComputations`.

## Design

- **Blacklist, not whitelist.** Two predicates, `is_ocean_component` and `is_sea_ice_component`, default to `true`. Only the wrong type returns `false`: `is_sea_ice_component(::Simulation{<:HydrostaticFreeSurfaceModel}) = false` lives in `Oceans`; `is_ocean_component(::Simulation{<:SeaIceModel}) = false` lives in `SeaIces`. New ocean/sea-ice backends (Veros, future Speedy ocean, etc.) require no extension — the goal is to catch the canonical swap, not to police all inputs.
- **One error helper.** `invalid_component(constructor, position, expected, received)` formats the `ArgumentError` text consistently across all three constructors.

## Test plan

- [ ] CI passes
- [ ] `OceanSeaIceModel(ocean, sea_ice)` continues to work in existing tests/examples
- [ ] Supplying a swapped pair throws an `ArgumentError` naming the offending positional argument


navidcy:
Is this necessary? The `OceanSeaIceModel` docstring clearly states what goes where. 

I'm only asking because the PR then creates the question why a message for `OceanSeaIceModel` constructor and not other constructors?

taimoorsohail:
I guess not... Maybe I was just being dumb, but I'm pretty sure the order flipped very recently? Maybe this was a breaking change I wasn't aware of.

taimoorsohail:
I can close if people think this is too hand-holdy. I know it would have been helpful to me because the error that is thrown otherwise is in `InterfaceComputations` which is not at all clear...

navidcy:
> I guess not... Maybe I was just being dumb, but I'm pretty sure the order flipped very recently? Maybe this was a breaking change I wasn't aware of.

Not sure. If the order changed then it _should_ have been marked as breaking change; you are correct!

navidcy:
We might get a more informative error if we type the args in the OceanSeaIceModel constructor. What do you think?

taimoorsohail:
Do you mean something like `OceanSeaIceModel(; ocean=ocean, sea_ice=sea_ice, ...)`?

navidcy:
> Do you mean something like `OceanSeaIceModel(; ocean=ocean, sea_ice=sea_ice, ...)`?

No. I meant, enforce the type of the arguments. E.g.

https://github.com/NumericalEarth/NumericalEarth.jl/blob/6914080fc54fa78a3e55acb6d7ef20a05a544b50/src/EarthSystemModels/earth_system_model.jl#L304

we can have something like

```Julia
function OceanSeaIceModel(sea_ice::ClimaSeaIce.SeaIceModel, ocean::OceanModel; atmosphere=nothing, land=nothing, radiation=nothing...
```

The above would throw a more informative error if you give an ocean model as first argument; it won't need to reach the interface part to error; it will break immediately and tell you that it was expecting an arg of type X and got type Y.

But I do find that the order would be more intuitive if it was `ocean, sea_ice` to rhyme with `OceanSeaIce`Model...

navidcy:
@taimoorsohail I think the arg order changed in https://github.com/NumericalEarth/NumericalEarth.jl/pull/200 and it _was_ indeed marked as breaking.

navidcy:
Without typing on `main`:

```Julia
julia> OceanSeaIceModel(ocean, sea_ice; atmosphere, land, radiation)
ERROR: MethodError: no method matching NumericalEarth.EarthSystemModels.InterfaceComputations.ThreeEquationHeatFlux(::Simulation{…})
The type `NumericalEarth.EarthSystemModels.InterfaceComputations.ThreeEquationHeatFlux` exists, but no method is defined for this combination of argument types when trying to construct it.

Closest candidates are:
  NumericalEarth.EarthSystemModels.InterfaceComputations.ThreeEquationHeatFlux(::F, ::T, ::FT, ::FT, ::U) where {F, T, FT, U}
   @ NumericalEarth ~/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/Documents/Research/NumericalEarth.jl-v2/src/EarthSystemModels/InterfaceComputations/sea_ice_ocean_heat_flux_formulations.jl:120
  NumericalEarth.EarthSystemModels.InterfaceComputations.ThreeEquationHeatFlux(::Nothing; ...)
   @ NumericalEarth ~/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/Documents/Research/NumericalEarth.jl-v2/src/EarthSystemModels/InterfaceComputations/sea_ice_ocean_heat_flux_formulations.jl:163
  NumericalEarth.EarthSystemModels.InterfaceComputations.ThreeEquationHeatFlux(::Nothing, ::DataType; kwargs...)
   @ NumericalEarth ~/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/Documents/Research/NumericalEarth.jl-v2/src/EarthSystemModels/InterfaceComputations/sea_ice_ocean_heat_flux_formulations.jl:163
  ...

Stacktrace:
 [1] EarthSystemModel(radiation::PrescribedRadiation{…}, atmosphere::PrescribedAtmosphere{…}, land::PrescribedLand{…}, sea_ice::Simulation{…}, ocean::Simulation{…}; clock::Clock{…}, ocean_reference_density::Float64, ocean_heat_capacity::Float64, sea_ice_reference_density::Float64, sea_ice_heat_capacity::Float64, interfaces::Nothing)
   @ NumericalEarth.EarthSystemModels ~/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/Documents/Research/NumericalEarth.jl-v2/src/EarthSystemModels/earth_system_model.jl:191
 [2] EarthSystemModel(radiation::PrescribedRadiation{…}, atmosphere::PrescribedAtmosphere{…}, land::PrescribedLand{…}, sea_ice::Simulation{…}, ocean::Simulation{…})
   @ NumericalEarth.EarthSystemModels ~/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/Documents/Research/NumericalEarth.jl-v2/src/EarthSystemModels/earth_system_model.jl:146
 [3] OceanSeaIceModel(sea_ice::Simulation{…}, ocean::Simulation{…}; atmosphere::PrescribedAtmosphere{…}, land::PrescribedLand{…}, radiation::PrescribedRadiation{…}, kw::@Kwargs{})
   @ NumericalEarth.EarthSystemModels ~/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/Documents/Research/NumericalEarth.jl-v2/src/EarthSystemModels/earth_system_model.jl:295
 [4] top-level scope
   @ REPL[27]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

Not very clear what has happened indeed!

After typing the args:

```Julia
julia> OceanSeaIceModel(ocean, sea_ice; atmosphere, land, radiation)
ERROR: MethodError: no method matching OceanSeaIceModel(::Simulation{…}, ::Simulation{…}; atmosphere::PrescribedAtmosphere{…}, land::PrescribedLand{…}, radiation::PrescribedRadiation{…})
The function `OceanSeaIceModel` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  OceanSeaIceModel(::Simulation{<:SeaIceModel}, ::Simulation{<:HydrostaticFreeSurfaceModel}; atmosphere, land, radiation, kw...)
   @ NumericalEarth ~/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/Documents/Research/NumericalEarth.jl-v2/src/EarthSystemModels/earth_system_model.jl:298

Stacktrace:
 [1] top-level scope
   @ REPL[33]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

Better?

taimoorsohail:
Yes that's much better - at least I don't spend an hour trying to figure out what is wrong with the three equation model! Do you want me to close this PR and you open one with the types? Or I can modify this one? Whatever you prefer?

taimoorsohail:
Closed in favour of #231

navidcy:
On second thought, @glwagner is right in #231: we don't wanna restrict the types by typing the args.
So I'm reopening this.

navidcy:
But I suggest just validation of each argument and a message.
We don't need to give people the history/journal of the situation that at some point args were flipped etc.
We only to give them a message in the spirit: "The provided `arg1` isa `typeX` but the constructor expects `typeY`."

See, e.g., https://github.com/NumericalEarth/NumericalEarth.jl/pull/230/changes#r3231713544

src/EarthSystemModels/earth_system_model.jl
```diff
@@ -292,8 +301,15 @@ This is a convenience constructor for [`EarthSystemModel`](@ref) with an explici
...
 struct convention (top→bottom): `sea_ice` then `ocean`.
 """
-OceanSeaIceModel(sea_ice, ocean; atmosphere=nothing, land=nothing, radiation=nothing, kw...) =
+function OceanSeaIceModel(sea_ice, ocean; atmosphere=nothing, land=nothing, radiation=nothing, kw...)
+    if _is_ocean_component(sea_ice) && _is_sea_ice_component(ocean)
+        throw(ArgumentError("OceanSeaIceModel expects positional arguments in the order " *
+                            "`OceanSeaIceModel(sea_ice, ocean; ...)`, but received " *
+                            "an ocean first and sea ice second. Swap the first two arguments."))
+    end
+
```
navidcy: We want to validate each argument, not only check the case that the user gave correct arguments but in reverse order.

E.g. something along the lines of:

```Julia
if !(is_sea_ice_component(sea_ice))
    throw(ArgumentError("OceanSeaIceModel expects a sea ice simulation as its first positional argument"))
end

if !(is_ocean_component(ocean))
    throw(ArgumentError("OceanSeaIceModel expects an ocean simulation as its second positional argument"))
end

Comment (navidcy) [reply src/EarthSystemModels/earth_system_model.jl]: What do you think @taimoorsohail?

Comment (taimoorsohail) [reply src/EarthSystemModels/earth_system_model.jl]: Yes I like that! I agree the original warning felt unnecessarily wordy.

Comment (navidcy) [reply src/EarthSystemModels/earth_system_model.jl]: It's also too specific in the sense that the message would _only_ appear to the user if they give the correct args in wrong order. But if they provide an atmospheric simulation instead of ocean they wouldn't get a helpful message.

Comment (taimoorsohail) [reply src/EarthSystemModels/earth_system_model.jl]: Fair!

Comment (taimoorsohail) [reply src/EarthSystemModels/earth_system_model.jl]: OK I implemented this change -- have a look @navidcy

File: src/EarthSystemModels/earth_system_model.jl [hunk 2]
```diff
@@ -270,6 +270,15 @@ function present_surfaces(ocean, sea_ice)
     return Tuple(surfaces)
 end
 
+# Helper predicates for conditional logic in `EarthSystemModel` construction and diagnostics.
+
+_is_sea_ice_component(::SeaIceModel) = true
+_is_sea_ice_component(::Simulation{<:SeaIceModel}) = true
+_is_sea_ice_component(component) = false
```
simone-silvestri:
```suggestion
is_sea_ice_component(::SeaIceModel) = true
is_sea_ice_component(::Simulation{<:SeaIceModel}) = true
```

I think these should be extended in the `SeaIces` module (the fallback `is_sea_ice_component(component) = false` can stay here

src/EarthSystemModels/earth_system_model.jl
```diff
@@ -270,6 +270,15 @@ function present_surfaces(ocean, sea_ice)
...
+
+_is_ocean_component(::Simulation{<:HydrostaticFreeSurfaceModel}) = true
```
simone-silvestri:
```suggestion
is_ocean_component(::Simulation{<:HydrostaticFreeSurfaceModel}) = true
```
Also this should be in the Oceans module and extended also in `NumericalEarthVerosExt` for a `VerosOceanSimulation`

simone-silvestri:
What do you think about `valid_ocean_component` instead of `is_ocean_component`?

src/EarthSystemModels/earth_system_model.jl
simone-silvestri: and probably the `is_sea_ice_component(::SeaIceModel) = true` is redundant

simone-silvestri:
Also, I @glwagner's idea that it is probably better to blacklist rather than whitelist, so that we do not necessarily need to extend these functions all the time we add new components. But I am ok also with this (eventually we would need to blacklist also new components)

glwagner:
> What do you think about `valid_ocean_component` instead of `is_ocean_component`?

the word "valid" seems a little haughty to me

glwagner:
nobody wants to be invalid @simone-silvestri

src/EarthSystemModels/earth_system_model.jl
```diff
@@ -270,6 +270,15 @@ function present_surfaces(ocean, sea_ice)
...
+_is_ocean_component(component) = false
```
glwagner: why the underscores?

also I think this should use blacklisting to avoid a situation where new components cannot be used if they are not defined in NumericalEarth

navidcy: > why the underscores?

Yes, the underscores should go away

> this should use blacklisting to avoid a situation where new components cannot be used if they are not defined in NumericalEarth

What do you mean exactly? That the fallback is `true`? Then it beats the purpose because if you give a SeaIceModel in the arg expecting an OceanModel it won't complain, right? Or am I missing something?

glwagner: The concept I am proposing is have:

```julia
is_sea_ice_component(component) = true
is_sea_ice_component(::Simulation{<:HydrostaticFreeSurfaceModel}) = false
```

in this way, we can catch common mistakes, which if I understand correctly will achieve the goal of this PR. At the same time, with a neutral fallback, we allow new components to come in from elsewhere and we do not have to pretend we are the "masters of the valid components".

glwagner: this is called "blacklisting" because it amounts to a statement "everything is valid, EXCEPT for X".

taimoorsohail: I had a chat with @navidcy . I think this is a good idea, but it does substantially reduce the scope of the check so it catches potentially very few cases, and also has the potential to become outdated if types change. It begs the question of whether this PR is worth the effort?

glwagner: I do think it is worth it! What other cases are you worried about? We can add them.

taimoorsohail: Hm, I guess none at the moment, just wondering about if this is increasing future burden to update things. But you know what, I am happy to push it if everyone else is... @navidcy ?

navidcy:
I am leaning in closing this.

I don't see the purpose of just validating and giving an error only for a very specific case that somebody gives the args in reverse order.

taimoorsohail:
I agree, closing for now, happy to reopen if anyone feels strongly about it.

glwagner:
Are you two ok if I add this?

src/EarthSystemModels/earth_system_model.jl
glwagner: The nice thing about the blacklist approach is that it doesn't require any future maintainence. That's one of the big reasons to prefer the blacklist approach vs whitelist. For example, it will never become valid to use a sea ice model as an ocean model. 

On the other hand, hopefully the list of valid ocean models will grow in time. So a whitelist approach requires constant updating. But a blacklist approach does not.

taimoorsohail:
Yeah absolutely! I'll reopen if you are happy to push it.

taimoorsohail:
Actually I don't know how to reopen on my phone please feel free to do it

navidcy:
Please do @glwagner!

navidcy (approved):
I like it.
Bump patch/minor release?

PR #231, opened 2026-05-13
Type some of the convenience ESM constructors args for informative error messages

navidcy:
This PR types some of the constructor arguments to minimise ambiguous error messages as those discussed in https://github.com/NumericalEarth/NumericalEarth.jl/pull/230#issuecomment-4437039321

In place of https://github.com/NumericalEarth/NumericalEarth.jl/pull/230

### DIFF
--- a/src/EarthSystemModels/earth_system_model.jl
+++ b/src/EarthSystemModels/earth_system_model.jl
@@ -3,6 +3,7 @@ using Oceananigans.TimeSteppers: Clock
 using Oceananigans: SeawaterBuoyancy
 using ClimaSeaIce.SeaIceThermodynamics: melting_temperature
 using KernelAbstractions: @kernel, @index
+using ClimaSeaIce: SeaIceModel
 
 mutable struct EarthSystemModel{R, A, L, I, O, F, C, Arch} <: AbstractModel{Nothing, Arch}
     architecture :: Arch
@@ -16,6 +17,8 @@ mutable struct EarthSystemModel{R, A, L, I, O, F, C, Arch} <: AbstractModel{Noth
 end
 
 const ESM = EarthSystemModel
+const OceanSim = Oceananigans.Simulation{<:Oceananigans.HydrostaticFreeSurfaceModel}
+const SeaIceSim = Oceananigans.Simulation{<:SeaIceModel}
 
 function Base.summary(model::ESM)
     A = nameof(typeof(architecture(model)))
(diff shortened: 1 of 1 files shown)


src/EarthSystemModels/earth_system_model.jl
```diff
@@ -302,7 +305,7 @@ Construct a coupled atmosphere--ocean model.
 Convenience constructor for [`EarthSystemModel`](@ref) with an atmosphere and ocean
 but no sea ice. All keyword arguments are forwarded to `EarthSystemModel`.
 """
-AtmosphereOceanModel(atmosphere, ocean; land=nothing, radiation=nothing, kw...) =
+AtmosphereOceanModel(atmosphere, ocean::OceanSim; land=nothing, radiation=nothing, kw...) =
```
taimoorsohail, line 308: Shouldn't we type atmosphere here? Just to be consistent?

taimoorsohail (approved):

src/EarthSystemModels/earth_system_model.jl
```diff
@@ -16,6 +17,8 @@ mutable struct EarthSystemModel{R, A, L, I, O, F, C, Arch} <: AbstractModel{Noth
 end
 
 const ESM = EarthSystemModel
+const OceanSim = Oceananigans.Simulation{<:Oceananigans.HydrostaticFreeSurfaceModel}
```
glwagner, line 20: Are we sure we don't want to allow other models? What about veros or NonhydrostaticModel?

src/EarthSystemModels/earth_system_model.jl
```diff
@@ -16,6 +17,8 @@ mutable struct EarthSystemModel{R, A, L, I, O, F, C, Arch} <: AbstractModel{Noth
...
+const SeaIceSim = Oceananigans.Simulation{<:SeaIceModel}
```
glwagner, line 21: Why do we want to restrict this to only ClimaSeaIce?

glwagner:
I'm not sure I support the philosophy or design here. I think we would like to support as many models as possible.

glwagner:
I think we can achieve the objective here with a "blacklisting" technique, rather than "whitelisting". In other words, you can write a function

```julia
is_ocean_component(component) = true
is_ocean_component(::SeaIceSimulation) = false
```

this will allow you to catch some errors while allowing the code to be extended to support new models without requiring source code changes / whitelisting the new models.

src/EarthSystemModels/earth_system_model.jl
navidcy: We could, but that's a SpeedyWeather atmosphere and we don't have access to SpeedyWeather types.
I actually thought about it and then I couldn't do it.
Let me think more how to do it.

navidcy: Perhaps `::Oceananigans.Simulation{<:Any}`? It's not quite specific tho and it won't complain if you give an ocean simulation... :(

navidcy:
OK, @glwagner your suggestion aligns more with #230 approach.
And I see your point. I agree!

navidcy:
Close in favour of #230

Issue #232, opened 2026-05-13
Add map projection support to `set!` for `RectilinearGrid` targets


ewquon:
## Motivation

A common downstream use case is regional simulations on a Cartesian `RectilinearGrid` in meters, where a map projection (Lambert conformal conic, Mercator, polar stereographic) connects `(x m, y m) ↔ (λ°, φ°)`. Today, `set!(rectilinear_field, era5_metadatum)` doesn't error: NumericalEarth builds the source `Field` on its native `LatitudeLongitudeGrid` (degrees), the vertical-extent check passes, and `Oceananigans.interpolate!` runs — but every target node falls outside the source's lon/lat bounds, so the target is left at zero. We should make the projection explicit at the `set!` call site instead of relying on the user to spot the silent failure.

## Proposed API

```julia
projection = LambertConformalConic(central_longitude=-97.485, central_latitude=36.605,
                                   standard_parallels=(33, 45))
set!(rectilinear_field, era5_metadata; projection)
```

`set!` gains a `projection` kwarg. When non-nothing, NumericalEarth inverse-projects each target node `(x, y)` back to `(λ, φ)` before bilinearly sampling the source.

## Implementation tiers

- **Built-in, zero deps**: hand-rolled spherical-Earth `LambertConformalConic`, `Mercator`, `PolarStereographic`. ~30–50 lines each.
- **`Proj.jl` extension**: arbitrary EPSG codes, ellipsoidal Earth.

## Reference implementation

```julia
struct LCC
    n      :: Float64
    F      :: Float64
    ρ₀     :: Float64
    λ₀_rad :: Float64
    R      :: Float64

    # Inner constructor (not outer) — an outer constructor with the same
    # arity would be shadowed by the default field-wise constructor.
    function LCC(; central_latitude, central_longitude, standard_parallels,
                   earth_radius=6_371_000.0)
        φ₀r = deg2rad(central_latitude)
        φ₁r, φ₂r = deg2rad.(standard_parallels)
        n  = log(cos(φ₁r) / cos(φ₂r)) / log(tan(π/4 + φ₂r/2) / tan(π/4 + φ₁r/2))
        F  = cos(φ₁r) * tan(π/4 + φ₁r/2)^n / n
        ρ₀ = earth_radius * F / tan(π/4 + φ₀r/2)^n
        return new(n, F, ρ₀, deg2rad(central_longitude), earth_radius)
    end
end

[...]

Issue #143, opened 2026-03-24
Method to enforce "predownloading" of data, when desired


glwagner:
## Problem

Currently, datasets are downloaded on demand when `Field(metadata)`, `FieldTimeSeries(metadata)`, or `DatasetRestoring(metadata)` is called and data is not already present. This method is useful for writing scripts that will "work anywhere" on the first call without extra steps (eg downloading necessary data from somewhere, which might requiring diving into documentation...)

However, for expensive applications it can be useful to enforce the opposite; that a script will not run unless the data has already been downloading. Similarly, it'd be convenient to be able to take a script written by someone else, and automatically download all of the data it needs (eg on a login node) before submitting a job to a cluster. In other words we want

1. **Pre-downloading for cluster jobs**: When running simulations on a cluster, data should be downloaded on a login node *before* submitting a Slurm job, to avoid wasting compute resources (and because compute nodes may lack internet access).

2. **Guarding against accidental downloads**: For expensive or production simulations, there should be a way to guarantee that all data is already on disk — and error immediately if it isn't — rather than silently downloading mid-simulation.

## Proposed solution

Three layers, from low-level to high-level:

### 1. `NUMERICALEARTH_DATA` environment variable

Control download behavior at runtime:

| Value | Behavior |
|-------|----------|
| unset / `"auto"` | Download on demand (current default) |
| `"existing"` | Error if data isn't already on disk — never download |

Analogous to Julia's `--pkgimages=existing`. Usage:

```bash
NUMERICALEARTH_DATA=existing srun julia --project run_simulation.jl
```

Implementation: a check at the top of each `download_dataset` method that verifies files exist when in `existing` mode, and errors with a clear message listing the missing files.

### 2. Metadata extraction from high-level constructors

Refactor constructors like `ERA5PrescribedAtmosphere` so their metadata-creation logic is accessible separately:

```julia
# Returns Vector{Metadata} for all 8 ERA5 variables
metadata = ERA5PrescribedAtmosphere_metadata(;
    dataset = ERA5Hourly(),
    start_date = DateTime(2020, 1, 1),
    end_date = DateTime(2020, 12, 31),
    region = BoundingBox(longitude=(200, 220), latitude=(35, 55)))
```

This enables constructing metadata (cheap, no I/O) without building fields.

A convenience function for batch downloading:

```julia
download_datasets(metadata)               # Vector{Metadata}
download_datasets(metadata1, metadata2...) # varargs
```

[...]

Issue #141, opened 2026-03-24
Aggregate dataset downloads via Vector of Metadata


glwagner:
what about something like

```julia
download_dataset([T_meta, q_meta, u_meta, v_meta])
```

which we call after building the `Metadata` ?

_Originally posted by @glwagner in https://github.com/NumericalEarth/NumericalEarth.jl/pull/93#discussion_r2927613738_

glwagner:
cc @ewquon

PR #233, opened 2026-05-13
NestedModel abstraction + ERA5 regional hindcast example

ewquon:
## Summary

@glwagner @kaiyuan-cheng

Adds `examples/era5_breeze.jl` — a building block for a regional modeling example that will eventually couple Breeze (compressible solver, in development) to forthcoming SlabLand and SlabOcean components.

We download ERA5 reanalysis over an SGP-centered LAM bounding box (HI-SCALE 2016-09-10 case day) and interpolate onto a `LatitudeLongitudeGrid` sized for ~3 km horizontal cells at the domain center latitude. `Tᵥ` is computed as a derived field using `Breeze.ThermodynamicConstants` for the `Rᵥ/Rd − 1` coefficient.

Punch list
- [x] Breeze model construction
- [x] open boundary conditions
- [ ] dynamical initialization (DFI)
- [ ] terrain
- [ ] acoustic substepping over terrain
- [ ] land/ocean coupling, MOST

## Notes

- Surface pressure is read onto a 3-D grid with `Nz=1` rather than a 2-D `(Bounded, Bounded, Flat)` grid, sidestepping `CliMA/Oceananigans.jl#5473` (Flat↔non-Flat `interpolate!` errors with a cryptic `BoundsError`; [#5474](https://github.com/CliMA/Oceananigans.jl/pull/5474) will convert that to a clear `ArgumentError` but does not lift the restriction). Mirrors the pattern in `examples/ERA5_hourly_data.jl`.
- This example uses a `LatitudeLongitudeGrid`. A future variant on a projected Cartesian grid would benefit from the `set!` projection support proposed in #232.

## Test plan
- [x] Run end-to-end against the CDS API on a machine with `~/.cdsapirc` configured.


examples/era5_breeze.jl
```diff
@@ -0,0 +1,216 @@
...
+# - [ ] open boundary conditions
+# - [ ] land/ocean coupling
+# - [ ] terrain
+
+using NumericalEarth
+using NumericalEarth.DataWrangling: BoundingBox, Metadatum, download_dataset
+using NumericalEarth.DataWrangling.ERA5
+using CDSAPI  # activates NumericalEarthCDSAPIExt for download_dataset
+using Oceananigans
+using Oceananigans.Fields: CenterField, XFaceField, YFaceField
```
glwagner: I think these are exported by `using Oceananigans` right?

examples/era5_breeze.jl
```diff
@@ -0,0 +1,216 @@
...
+# - [ ] initial state setting (set! the model from ingested fields)
+# - [ ] acoustic substepping
+# - [ ] dynamical initialization
+# - [ ] open boundary conditions
+# - [ ] land/ocean coupling
+# - [ ] terrain
+
+using NumericalEarth
+using NumericalEarth.DataWrangling: BoundingBox, Metadatum, download_dataset
+using NumericalEarth.DataWrangling.ERA5
```
glwagner: can we add the appropriate exports so that these additional outputs are not needed?

also for the example, I recommend avoiding manual calls to `download_dataset`. I don't want to codify this workflow (we should instead develop the utility on #141). The issue is that people will adopt this style and not realize that it isn't recommended anymore once we have the utility

examples/era5_breeze.jl
```diff
@@ -0,0 +1,216 @@
...
+using CDSAPI  # activates NumericalEarthCDSAPIExt for download_dataset
+using Oceananigans
+using Oceananigans.Fields: CenterField, XFaceField, YFaceField
+using Breeze
+using Breeze.Thermodynamics
```
glwagner: what's this for?

examples/era5_breeze.jl
```diff
@@ -0,0 +1,216 @@
...
+function build_z_faces()
+    zᶠ = zeros(Nz + 1)
+    for k in 1:Nz_strch
+        zᶠ[k+1] = zᶠ[k] + Δz_surf * r_z^(k-1)
+    end
+    for k in (Nz_strch+1):Nz
+        zᶠ[k+1] = zᶠ[k] + Δz_top
+    end
+    return zᶠ
+end
```
glwagner: can we use one of the bulit-in vertical discretization constructors?

examples/era5_breeze.jl
```diff
@@ -0,0 +1,216 @@
...
+qᵛ = CenterField(grid)
+qᶜ = CenterField(grid)
+qⁱ = CenterField(grid)
+
+set!(u,  Metadatum(:eastward_velocity;                   dataset=ds_pl, date=start_date, meta_common...))
+set!(v,  Metadatum(:northward_velocity;                  dataset=ds_pl, date=start_date, meta_common...))
+set!(T,  Metadatum(:temperature;                         dataset=ds_pl, date=start_date, meta_common...))
+set!(qᵛ, Metadatum(:specific_humidity;                   dataset=ds_pl, date=start_date, meta_common...))
+set!(qᶜ, Metadatum(:specific_cloud_liquid_water_content; dataset=ds_pl, date=start_date, meta_common...))
+set!(qⁱ, Metadatum(:specific_cloud_ice_water_content;    dataset=ds_pl, date=start_date, meta_common...))
```
glwagner: do we need a `MetadataSet` object? I can open an issue about this

glwagner: why don't we use `set!(model, ...)` here?

ewquon: This is something we've discussed before. That would clean things up here

examples/era5_breeze.jl
ewquon: The utility in #141 is based on `download_dataset`, I think you mean https://github.com/NumericalEarth/NumericalEarth.jl/issues/143

ewquon: When the new utility lands, we can revise this example as well as `ERA5_hourly_data.jl`

examples/era5_breeze.jl
ewquon: We don't have a model yet in the example 😅

examples/era5_breeze.jl
ewquon: Correct!

examples/era5_breeze.jl
glwagner: there's one at the bottom isn't there?

glwagner: https://github.com/NumericalEarth/NumericalEarth.jl/blob/5463396503db24eef7c72bbaed329bd7a85489e7/examples/era5_breeze.jl#L202-L204

glwagner: you can use `set!(atmos.model, ....)`

examples/era5_breeze.jl
ewquon: yep :)

examples/era5_breeze.jl
glwagner: on it

examples/era5_breeze.jl
glwagner: Until then, let's avoid anything special (it will download automatically under the hood) so we don't have to revert it or change the documentation

examples/era5_breeze.jl
ewquon: `vapor_gas_constant` isn't exported from `Breeze.Thermodynamics`

src/DataWrangling/DataWrangling.jl
```diff
@@ -4,7 +4,8 @@ restoring, or validation.
 """
 module DataWrangling
 
-export Metadata, Metadatum, DatewiseFilename, ECCOMetadatum, EN4Metadatum, all_dates, first_date, last_date
+export Metadata, Metadatum, DatewiseFilename, ECCOMetadatum, EN4Metadatum
+export download_dataset, all_dates, first_date, last_date
```
glwagner: I propose removing manual calls to `download_dataset` for now and not exporting the function. I think this API will change in the near future. We can rely on autodownloading for now.

examples/era5_breeze.jl
ewquon: https://github.com/NumericalEarth/Breeze.jl/pull/699

examples/era5_breeze.jl
ewquon: Yes I know, that was Claude getting ahead of itself. But there are some issues with pressure, needed to convert T to theta https://github.com/NumericalEarth/NumericalEarth.jl/issues/236

src/DataWrangling/DataWrangling.jl
ewquon: done

ewquon:
Okay, the model state should be ready for dynamic initialization.
era5_breeze_profiles

A couple of important notes:
1. As noted in #236, we need to better address geopotential height variability and we should discuss the approach to this. I've prototyped per-column vertical interpolation to unblock the ICs work and to have as a reference when developing the data wrangling improvement.
2. Because we don't have terrain yet, I'm using the _height above ground level_ to interpolate onto the flat-bottomed Breeze grid. (Otherwise, portions of the Breeze domain would be essentially underground and I imagine this would be problematic for the initialization)

glwagner:
@ewquon should we add terrain? Terrain following should work for the fully compressible formulation (just not substepping)

ewquon:
> @ewquon should we add terrain? Terrain following should work for the fully compressible formulation (just not substepping)

Hallelujah! I'm happy to add that in. Anything that gets us closer to reality sooner rather than later is a win in my book.

examples/era5_breeze.jl
glwagner, line 1: this needs to be added to `docs/make.jl`

docs/make.jl
```diff
@@ -35,7 +35,8 @@ examples = [
     Example("Global climate simulation", "global_climate_simulation", false),
     Example("Veros ocean simulation", "veros_ocean_forced_simulation", false),
     Example("Breeze over two oceans", "breeze_over_two_oceans", false),
-    Example("ERA5 hourly data", "ERA5_hourly_data", true),
+    Example("ERA5 hourly data retrieval", "ERA5_hourly_data", true),
+    Example("ERA5 hindcast simulation", "era5_breeze", true),
```
glwagner:
```suggestion
    Example("Atmosphere hindcast nested in ERA5", "era5_breeze", false),
```

glwagner: This will be a heavy case so we shoudln't run every commit (change true to false). Also we should clarify, this is an atmosphere case (eventually atmosphere-land), and probably should also indicate it uses Breeze

glwagner:
@ewquon I added the "build all examples" label so that it will build even if the "build always" tag is false.

glwagner:
@giordano do you know what the GPU error is?

https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/25867031203/job/76011977037?pr=233#step:2:211

giordano:
I believe
```
  failed to register layer: write /usr/local/share/julia/artifacts/75ae42466e7dd21641ff521f63cfb456d09433e2/lib/libcudnn_engines_precompiled.so.9: no space left on device
  Warning: Docker pull failed with exit code 1, back off 8.515 seconds before retry.
```
simply happens when the runner was reused from a previous run from somewhere else in the organisation, but the new build uses a new Docker image than the previous run, and then it fails to pull the second one because the combination of the two images fills the partition.

In summary, it's an inoffensive, albeit annoying, issue, restarting the job should succeed (unless you rerun in the same situation :smile:)

giordano:
https://github.com/NumericalEarth/Breeze.jl/pull/715 recovers some space on the VMs also in the jobs we run in the Breeze.jl repository (at the moment we do the cleanup only in this repo), _**hopefully**_ this will make the error less frequent.

glwagner:
working syntax (updated for the latest API on the branch — `child_simulation` + `NestedSimulation` wrap a `NestedModel`):

```julia
using NumericalEarth, Oceananigans, Breeze
using Breeze.AnelasticEquations: AnelasticDynamics
using Breeze.Thermodynamics: ReferenceState

# Lamb-Oseen vortex translating at U in x
const Γ, a, U, x₀, y₀ = 5e3, 200.0, 10.0, 600.0, 1000.0
const ρ̄ = 1.225

function lamb_oseen_uv(x, y, t)
    dx, dy = x - x₀ - U*t, y - y₀
    r² = dx^2 + dy^2
    uθ_over_r = r² < eps() ? 0.0 : (Γ / (2π * r²)) * (1 - exp(-r² / a^2))
    return (U - uθ_over_r * dy, uθ_over_r * dx)
end

# Parent: coarser and strictly larger than the child, so the FTS brackets every
# child sampling node (Interpolated BC + Relaxation-on-FTS both require this).
parent_grid = RectilinearGrid(size = (40, 16, 8),
                              x = (-500, 4500), y = (-500, 2500), z = (-20, 120),
                              topology = (Bounded, Periodic, Bounded))

parent = PrescribedAtmosphere(parent_grid, collect(0.0:5.0:305.0))
set!(parent.velocities.u, (x, y, z, t) -> ρ̄ * lamb_oseen_uv(x, y, t)[1])  # ρ̄·u → Breeze momentum BC
set!(parent.velocities.v, (x, y, z, t) -> ρ̄ * lamb_oseen_uv(x, y, t)[2])

# Child: finer LAM grid
child_grid = RectilinearGrid(size = (128, 32, 4),
                             x = (0, 4000), y = (0, 2000), z = (0, 100),
                             topology = (Bounded, Periodic, Bounded))

# Sponge mask: ramps from 0 in the interior to 1 in the outer 15% of x.
const Lx, x_sponge = 4000.0, 0.85 * 4000.0
sponge_mask(x, y, z) = x < x_sponge ? 0.0 : ((x - x_sponge) / (Lx - x_sponge))^2

# child_simulation builds the Breeze child model, wires Open BCs from the parent
# on the requested sides, and adds Relaxation-on-FTS interior nudging.
# It returns the constructed AtmosphereModel.
ref   = ReferenceState(child_grid; surface_pressure = 101325.0, potential_temperature = 300.0)
child = child_simulation(AtmosphereModel, child_grid, parent;
                    sides           = (:west, :east),
                    relaxation_rate = 1 / 60,        # 60-second sponge timescale
                    relaxation_mask = sponge_mask,
                    dynamics    = AnelasticDynamics(ref),
                    formulation = :LiquidIcePotentialTemperature,
                    advection   = WENO(order = 5))

set!(child; θ  = 300.0,
            ρu = (x, y, z) -> ρ̄ * lamb_oseen_uv(x, y, 0)[1],
            ρv = (x, y, z) -> ρ̄ * lamb_oseen_uv(x, y, 0)[2],
            ρw = 0.0)

# NestedSimulation(parent, child; sim_kwargs...) is a convenience for
# `Simulation(NestedModel(parent, child); sim_kwargs...)`. The NestedModel's
# `time_step!` steps the child, then ticks the parent's clock to match.
nested = NestedSimulation(parent, child; Δt = 0.2, stop_time = 300.0)
run!(nested)
```

What's in play:

- `child_simulation(modeltype, grid, parent; …)` builds the child model and wires up its BCs and forcings from the parent. Returns the constructed `AbstractModel`. Forwards `model_kwargs` to the model constructor.
- Variable mapping is chosen by dispatch on `default_parent_variables(modeltype, parent)`. `Oceananigans.NonhydrostaticModel` → `(u, v) ← parent.velocities.(u, v)` (core). `Breeze.AtmosphereModel` → `(ρu, ρv) ← parent.velocities.(u, v)` (in `NumericalEarthBreezeExt`, so Breeze stays a weakdep).
- BCs go through an internal `Interpolated(fts)` wrapper that lets `OpenBoundaryCondition` consume an FTS directly. Standard Oceananigans regularization tags it with side/dim/location; our `getbc` does trilinear-in-space + linear-in-time interpolation against the parent grid.
- `relaxation_rate` / `relaxation_mask` route through Oceananigans' native FTS-`Relaxation` path (CliMA/Oceananigans.jl#5575) for interior nudging.
- `NestedModel(parent, child::AbstractModel) <: AbstractModel` pairs the two and overrides `time_step!` to step the child, then advance the parent. `NestedSimulation(parent, child; sim_kwargs...)` is a one-liner for `Simulation(NestedModel(parent, child); sim_kwargs...)`.

Needs Breeze on `eq/atmosphere-model-open-bc-velocity` (PR NumericalEarth/Breeze.jl#722) and Oceananigans main (post-0.108.0) for FTS-`Relaxation`.

examples/era5_breeze.jl
```diff
@@ -529,14 +530,19 @@ davies = (u  = Relaxation(rate = 1/τ_relax, mask = lateral_mask, target = paren
...
+# (radiatively decoupled) until materialized inside an `EarthSystemModel`.
 
 p̄₀ = mean(interior(p₀))
 
 model = atmosphere_simulation(grid;
                               thermodynamic_constants = constants,
                               dynamics            = CompressibleDynamics(; surface_pressure = p̄₀),
                               boundary_conditions = bcs,
-                              forcing             = davies)
+                              forcing             = davies).model
```
glwagner: 👀

ewquon: Change to the `atmosphere_simulation` API?

ewquon:
**Nesting (i.e., online parent-child boundary coupling) demonstration** based on Breeze's baroclinic wave case: This is an idealized nesting configuration in which the child grid is a subset of the global parent, with matching resolution. The offline case involves sampling from the original global Breeze example and running the LAM using the ERA5 path laid out in this PR.

There is one time-stepping caveat, issue pending.

https://github.com/user-attachments/assets/a524a095-c8fc-41c0-b731-db6befa02b64

src/Atmospheres/prescribed_atmosphere.jl
```diff
@@ -29,17 +29,36 @@ function Base.show(io::IO, pa::PrescribedAtmosphere)
...
-    ua = FieldTimeSeries{Center, Center, Nothing}(grid, times)
-    va = FieldTimeSeries{Center, Center, Nothing}(grid, times)
-    return (u=ua, v=va)
+# `volumetric` selects the atmosphere's field set. The default (`false`) builds a
+# conventional *surface* atmosphere — 2D `(Center, Center, Nothing)` fields, the
+# form the ocean / sea-ice coupling expects, regardless of how many vertical cells
+# the grid carries. `volumetric = true` builds 3D `(Center, Center, Center)` fields
+# (adding `w`, dropping the surface freshwater flux), used when a
+# `PrescribedAtmosphere` plays the role of a `NestedSimulation` parent.
+function default_atmosphere_velocities(grid, times; volumetric=false)
```
simone-silvestri: what about `three_dimensional`? It;s probablky a bit clearer

simone-silvestri: in this way we can reduce the outer comment. If we need too much comments means the code is not clear enough :)

src/Atmospheres/prescribed_atmosphere.jl
```diff
@@ -120,30 +147,41 @@ EarthSystemModels.adopt_clock(atmosphere::PrescribedAtmosphere, clock) = EarthSy
 """
     PrescribedAtmosphere(grid, times=[zero(grid)];
                          clock = Clock{Float64}(time = 0),
+                         volumetric = false,
```
simone-silvestri: actually, since it is a keywork argument of `PrescribedAtmosphere` we probably want to pass `two_dimensional = true`

src/DataWrangling/set_region_data.jl
```diff
@@ -43,9 +43,20 @@ struct AverageNorthSouth end
...
-@inline mangle(i, j, k, data, ::ShiftSouth) = @inbounds data[i, max(j - 1, 1), k]
-@inline mangle(i, j, k, data, ::AverageNorthSouth) = @inbounds (data[i, j, k] + data[i, j + 1, k]) / 2
+#
+# Read indices are clamped to the data extent. A `BoundingBox` whose edge does
+# not land on a native cell face can produce a target grid one cell larger than
+# the downloaded data (CDS snaps the request outward to its 0.25° grid, and the
+# latitude axis is half-cell staggered). The clamp reads the nearest edge cell
+# there — the "clamp outside the hull" behaviour `restrict` documents — instead
+# of reading past the array bounds into uninitialized memory. For every read
+# that already lands in bounds (i.e. all aligned regions) the clamp is a no-op.
```
simone-silvestri:
```suggestion
# Clamp indices to avoid out-of-bounds access
```

src/EarthSystemModels/NestedSimulations/child_simulation.jl
```diff
@@ -0,0 +1,124 @@
+#####
+##### child_simulation: one-call construction of a parent-driven child model
+#####
+#
+# Despite the name (chosen for consistency with `ocean_simulation` /
+# `atmosphere_simulation`), `child_simulation` returns an `AbstractModel` — the
+# Breeze `atmosphere_simulation` helper returns a model, not a `Simulation`,
```
simone-silvestri: is this true? Would it be possible to have consistent naming? It might get confusing if the behavior is bit lost in here in a comment

simone-silvestri: actually would be nice to not need extensive commenting

src/EarthSystemModels/NestedSimulations/child_simulation.jl
```diff
@@ -0,0 +1,124 @@
...
+density-weighted, etc. — are determined by what the child consumes and what
+the parent provides. For example, Breeze `AtmosphereModel` consumes
+prognostic momentum (`ρu`, `ρv`, …), so a `PrescribedAtmosphere` parent
+populated for that use case must store ρ·u in its velocity slots; the
+method body documents the convention.
+"""
+function parent_variables end
+
+# Generic fallback — errors with a message that explains the situation and
+# suggests the two ways out.
```
simone-silvestri:
```suggestion
# Generic fallback — errors with a descriptive message.
```

src/EarthSystemModels/NestedSimulations/child_simulation.jl
```diff
@@ -0,0 +1,124 @@
...
+
+"""
+    _build_child_model(modeltype, grid; kwargs...)
+
+Construct the child model. Defaults to `modeltype(grid; kwargs...)`; package
+extensions override this for specific model types (e.g. the Breeze ext routes
+`AtmosphereModel` through `atmosphere_simulation` so Breeze's default
+advection / microphysics apply).
+"""
+_build_child_model(modeltype, grid; kwargs...) = modeltype(grid; kwargs...)
```
simone-silvestri:
```suggestion
build_child_model(modeltype, grid; kwargs...) = modeltype(grid; kwargs...)
```

src/EarthSystemModels/NestedSimulations/child_simulation.jl
```diff
@@ -0,0 +1,124 @@
...
+        "  • pass `variables = (...)` explicitly to `child_simulation`.\n",
+        "\n",
+        "Note: cross-realm nesting (e.g. an ocean child driven by an atmosphere ",
+        "parent) is intentionally unsupported as a default — the variable units ",
+        "do not match. Define a custom mapping only if you own the semantics."
+    )))
+end
+
+"""
+    _build_child_model(modeltype, grid; kwargs...)
```
simone-silvestri:
```suggestion
    build_child_model(modeltype, grid; kwargs...)
```

src/EarthSystemModels/NestedSimulations/child_simulation.jl
```diff
@@ -0,0 +1,124 @@
...
+
+If `relaxation_rate` is supplied, [`parent_forcings`](@ref) builds an
+Oceananigans `Relaxation` per variable that nudges the child interior toward
+the parent at rate `relaxation_rate`, weighted by `relaxation_mask` (default
+uniform). Per-variable values are supported via `NamedTuple`s. The resulting
+forcings are merged with any `forcing` already passed in `model_kwargs`.
+
+Wrap with `NestedSimulation(parent, child; Δt, stop_time, …)` to integrate.
+"""
+function child_simulation(modeltype, grid, parent;
```
simone-silvestri: so this should be
```suggestion
function child_model(modeltype, grid, parent;
```?

File: src/EarthSystemModels/NestedSimulations/interpolated_fts_boundary.jl [hunk 17]
```diff
@@ -0,0 +1,157 @@
...
+    LX = typeof(loc[1])
+    LY = typeof(loc[2])
+    LZ = typeof(loc[3])
+    _validate_source_bracket(c.source, grid, LX, LY, LZ)
+    return Interpolated{dim, SideType, LX, LY, LZ, typeof(c.source), typeof(c.source_grid)}(c.source, c.source_grid)
+end
+
+# Match the strict "source must bracket every child sampling node" check that
+# Oceananigans uses for `Relaxation`-on-FTS. Same logic for FTS and AbstractField.
+function _validate_source_bracket(source, grid, ::Type{LX}, ::Type{LY}, ::Type{LZ}) where {LX, LY, LZ}
```
simone-silvestri:
```suggestion
function validate_source_bracket(source, grid, ::Type{LX}, ::Type{LY}, ::Type{LZ}) where {LX, LY, LZ}
```

src/EarthSystemModels/NestedSimulations/interpolated_fts_boundary.jl
```diff
@@ -0,0 +1,157 @@
...
+
+#####
+##### Regularization: convert the location-less wrapper to the tagged form.
+#####
+
+function regularize_boundary_condition(c::Interpolated{Nothing}, grid, loc, dim, SideType, args...)
+    LX = typeof(loc[1])
+    LY = typeof(loc[2])
+    LZ = typeof(loc[3])
+    _validate_source_bracket(c.source, grid, LX, LY, LZ)
```
simone-silvestri:
```suggestion
    validate_source_bracket(c.source, grid, LX, LY, LZ)
```

src/EarthSystemModels/NestedSimulations/nested_model.jl
```diff
@@ -0,0 +1,92 @@
...
+# `Relaxation` forcings that reference the parent's `FieldTimeSeries`. Wrapping
+# the child in `NestedModel` lifts the parent-sync into the integrator itself
+# (where it belongs — it's not output behavior), so a plain
+# `Simulation(NestedModel(parent, child))` just works.
+#
+# All Oceananigans model-protocol calls (`clock`, `grid`, `fields`,
+# `prognostic_fields`, `update_state!`, ...) forward to the child via
+# `Base.getproperty` plus a small set of explicit dispatches. Only
+# `time_step!` carries new behavior — step the child, then advance the parent
+# by the elapsed Δt.
```
simone-silvestri: all this probably belongs in a docs page?

src/EarthSystemModels/NestedSimulations/nested_model.jl
```diff
@@ -0,0 +1,92 @@
...
+    NestedModel{typeof(parent), typeof(child), TS, A}(parent, child)
+
+# Property forwarding: `nested.clock`, `nested.grid`, `nested.velocities`, etc.
+# all resolve through the child. Only `parent` and `child` are direct fields.
+@inline function Base.getproperty(nm::NestedModel, name::Symbol)
+    if name === :parent || name === :child
+        return getfield(nm, name)
+    end
+    return getproperty(getfield(nm, :child), name)
+end
```
simone-silvestri, line 52: so the property of a nested model are redirected to the child? Maybe the nested model should have its own clock? Also this means that `nested.grid` is the grid of the child model?

src/EarthSystemModels/NestedSimulations/interpolated_fts_boundary.jl
```diff
@@ -0,0 +1,157 @@
...
+    Oceananigans.Fields.interpolate(X, source, loc, source_grid)
+
+@inline _boundary_index(::Type{LeftBoundary},  N) = 1
+@inline _boundary_index(::Type{RightBoundary}, N) = N + 1
+
+# `fill_halo_regions!` is sometimes invoked without a clock — most notably
+# during `set!`-time IC setup, where the kernel is launched with an empty
+# `args` tuple. Default `time = 0` in that case so dispatch succeeds.
+@inline _clock_time(clock) = clock.time
+@inline _clock_time(::Nothing) = 0.0
```
simone-silvestri:
```suggestion
@inline clock_time(clock) = clock.time
@inline clock_time(::Nothing) = 0.0
```

src/Atmospheres/prescribed_atmosphere.jl
```diff
@@ -64,7 +83,11 @@ PrescribedPrecipitationFlux(; rain=nothing, snow=nothing) =
 Adapt.adapt_structure(to, ff::PrescribedPrecipitationFlux) =
     PrescribedPrecipitationFlux(adapt(to, ff.rain), adapt(to, ff.snow))
 
-function default_freshwater_flux(grid, times)
+# Surface freshwater fluxes are meaningless for a volumetric atmosphere acting as
+# a nesting parent; default to `nothing` there. The surface_*_flux accessors and
+# `extract_field_time_series` already handle the Nothing branch.
+function default_freshwater_flux(grid, times; volumetric=false)
```
simone-silvestri: ERA5 does not come with 2D freshwater fluxes even if other fields are 3D?

src/Atmospheres/prescribed_atmosphere.jl
ewquon: Fair point!

src/EarthSystemModels/NestedSimulations/child_simulation.jl
```diff
@@ -0,0 +1,124 @@
...
+    rest_kwargs    = Base.structdiff(model_kwargs_nt, NamedTuple{(:forcing,)})
+
+    forcing = if relaxation_rate === nothing
+        user_forcing
+    else
+        relaxation = parent_forcings(; variables, rate = relaxation_rate, mask = relaxation_mask)
+        merge(user_forcing, relaxation)
+    end
+
+    return _build_child_model(modeltype, grid;
```
ewquon:
```suggestion
    return build_child_model(modeltype, grid;
```

src/EarthSystemModels/NestedSimulations/child_simulation.jl
```diff
@@ -0,0 +1,124 @@
...
+                     variables = parent_variables(modeltype, parent),
+                     relaxation_rate = nothing,
+                     relaxation_mask = 1,
+                     model_kwargs...)
+
+Construct a child model (of type `modeltype`) on `grid` with Open boundary
+conditions driven by `parent`. Returns the constructed `AbstractModel`.
+
+`parent_boundary_conditions` wires the BCs from `variables`; `model_kwargs`
+are forwarded to the underlying constructor (see `_build_child_model`).
```
ewquon:
```suggestion
are forwarded to the underlying constructor (see `build_child_model`).
```

src/EarthSystemModels/NestedSimulations/child_simulation.jl
```diff
@@ -0,0 +1,124 @@
...
+#     nested = NestedSimulation(parent, child; Δt, stop_time, …)
+#
+# Variable mapping (which parent FTS drives which child boundary) dispatches on
+# `parent_variables(modeltype, parent)` — multi-dispatch on the (child, parent)
+# pair. Methods live wherever both types are visible (often a package
+# extension). Cross-realm pairs (e.g. ocean child × atmosphere parent) are
+# intentionally undefined; the generic fallback throws an explanatory
+# `ArgumentError` and the user must pass `variables=` explicitly to override.
+#
+# The underlying model constructor is selected by `_build_child_model(modeltype, grid; …)`
```
ewquon:
```suggestion
# The underlying model constructor is selected by `build_child_model(modeltype, grid; …)`
```

src/EarthSystemModels/NestedSimulations/interpolated_fts_boundary.jl
```diff
@@ -0,0 +1,157 @@
...
+    j = _boundary_index(S, grid.Ny)
+    X = node(i, j, k, grid, LX(), Face(), LZ())   # boundary face in the y-normal direction
+    return _query_source(bc.source, bc.source_grid, X, (LX(), LY(), LZ()), _clock_time(clock))
+end
+
+@inline function getbc(bc::Interpolated{3, S, LX, LY, LZ},
+                       i::Integer, j::Integer, grid::AbstractGrid, clock=nothing, args...) where {S, LX, LY, LZ}
+    k = _boundary_index(S, grid.Nz)
+    X = node(i, j, k, grid, LX(), LY(), Face())   # boundary face in the z-normal direction
+    return _query_source(bc.source, bc.source_grid, X, (LX(), LY(), LZ()), _clock_time(clock))
```
ewquon:
```suggestion
    return _query_source(bc.source, bc.source_grid, X, (LX(), LY(), LZ()), clock_time(clock))
end

@inline function getbc(bc::Interpolated{2, S, LX, LY, LZ},
                       i::Integer, k::Integer, grid::AbstractGrid, clock=nothing, args...) where {S, LX, LY, LZ}
    j = _boundary_index(S, grid.Ny)
    X = node(i, j, k, grid, LX(), Face(), LZ())   # boundary face in the y-normal direction
    return _query_source(bc.source, bc.source_grid, X, (LX(), LY(), LZ()), clock_time(clock))
end

@inline function getbc(bc::Interpolated{3, S, LX, LY, LZ},
                       i::Integer, j::Integer, grid::AbstractGrid, clock=nothing, args...) where {S, LX, LY, LZ}
    k = _boundary_index(S, grid.Nz)
    X = node(i, j, k, grid, LX(), LY(), Face())   # boundary face in the z-normal direction
    return _query_source(bc.source, bc.source_grid, X, (LX(), LY(), LZ()), clock_time(clock))
```

src/EarthSystemModels/NestedSimulations/nested_model.jl
ewquon: The nested model clock behavior will be updated when I address https://github.com/NumericalEarth/NumericalEarth.jl/issues/330

> nested.grid is the grid of the child model

Yes

examples/era5_breeze.jl
ewquon: https://github.com/NumericalEarth/NumericalEarth.jl/pull/220/changes#diff-b2b4862ceab49cc5b710b9799b499e04a4d73899c5fb33b6b0555b2720ba76ddL77-R98

src/EarthSystemModels/NestedSimulations/child_simulation.jl
ewquon: Thanks for probing this @simone-silvestri. Turns out this is totally unused code — I'm tossing it

src/Atmospheres/prescribed_atmosphere.jl
ewquon: ERA5 does accumulate the surface quantities that contribute to the freshwater flux. This comment and code design is narrowly focused (only lateral, no surface BCs yet) and is confusing. I will fix

ewquon:
Netcdf output (if desired) pending https://github.com/CliMA/Oceananigans.jl/pull/5689

examples/era5_breeze.jl
```diff
@@ -0,0 +1,938 @@
...
+                             halo      = (5, 5, 5),
+                             topology  = (Bounded, Bounded, Bounded))
+
+# ETOPO 2022 surface elevation (≥ 0; ocean clamped to sea level) regridded onto
+# the LAM horizontal grid — ETOPO's 60″ (~1.85 km) relief is finer than the ~9 km
+# cells. `materialize_terrain!` fills the grid's terrain-following coordinate from it
+# in place; `CompressibleDynamics` then builds the slope metrics it needs directly
+# from the grid (no `terrain_metrics` argument required).
+
+elevation = regrid_topography(grid; dataset = ETOPO2022())
```
glwagner: @ewquon what do you think about incorporating the two lines into `atmosphere_simulation`:

```julia
topography = regrid_topography(grid; dataset=topography_dataset)
materialize_terrain!(grid, topography)
```

then `topography_dataset` is a kwarg to `atmosphere_simulation`.

glwagner: This can also handle the smoothing below

examples/era5_breeze.jl
```diff
@@ -0,0 +1,938 @@
...
+# Visualize the nesting before stepping the model: the ERA5 forcing region that
+# supplies the parent state (lateral BCs + Davies fringe) and the 9 km LAM —
+# Fan (2017)'s Domain 2, the `NestedSimulation` child — over ETOPO terrain with
+# Natural Earth state/country boundaries, centered on ARM SGP. Drawn here (not
+# with the profile plots below) so the domain geometry is written even if the run
+# is cut short.
+
+using CairoMakie
+using NaturalEarth
+import GeoInterface as GI
```
glwagner, line 241: @ewquon wondering how we can ship the code below. What do you think about adding an extension to `GeoInterface` which plots the terrain (eg culminating in the heatmap for `ax_map`? We can put it in this PR.

examples/era5_breeze.jl
```diff
@@ -0,0 +1,938 @@
...
+                                    halo      = (5, 5, 5),
+                                    topology  = (Bounded, Bounded, Bounded))
+
+# --- PrescribedAtmosphere on the parent grid ---
+# Times in seconds since the first snapshot. PrescribedAtmosphere allocates
+# default Center-located FTSs for velocities, tracers (T, q), pressure.
+# qᶜ, qⁱ aren't standard slots; we own those alongside.
+
+parent_times = [Float64(Dates.value(d - start_date)) / 1000 for d in dates]
+parent = PrescribedAtmosphere(parent_grid, parent_times; two_dimensional = false, freshwater_flux = nothing, thermodynamics_parameters = nothing)
```
glwagner, line 398: Why doesn't `ERA5PrescribedAtmosphere` work for this? Do we need a new constructor for it?

examples/era5_breeze.jl
```diff
@@ -0,0 +1,938 @@
...
+bcs = merge(bcs, (; ρe  = FieldBoundaryConditions(west   = bcs.ρe.west,
+                                                  east   = bcs.ρe.east,
+                                                  south  = bcs.ρe.south,
+                                                  north  = bcs.ρe.north,
+                                                  bottom = ValueBoundaryCondition(ρθ_surface_placeholder)),
+                   ρqᵉ = FieldBoundaryConditions(west   = bcs.ρqᵉ.west,
+                                                  east   = bcs.ρqᵉ.east,
+                                                  south  = bcs.ρqᵉ.south,
+                                                  north  = bcs.ρqᵉ.north,
+                                                  bottom = ValueBoundaryCondition(ρqᵉ_surface_placeholder))))
```
glwagner, line 581: This lines make me think we may need a new function / mode for `atmosphere_simulation`, like

```julia
parent = # set up parent simulation
child = atmosphere_simulation(grid; parent=..., kwargs...)
```

This could be accompanied by an internal function `nested_atmosphere_simulation` which correctly builds the boundary conditions and forcings from `parent`.

examples/era5_breeze.jl
```diff
@@ -0,0 +1,938 @@
...
+# non-const module globals produce dynamic-dispatch IR).
+lateral_mask = let λ_w = λ_west, λ_e = λ_east, φ_s = φ_south, φ_n = φ_north, fringe = fringe_deg
+    (λ, φ, z) -> begin
+        dW = λ - λ_w
+        dE = λ_e - λ
+        dS = φ - φ_s
+        dN = φ_n - φ
+        d  = min(dW, dE, dS, dN)
+        d >= fringe && return zero(λ)
+        return 0.5 * (1 + cos(π * d / fringe))
```
glwagner: `0.5` here is `Float64` (not sure if this is performance-critical but looks like it might be)

examples/era5_breeze.jl
glwagner: Why do we need to have `model` here? Woudl this work as well:

```julia
atmos = atmosphere_simulation(grid; kw...)
atmos_model = atmos.model
```

is there a reason we need to throw away the `Simulation`?

examples/era5_breeze.jl
```diff
@@ -0,0 +1,938 @@
...
+θ_arr  = Array(interior(θ_lam))
+qᵗ_arr = Array(interior(qᵗ))
+
+# Post-run LAM state. Specific quantities (θ, qᵗ) are derived from the
+# prognostic ρθ, ρqᵉ divided by ρ.
+ρ_final_arr  = Array(interior(model.dynamics.density))
+u_final_arr  = Array(interior(model.velocities.u))
+v_final_arr  = Array(interior(model.velocities.v))
+ρθ_final     = Array(interior(model.formulation.potential_temperature_density))
+ρqᵉ_final    = Array(interior(model.moisture_density))
```
glwagner: All of these should use field plotting directly instead of plotting arrays, I think

glwagner (commented):
some comments, review in progress

kaiyuan-cheng:
@ewquon @glwagner While the last commit avoids the spurious density at bottom levels, it seems to deviate from the ERA5 data. I think we should still use ln(p) interpolation.

glwagner:
> @ewquon @glwagner While the last commit avoids the spurious density at bottom levels, it seems to deviate from the ERA5 data. I think we should still use ln(p) interpolation.

I took a look at https://github.com/NumericalEarth/NumericalEarth.jl/pull/233/commits/2949c861d76252b34f09c9a16d842b46ba3ca7fd but I wasn't sure exactly which lines you were referring to, maybe comment on the commit directly? The code is a bit messy in there as well, which we will probably need to clean up soon. It might be good to have a benchmark of some kind for that refactoring process.

examples/breeze_over_four_oceans.jl
```diff
@@ -51,10 +51,10 @@ Tᵒᶜ = 290 # K
...
 coriolis = FPlane(latitude=33)
 
-prescribed_ocean_atmos = atmosphere_simulation(grid; potential_temperature=θᵃᵗ, coriolis)
-slab_ocean_atmos       = atmosphere_simulation(grid; potential_temperature=θᵃᵗ, coriolis)
-full_ocean_atmos       = atmosphere_simulation(grid; potential_temperature=θᵃᵗ, coriolis)
-nh_ocean_atmos         = atmosphere_simulation(grid; potential_temperature=θᵃᵗ, coriolis)
+prescribed_ocean_atmos = atmosphere_simulation(grid; potential_temperature=θᵃᵗ, coriolis).model
+slab_ocean_atmos       = atmosphere_simulation(grid; potential_temperature=θᵃᵗ, coriolis).model
+full_ocean_atmos       = atmosphere_simulation(grid; potential_temperature=θᵃᵗ, coriolis).model
+nh_ocean_atmos         = atmosphere_simulation(grid; potential_temperature=θᵃᵗ, coriolis).model
```
glwagner: How does this work? `EarthSystemModel` requires `Simulation`. I don't think we want to change that in this PR.

docs/make.jl
```diff
@@ -37,16 +37,17 @@ mkpath(OUTPUT_DIR)
...
+    #Example("Coupled energy and freshwater conservation", "coupled_conservation", true),
+    #Example("One-degree ocean--sea ice simulation", "one_degree_simulation", false),
+    #Example("Near-global ocean simulation", "near_global_ocean_simulation", false),
+    #Example("Global climate simulation", "global_climate_simulation", false),
+    #Example("Veros ocean simulation", "veros_ocean_forced_simulation", false),
+    #Example("Breeze over four oceans", "breeze_over_four_oceans", false),
+    #Example("ERA5 and GloFAS reanalysis data", "exploring_era5_reanalysis_data", true),
+    #Example("ERA5-forced slab land", "era5_forced_slab_land", true),
+    #Example("Breeze over slab land", "breeze_over_slab_land", true),
+    Example("ERA5 downscaling with Breeze and NestedSimulation", "era5_breeze", false),
```
glwagner:
```suggestion
    Example("Single-column ocean simulation", "single_column_os_papa_simulation", true),
    Example("Coupled energy and freshwater conservation", "coupled_conservation", true),
    Example("One-degree ocean--sea ice simulation", "one_degree_simulation", false),
    Example("Near-global ocean simulation", "near_global_ocean_simulation", false),
    Example("Global climate simulation", "global_climate_simulation", false),
    Example("Veros ocean simulation", "veros_ocean_forced_simulation", false),
    Example("Breeze over four oceans", "breeze_over_four_oceans", false),
    Example("ERA5 and GloFAS reanalysis data", "exploring_era5_reanalysis_data", true),
    Example("ERA5-forced slab land", "era5_forced_slab_land", true),
    Example("Breeze over slab land", "breeze_over_slab_land", true),
    # Example("ERA5 downscaling with Breeze and NestedSimulation", "era5_breeze", false),
```

examples/breeze_over_four_oceans.jl
glwagner:
```suggestion
prescribed_ocean_atmos = atmosphere_simulation(grid; potential_temperature=θᵃᵗ, coriolis)
slab_ocean_atmos       = atmosphere_simulation(grid; potential_temperature=θᵃᵗ, coriolis)
full_ocean_atmos       = atmosphere_simulation(grid; potential_temperature=θᵃᵗ, coriolis)
nh_ocean_atmos         = atmosphere_simulation(grid; potential_temperature=θᵃᵗ, coriolis)
```

glwagner:
both tests fail because of ECCO downloading error... so I think this PR is good to merge if you are ready @ewquon 

https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/28185821542/job/83487532646?pr=233#step:12:3961

glwagner (approved):

Issue #234, opened 2026-05-13
GPU on Docs built doesn't work


navidcy:
On main I see the error:

```Julia
┌ Error: Your NVIDIA TITAN V GPU (compute capability 7.0) is not supported on CUDA 13+.
│ Please use a device with at least capability 7.5, or downgrade your NVIDIA driver to below v580.
└ @ CUDACore /storage5/github-action-runners/julia-depot/packages/CUDACore/eDgUR/lib/cudadrv/state.jl:225
```

https://github.com/NumericalEarth/NumericalEarth.jl/pull/225 didn't seem to have resolve this

Any ideas how to resurrect the docs?

navidcy:
But the same fix as in #225 seems to have done the job for the Oceananigans docs: https://github.com/CliMA/Oceananigans.jl/pull/5489

One difference is that those are build via buildkite while here it's via GitHub Actions.

cc @simone-silvestri @glwagner @giordano

giordano:
It really depends on the version of the Nvidia driver and the CUDA toolkit used.  The driver is really the hardest component to deal with, hopefully that hasn't been touched?  The CUDA toolkit can be handled more easily.  What are the versions used here?

navidcy:
Shall I have `docs/make.jl` print out a `CUDA.versioninfo()` statement? Would that give us the information you are asking for @giordano?

giordano:
Yes, that'd help, thanks (don't need to actually build the docs after that, just run that command)

PR #237, opened 2026-05-14
Built docs on GPU

navidcy:
Attempt to deal with #234


navidcy:
@giordano see the `CUDA.versioninfo()` output at https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/25834595689/job/75906773612?pr=237#step:7:21

```Julia
CUDA toolchain: 
- runtime 12.9, artifact installation
- driver 535.247.1 for 12.2
- compiler 12.9
CUDA libraries: 
- cuBLAS: 12.9.1
- cuSPARSE: 12.5.10
- cuSOLVER: 11.7.5
- cuFFT: 11.4.1
- cuRAND: 10.3.10
- CUPTI: 2025.2.1 (API 12.9.1)
- NVML: 12.0.0+535.247.1
Julia packages: 
- CUDACore: 6.0.0
- GPUArrays: 11.5.3
- GPUCompiler: 1.10.0
- KernelAbstractions: 0.9.41
- CUDA_Driver_jll: 13.2.1+2
- CUDA_Compiler_jll: 0.4.3+0
- CUDA_Runtime_jll: 0.21.0+1
Toolchain:
- Julia: 1.12.5
- LLVM: 18.1.7
Environment:
- JULIA_CUDA_USE_COMPAT: false
1 device:
  0: NVIDIA TITAN V (sm_70, 11.770 GiB / 12.000 GiB available)
CUDA.versioninfo() = nothing
```

giordano:
That wouldn't explain the error
```
┌ Error: Your NVIDIA TITAN V GPU (compute capability 7.0) is not supported on CUDA 13+.
│ Please use a device with at least capability 7.5, or downgrade your NVIDIA driver to below v580.
└ @ CUDACore /storage5/github-action-runners/julia-depot/packages/CUDACore/eDgUR/lib/cudadrv/state.jl:225
```
the driver is way below v580, and the toolkit is 12.9, not 13+.

navidcy:
So how do we upgrade the toolkit? Could I do it?

navidcy:
Or the driver also?

giordano:
No, nothings should be changed!  My point is that the output isn't consistent with the error message.  Also, the build is actually progressing, no?

giordano:
Upgrading the driver would actually wreak everything, no one should touch it.

docs/make.jl
```diff
@@ -1,9 +1,13 @@
 using NumericalEarth
+using CairoMakie
+using CopernicusMarine
```
giordano: Not needed?
```suggestion
```

docs/make.jl
```diff
@@ -1,9 +1,13 @@
...
 using CUDA
 using Documenter
 using DocumenterCitations
 using Literate
 
+@show CUDA.versioninfo()
```
giordano: `CUDA.versioninfo()` prints by itself
```suggestion
CUDA.versioninfo()
```

giordano (approved):
I'd merge this PR and keep monitoring the situation.  When there's an error we can see what was `versioninfo()`.  As I said above, the current output doesn't explain the error message you saw before at all

navidcy:
The build progresses but if I include a literated example on GPU it'll fail!

So how do we make the GPU work? This is I think tartarus GPU, so perhaps it reached its lifetime?

giordano:
> The build progresses but if I include a literated example on GPU it'll fail!

But the error doesn't make any sense.

docs/make.jl
```diff
@@ -30,7 +32,7 @@ mkpath(OUTPUT_DIR)
 # on pushes to `main`/tags, or when the `build all examples` label is added to a PR.
 examples = [
     Example("Single-column ocean simulation", "single_column_os_papa_simulation", true),
-    Example("One-degree ocean--sea ice simulation", "one_degree_simulation", false),
+    Example("One-degree ocean--sea ice simulation", "one_degree_simulation", true),
```
giordano: I had [set the label](https://github.com/NumericalEarth/NumericalEarth.jl/pull/237#event-25507333445) to build the literate examples automatically 😉

navidcy: Gotcha

giordano:
This got past the point where there was the error in previous builds.

This may have been fixed by https://github.com/JuliaPackaging/Yggdrasil/pull/13712, if so that'd finally explain the error message: before that PR, CUDA toolkit 13+ was installed, causing the error message above (which would then make sense again). But now we're getting CUDA toolkit 12.9 being installed automatically, and that solves the compatibility issues.

navidcy:
So nothing we did here, just coincidentally it was fixed upstream?

Great!

giordano:
Yes. Not entirely coincidental though: that PR was spurred by https://github.com/JuliaGPU/CUDA.jl/pull/3132. However Oceananigans is still broken: https://github.com/CliMA/Oceananigans.jl/pull/5577. Can you please do the same debugging? I'm away from computer and sorely need to sleep at this point

giordano:
Uhm, actually the label didn't work because https://github.com/NumericalEarth/NumericalEarth.jl/blob/f4dcc011abba3b38be28391a8e30a5102e710c74/.github/workflows/docs.yml#L25 doesn't match the label name. Could you fix it? I really must go now.

navidcy:
Seems that the GPU examples didn't built? Despite the label?

simone-silvestri:
changed the label to match the if condition and restarted the build

giordano:
The fact that https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/25838710508/job/75919202625 has been running for over 4 hours and a half is a good sign, right?

simone-silvestri:
GPU goes through, the speedy-weather coupled simulation crashes though
https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/25838710508/job/75919202625

giordano:
Ok, that seems unrelated to the GPU issue though

navidcy:
```Julia
ERROR: LoadError: LoadError: ArgumentError: write failed, IOStream is not writeable
in expression starting at /storage5/github-action-runners/actions-runner-3/_work/NumericalEarth.jl/NumericalEarth.jl/docs/src/literated/global_climate_simulation.md:1
```

https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/25845095025/job/75938479243?pr=237#step:7:64

Is   tartarus out of space?

navidcy:
I get the same error locally so it's not tartarus fault; something changed in Oceananigans/SpeedyWeather/ClimaSeaIce?

navidcy:
I merged and let's see...
I Veros is breaking; see #263

PR #239, opened 2026-05-14
Cleans implicit imports + Add ExplicitImports quality assurance testsuite

### LINKED ISSUE #229: Cleaning imports
I'm pretty sure we can drop this import:

https://github.com/NumericalEarth/NumericalEarth.jl/blob/64dfb25104771c748fb4154dacf67174e4efd080/src/NumericalEarth.jl#L78

But from noticing this unused import, I was thinking how we should perhaps move away from using import from extending, similar to what we did in Oceananigans (an effort pushed by @giordano; thanks!)

navidcy:
Closes #229


navidcy:
@giordano I tried to follow Oceananigans practices. The tests pass here now. Could you have a look? There is a chance I misunderstood something.

I am a bit confused because I was thinking that the end goal would be that there is no `import ...` whatsoever.

navidcy:
https://github.com/NumericalEarth/NumericalEarth.jl/pull/239/commits/8497c2ae0dba8145ae3579d9425f6bb446018ac5 tried to have all extension of methods explicit.

giordano:
For what is worth, the way I go when I want to implement ExplicitImports is to run
```julia
using ExplicitImports, MyPackage
print_explicit_imports(MyPackage)
```
or
```julia
using ExplicitImports, MyPackage
print_explicit_imports(MyPackage.Submodule)
```
if MyPackage is very large, and working on a submodule-basis is more efficient (that's what I've been doing in Oceananigans). This is the [recommended approach in the docs](https://juliatesting.github.io/ExplicitImports.jl/stable/) and is quite nice because it shows you exactly what's needed to change, and how to change it. Then I use some judgement (for example it always suggests `using AnotherPackage: AnotherPackage`, even if not actually needed, I think as a precaution), but I have some good guidance.

navidcy:
I've done it... I believe this is ready to merge?
I haven't cleaned all extensions.

navidcy:
Close in favour of #246

PR #241, opened 2026-05-14
Per-column z ingest for ERA5 pressure-level data

### LINKED ISSUE #236: ERA5 pressure-level ingest: per-column geopotential and surface-pressure masking
## Motivation

`set!(field, era5_pressure_level_metadatum)` constructs the ERA5 source `Field` on a `LatitudeLongitudeGrid` whose z-axis is a single 1-D vector returned by `mean_geopotential_heights(metadata)` — the space-time average of Φ/g over the dataset's bbox. This single mean profile is applied uniformly to every (λ, φ) column, and `Oceananigans.interpolate!` then treats ERA5 pressure surfaces as constant-z. That assumption only holds when Φ(p) is spatially uniform.

Two regimes where it breaks:

1. **Orography.** ERA5 pressure-level output includes values below the local model surface (Himalayas, Andes, US High Plains, …). At these levels ERA5 has performed sub-surface extrapolation (see below); the values are non-physical and should be discarded.
2. **Synoptic systems.** Even over flat terrain, weather features displace Φ(p) surfaces horizontally. The 950 hPa surface differs in height by ~40 m between two columns whose sea-level pressures (SLP) differ by 5 hPa (`δz ≈ −H · δp_sfc / p_sfc`, with scale height H ≈ 8 km) — comparable to a limited-area model (LAM) Δz ≈ 50 m near the ground.

Concrete example: in a 768 × 768 km LAM centered on Lamont, OK in the US, `p_surface(x, y)` from ERA5 ranges 89.7–100.8 kPa — an 11 kPa span driven mostly by ~1000 m of terrain relief inside the bbox. The single mean ⟨Φ/g⟩ profile bears no clean relation to the actual column at any (λ, φ).

## Proposed approach

Per-column vertical interpolation in the `set!` pipeline:

1. Use the **instantaneous** Φ(λᵢ, φⱼ, p)/g for each ERA5 column as the z-mapping, not the spatial-time mean.
2. Mask levels where p > p_surface(λᵢ, φⱼ) — these are sub-surface ERA5 extrapolations, not real atmospheric state.
3. When interpolating a target column at (λₜ, φₜ, zₜ), bilinearly select the 4 surrounding ERA5 columns and interpolate linearly along each column's valid (z, var) pairs, then bilinearly combine.

This is the standard pre-processing pattern used by WRF (`real.exe`), MPAS-A, and similar regional models initializing from ECMWF pressure-level reanalysis.

## Related: `pressure_field` should mask sub-surface levels

`pressure_field(metadatum)` returns a Field with the dataset's `pressure_levels` populated uniformly at every (λ, φ) cell. There's no notion of "valid below the surface" — every cell gets a pressure value regardless of whether that level is below ground there. Downstream consumers (e.g., computing θ = T·(pₛₜ/p)^κ on every ERA5 cell, or building a 3-D pressure field for an IC) silently include sub-surface samples.

Same root cause as the main issue; the fix is to pair the field with a surface-pressure mask.

## ERA5 sub-surface fill behavior

ECMWF's pressure-level interpolation (137 hybrid model levels → 37 pressure levels) extrapolates sub-surface values as follows:

- **Wind components and specific humidity**: held constant at the lowest model level.
- **Temperature**: linearly interpolated from the lowest model level down to a surface temperature, then cubic-polynomial extrapolation below.
- **Geopotential**: continues hydrostatically using the extrapolated temperature.

ECMWF explicitly recommends masking: *"The data which is below the surface can be masked out using the surface pressure or the surface geopotential."*

References:
- ERA5 data documentation, ECMWF Confluence: https://confluence.ecmwf.int/display/CKB/ERA5%3A+data+documentation
- "Atmospheric variables below model terrain", ECMWF Forum: https://forum.ecmwf.int/t/atmospheric-variables-below-model-terrain/1225
- ECMWF Knowledge Base, computing pressure and geopotential on model levels: https://confluence.ecmwf.int/display/CKB/ERA5:+compute+pressure+and+geopotential+on+model+levels,+geopotential+height+and+geometric+height

glwagner:
## Summary

- **`PressureLevelVerticalDiscretization` + `PressureLevelGrid`** (new `src/Grids` submodule). Stores raw geopotential `Φ(λ,φ,p)` (m²/s²) and `gravitational_acceleration`; `rnode(i,j,k,grid)` reads `Φ[i,j,k] / g`. `_fractional_indices` is specialized so `Oceananigans.interpolate!` bisects each column's actual heights inside the kernel, instead of the 1-D `znodes(grid, loc)` vector that the default assumes.
- **ERA5 pressure-level ingest is per-column by default.** `z_interfaces(::ERA5PressureMetadata)` builds the discretization from the instantaneous Φ and clips at the local surface geopotential via `clip_subsurface!`. No new dataset keyword — old `z_mode` / `mean_geopotential_height` plumbing is gone.
- **`znodes` dispatch on `AbstractField`.** Returns the column-mean `Vector{Float64}` when the field is horizontally absent (`Flat` x/y topology or `Nothing` horizontal locations from `mean(...; dims=(1,2))`); otherwise a 3-D per-cell `Field`. Same logic applies to `FieldTimeSeries`, so `mean(znodes(T_series), dims=(1,2))` yields a `Field{Nothing, Nothing, Center}` as expected.
- **CDS request aggregation.** Single-variable and multi-variable ERA5 downloads now batch by `(year, month)` (using CDS's Cartesian `year`/`month`/`day`/`time` semantics) and further cap each batch at 5000 fields per request to avoid HTTP 403 cost-limit errors on pressure-level multi-var pulls.
- **Example: `examples/ERA5_hourly_data.jl`.** Adds a §4 that animates ERA5 specific humidity downscaled 4× from the 0.25° native grid onto a fixed-altitude target, illustrating the per-column-z interpolation end-to-end. Plotting routes through the Oceananigans Makie ext (Fields and AbstractOperations as plot args; `Observable{<:Field}` for animation). The §4 frame loop reuses the §3c `q_series` `FieldTimeSeries` via `interpolate!`, so no per-frame metadata lookup or grid reconstruction.
- **Example size tuned for CI.** Date range trimmed from 7 days (Dec 27–Jan 2) to 4 days entirely in December, `rico_region` from 1.5°×1.5° to 1°×1°. Cold-cache local repro: **53 min → 30 min**.

Closes #236.

### Why
`set!(field, era5_pressure_level_metadatum)` previously placed the source ERA5 field on a `LatitudeLongitudeGrid` whose z-axis was a single 1-D vector of `⟨Φ⟩/g` averaged over space and time. Over real terrain (sub-surface ERA5 fill leaks through that flat profile) and under synoptic Φ-surface displacement (~40 m for 5 hPa SLP gradients — comparable to LAM near-surface Δz), that profile is wrong by O(50–500 m) in cell placement. See #236 for details and the [design write-up](https://github.com/NumericalEarth/NumericalEarth.jl/issues/236#issuecomment-4452907548).

### Design highlights
- **No advection-side changes.** The discretization participates only in `rnode` and `_fractional_indices`; advection, halo, and model-side kernels never touch it because pressure-level grids aren't used as model grids.
- **Sub-surface clip at build time.** `clip_subsurface!` rewrites `Φ[i,j,k] ← max(Φ[i,j,k], Φ_sfc[i,j])` in place and refills halos. Columns are guaranteed monotone in k, so `column_fractional_z_index` needs no branching and the kernel stays clean.
- **Surface-Φ paired by dispatch.** `_matching_single_level_dataset(::ERA5HourlyPressureLevels) = ERA5HourlySingleLevel()` and the monthly counterpart, so the surface-clip source matches the cadence of the data being clipped.
- **`znodes(::Field)` / `znodes(::FieldTimeSeries)` route via `_znodes_for_plg`.** Two concrete-type methods (one on `Field{…<:PressureLevelGrid}`, one on `FieldTimeSeries{…<:PressureLevelGrid}`) — each more specific than Oceananigans' own `znodes` defaults on their respective wrappers, so dispatch is unambiguous.
- **CDS batching helper `_batch_datetimes_for_cds`.** Groups by `(year, month)` first (preserving CDS's Cartesian product), then chunks each month into ≤ `5000 / (num_vars × num_levels)` datetimes. Recovers most of the daily→monthly speedup for small payloads (§2/§3a: 2.5–3.3×) while staying safely under CDS's pressure-level cost limit.

### What this defers
- **FTS/`TimeSeriesInterpolation`-backed `Φ`** for forecast-time use. The discretization is already structurally compatible (its `geopotential` field accepts a TSI), but `ERA5PrescribedAtmosphere` and the `geopotential` field on `PrescribedAtmosphere` aren't here — separate PR.
- **Per-donor-column bisection in `_interpolate`.** `column_fractional_z_index` bisects a single representative column — the cell at the integer indices `clamp(unsafe_trunc(Int, ii), 1, Nx)`, `clamp(unsafe_trunc(Int, jj), 1, Ny)`. The bilinear blend in `Oceananigans.interpolate!` then weights values at the neighbouring `(i+1, j)`, `(i, j+1)`, `(i+1, j+1)` columns with the same `(k_low, k_high)`, even though those columns have slightly different z values at the same k. Near sharp orographic gradients this biases the result. Tractable as a kernel-level follow-up.
- **`FieldTimeSeries` support in `OceananigansMakieExt`** — `heatmap!(ax, fts)` would let the §3 Hovmöller drop its manual `(Nt × Nz)` matrix assembly. Tracked upstream as [CliMA/Oceananigans.jl#5599](https://github.com/CliMA/Oceananigans.jl/issues/5599).

## Test plan
- [x] CPU unit tests (`test/test_pressure_level_grid.jl`, 58 tests): `PressureLevelVerticalDiscretization` constructor, `generate_coordinate` dim/axis guards, `clip_subsurface!` behavior on a `Field`-backed Φ, grid-level `znodes`/`rnodes` agreement, `znodes(::Field)` per-cell vs column-mean dispatch on all four cases (horizontally resolved, `Flat`-`Flat` topology, `Nothing`-`Nothing` location, `FieldTimeSeries` parity).
- [x] CDS batching unit tests (`test/test_cds_downloading.jl`): `_group_by_calendar_month`, `_batch_datetimes_for_cds` (single-level monthly fits in one batch; pressure-level many-vars splits within a month; `_max_dts_per_cds_request` arithmetic; chronological ordering under shuffled input).
- [x] ExplicitImports QA after the `Adapt` import cleanup in `src/Grids/pressure_level_vertical_discretization.jl`.
- [x] End-to-end ERA5 example builds in CI Documenter (was hanging 7h+ before this PR's aggregations; now finishes in ~46 min).
- [x] Hurricane Helene demo (out-of-tree): 37-frame animation of ERA5 wind speed at z = 1.5 km over a 15°×18° SE-US bbox, downscaled 4×. Storm structure (eye, spiral, rapid translation) clearly resolved through landfall and inland decay.
- [ ] Run `examples/era5_breeze.jl` and confirm field ranges match the previous two-stage workaround to within rectification bias — out of scope for this PR.


glwagner:
here's a movie that can be created after this PR, atmos state interpolated to a constant z-level

https://github.com/user-attachments/assets/b3e330f7-3a13-4db4-a53c-2a3ef874af14

examples/ERA5_hourly_data.jl
```diff
@@ -290,7 +290,12 @@ qʳ_col_meta   = Metadata(:specific_rain_water_content;         dataset = ds_pl,
 qᶜ_col_series = FieldTimeSeries(qᶜ_col_meta)
 qʳ_col_series = FieldTimeSeries(qʳ_col_meta)
 
-z_col  = znodes(qᶜ_col_series[1])
+# Pressure-level grids carry a per-column vertical coordinate, so `znodes`
+# returns a lazy Oceananigans operation. Materialize it before passing heights
+# to plotting functions.
+pressure_level_heights(field) = Array(interior(compute!(Field(znodes(field)))))
```
navidcy: What's the Array do here after interior?
To ensure it's not a view?

glwagner: no, I think we should use view here

ewquon:
I don't think this description makes sense and is stale:
> Single sentinel face vector. cᵃᵃᶠ = cᵃᵃᶜ = 0:Nz

ext/NumericalEarthCDSAPIExt.jl
```diff
@@ -168,44 +168,95 @@ function download_dataset(meta::ERA5Metadatum; skip_existing=true)
...
-##### Multi-date download — batches by calendar day
+##### Multi-date download — batches by calendar month, capped by CDS cost
+#####
+##### CDS interprets `year`/`month`/`day`/`time` as a Cartesian product, so a
+##### single request can cover many days × hours per call as long as `year`
+##### and `month` stay singletons. The remaining constraint is the per-request
+##### cost limit: roughly `num_vars × num_pressure_levels × num_datetimes`
+##### must stay under ~7500 for pressure-level data (CDS returns HTTP 403
+##### "cost limits exceeded" otherwise). We pick a conservative cap and split
+##### each month into smaller contiguous chunks when needed.
```
ewquon, line 179: Looks like you've further aggregated the data requests, nice! Should update the PR description?

src/DataWrangling/ERA5/ERA5_pressure_levels.jl
```diff
@@ -2,37 +2,52 @@ abstract type ERA5PressureLevelsDataset <: ERA5Dataset end
...
 dataset_name(::ERA5MonthlyPressureLevels) = "ERA5MonthlyPressureLevels"
 
 #####
 ##### ERA5 pressure-level data availability
 #####
 
 # ERA5 reanalysis data available from 1940 to present (we use a practical range here)
-all_dates(::ERA5HourlyPressureLevels,  var) = range(DateTime("1940-01-01"), stop=DateTime("2024-12-31"), step=Hour(1))
+all_dates(::ERA5HourlyPressureLevels, var) = range(DateTime("1940-01-01"), stop=DateTime("2024-12-31"), step=Hour(1))
 all_dates(::ERA5MonthlyPressureLevels, var) = range(DateTime("1940-01-01"), stop=DateTime("2024-12-01"), step=Month(1))
```
ewquon:
```suggestion
all_dates(::ERA5HourlyPressureLevels, var) = range(DateTime("1940-01-01"), stop=DateTime("2025-12-31"), step=Hour(1))
all_dates(::ERA5MonthlyPressureLevels, var) = range(DateTime("1940-01-01"), stop=DateTime("2025-12-01"), step=Month(1))
```

src/DataWrangling/ERA5/ERA5_pressure_levels.jl
```diff
@@ -239,6 +239,53 @@ function stagger(zc::AbstractVector)
...
+to break the recursive `Field(ϕ_meta) → native_grid → z_interfaces` chain. The
+static z values are arbitrary placeholders; only their shape matters.
+"""
+function per_column_geopotential_discretization(metadata::ERA5PressureMetadata)
+    plain_z = standard_atmosphere_z_interfaces(metadata.dataset.pressure_levels)
+    plain_ds = _with_z(metadata.dataset, plain_z)
+    sl_ds = _matching_single_level_dataset(metadata.dataset)
+
+    ϕ_meta = Metadata(:geopotential; dataset=plain_ds,
+                      dates=metadata.dates, region=metadata.region, dir=metadata.dir)
```
ewquon:
```suggestion
    static_z = standard_atmosphere_z_interfaces(metadata.dataset.pressure_levels)
    static_ds = _with_z(metadata.dataset, static_z)
    sl_ds = _matching_single_level_dataset(metadata.dataset)

    ϕ_meta = Metadata(:geopotential; dataset=static_ds,
                      dates=metadata.dates, region=metadata.region, dir=metadata.dir)
```
`static` naming matches the docstring, "The static z values are arbitrary placeholders; only their shape matters."

src/Grids/pressure_level_vertical_discretization.jl
```diff
@@ -0,0 +1,270 @@
...
+    dim == 3              || throw(ArgumentError("PressureLevelVerticalDiscretization requires dim=3"))
+    coordinate_name == :z || throw(ArgumentError("PressureLevelVerticalDiscretization requires coordinate_name=:z"))
+
+    g = coord.gravitational_acceleration
+    Φi = _geopotential_data_for_extrema(coord.geopotential)
+    z_lo, z_hi = FT.(extrema(Φi) ./ g)
+    Lz = z_hi - z_lo
+
+    moved = PressureLevelVerticalDiscretization(g, on_architecture(arch, coord.geopotential))
+    return Lz, moved
```
ewquon:
```suggestion
    return Lz, PressureLevelVerticalDiscretization(g, on_architecture(arch, coord.geopotential))
```
Nit: At least to me, `moved` seems to suggest that there was a change to the grid geometry

glwagner: i called it `arch_discretization`

ext/NumericalEarthCDSAPIExt.jl
glwagner: done

glwagner:
> I don't think this description makes sense and is stale:
> 
> > Single sentinel face vector. cᵃᵃᶠ = cᵃᵃᶜ = 0:Nz

removed!

src/Grids/pressure_level_vertical_discretization.jl
```diff
@@ -0,0 +1,270 @@
...
+                                        Adapt.adapt(to, c.geopotential))
+
+on_architecture(arch, c::PressureLevelVerticalDiscretization) =
+    PressureLevelVerticalDiscretization(c.gravitational_acceleration,
+                                        on_architecture(arch, c.geopotential))
+
+function Base.show(io::IO, z::PressureLevelVerticalDiscretization)
+    print(io, "PressureLevelVerticalDiscretization with $(size(z.geopotential, 3)) levels, ",
+              "g = ", z.gravitational_acceleration, " m/s²")
+end
```
ewquon:
```suggestion
Adapt.adapt_structure(to, coord::PressureLevelVerticalDiscretization) =
    PressureLevelVerticalDiscretization(coord.gravitational_acceleration,
                                        Adapt.adapt(to, coord.geopotential))

on_architecture(arch, coord::PressureLevelVerticalDiscretization) =
    PressureLevelVerticalDiscretization(coord.gravitational_acceleration,
                                        on_architecture(arch, coord.geopotential))

function Base.show(io::IO, coord::PressureLevelVerticalDiscretization)
    print(io, "PressureLevelVerticalDiscretization with $(size(coord.geopotential, 3)) levels, ",
              "g = ", coord.gravitational_acceleration, " m/s²")
end
```
Minor nits, for consistency with `PressureLevelVerticalDiscretization` use above

glwagner: or we can change the above to `z`? This is the object passed to the kwarg `z` of a grid (eg `RectilinearGrid(z=PressureLevelVerticalDiscretization())`. I don't feel strongly though.

src/NumericalEarth.jl
```diff
@@ -98,7 +98,9 @@ export
     net_ocean_freshwater_flux, sea_ice_ocean_freshwater_flux, atmosphere_ocean_freshwater_flux,
     meridional_heat_transport,
     location,
-    native_grid
+    native_grid,
+    PressureLevelVerticalDiscretization,
```
ewquon: Does this (and `PressureLevelGrid`) need to be exported?

glwagner: I'd say it shouldn't be exported (since it typically is built under the hood), so I will remove this

src/Grids/pressure_level_vertical_discretization.jl
```diff
@@ -0,0 +1,270 @@
...
+
+# TSI path is CPU-only for now (no FTS-backed discretization is exercised in
+# the current pipeline). When that lands, swap the loop for a 4-D launch.
+function clip_subsurface!(geopotential::TimeSeriesInterpolation, surface_geopotential)
+    fts_parent = parent(geopotential.time_series)
+    Φs = interior(surface_geopotential)
+    Nx, Ny, Nz, Nt = size(fts_parent)
+    @inbounds for t in 1:Nt, k in 1:Nz, j in 1:Ny, i in 1:Nx
+        fts_parent[i, j, k, t] = max(fts_parent[i, j, k, t], Φs[i, j, 1])
```
ewquon:
```suggestion
    fts = geopotential.time_series
    for t in 1:length(fts.times)
        clip_subsurface!(fts[t], surface_geopotential)
```
Original gives 
> ERROR: BoundsError: attempt to access 4×4×1 view ... at index [5, 1, 1]

Reproducer:
```julia
using Oceananigans
using Oceananigans.BoundaryConditions: fill_halo_regions!
using Oceananigans.OutputReaders: TimeSeriesInterpolation
using NumericalEarth.Grids: clip_subsurface!

g = 9.81
Φ_grid = LatitudeLongitudeGrid(CPU(); size=(4, 4, 4),
                                longitude=(0, 1), latitude=(0, 1), z=(0, 1));
Φ_fts = FieldTimeSeries{Center, Center, Center}(Φ_grid, [0.0, 1.0]);
for t in 1:2, k in 1:4, j in 1:4, i in 1:4
    Φ_fts[t][i, j, k] = 1000.0 * k * g
end
fill_halo_regions!(Φ_fts[1]); fill_halo_regions!(Φ_fts[2])

Φ_sfc_grid = LatitudeLongitudeGrid(CPU(); size=(4, 4, 1),
                                    longitude=(0, 1), latitude=(0, 1), z=(0, 1));
Φ_sfc = CenterField(Φ_sfc_grid);
interior(Φ_sfc) .= 2500.0 * g

tsi = TimeSeriesInterpolation(Φ_fts, Φ_fts.grid; clock = Clock(time = 0.0));
clip_subsurface!(tsi, Φ_sfc)
```

src/Grids/pressure_level_vertical_discretization.jl
```diff
@@ -0,0 +1,270 @@
...
+
+Adapt.adapt_structure(to, c::PressureLevelVerticalDiscretization) =
+    PressureLevelVerticalDiscretization(c.gravitational_acceleration,
+                                        Adapt.adapt(to, c.geopotential))
+
+on_architecture(arch, c::PressureLevelVerticalDiscretization) =
+    PressureLevelVerticalDiscretization(c.gravitational_acceleration,
+                                        on_architecture(arch, c.geopotential))
+
+function Base.show(io::IO, z::PressureLevelVerticalDiscretization)
```
ewquon, line 106: Definition of `function Base.show(io::IO, ::MIME"text/plain", grid::PressureLevelGrid)` is also needed, otherwise when creating a `PressureLevelGrid` in the REPL, we get 
> ERROR: 1-element ExceptionStack:
FieldError: type PressureLevelVerticalDiscretization has no field `cᵃᵃᶠ`, available fields: `gravitational_acceleration`, `geopotential`

src/Grids/pressure_level_vertical_discretization.jl
```diff
@@ -0,0 +1,270 @@
...
+                                      grid::PressureLevelGrid, ℓx, ℓy, ℓz)
+    ii = fractional_x_index(x, (ℓx, ℓy, ℓz), grid)
+    jj = fractional_y_index(y, (ℓx, ℓy, ℓz), grid)
+    kk = column_fractional_z_index(z, ii, jj, (ℓx, ℓy, ℓz), grid)
+    return FractionalIndices(ii, jj, kk)
+end
+
+# Column-region source (Flat-Flat-Bounded): there's only one (i,j)=(1,1), so
+# bisect that single column directly. Mirrors the 3-D form's column logic.
+@inline function _fractional_indices((z,)::NTuple{1, Any},
```
ewquon, line 232: Do we want to also support slicing with `NTuple{2, Any}` for completeness?

src/Grids/pressure_level_vertical_discretization.jl
```diff
@@ -0,0 +1,270 @@
...
+                                      grid::PressureLevelGrid, ::Nothing, ::Nothing, ℓz)
+    kk = column_fractional_z_index(z, 1, 1, (nothing, nothing, ℓz), grid)
+    return FractionalIndices(nothing, nothing, kk)
+end
+
+@inline function column_fractional_z_index(z, ii, jj, locs, grid)
```
ewquon: `locs` unused

src/Grids/pressure_level_vertical_discretization.jl
```diff
@@ -0,0 +1,270 @@
...
+    Lz = z_hi - z_lo
+
+    moved = PressureLevelVerticalDiscretization(g, on_architecture(arch, coord.geopotential))
+    return Lz, moved
+end
+
+_geopotential_data_for_extrema(Φ::Field) = interior(Φ)
+# NOTE: For a TSI this reads every time slice. Fine while the FTS path isn't
+# exercised; switch to a per-time-slice extent if we ever advance the clock here.
+_geopotential_data_for_extrema(Φ::TimeSeriesInterpolation) = parent(Φ.time_series)
```
ewquon: This can give incorrect `znodes`, reproducer:
```julia
using Oceananigans
using Oceananigans.OutputReaders: TimeSeriesInterpolation
using NumericalEarth

g = 9.81
Φ_grid = LatitudeLongitudeGrid(CPU(); size=(4, 4, 4),
                                longitude=(0, 1), latitude=(0, 1), z=(0, 1));
Φ_fts = FieldTimeSeries{Center, Center, Center}(Φ_grid, [0.0, 1.0]);
for i in 1:4, j in 1:4, k in 1:4
    Φ_fts[1][i, j, k] = 1000.0 * k * g     # interior heights {1, 2, 3, 4} km
    Φ_fts[2][i, j, k] = 5000.0 * k * g     # interior heights {5, 10, 15, 20} km
end

tsi  = TimeSeriesInterpolation(Φ_fts, Φ_fts.grid; clock = Clock(time = 0.0));
plvd = PressureLevelVerticalDiscretization(tsi; gravitational_acceleration = g);
grid = LatitudeLongitudeGrid(CPU(); size=(4, 4, 4),
                              longitude=(0, 1), latitude=(0, 1), z=plvd);

@show znodes(grid, Center())
# [0.0, 0.0, 0.0, 480.0]  — dominated by halo zeros, not a time-mean
```
whereas `plvd = PressureLevelVerticalDiscretization(Φ_fts[n]; ...)` gives the expected result

src/Grids/pressure_level_vertical_discretization.jl
ewquon: I don't feel strongly about which one either, only that it'd be nice not to have 3 different variable names.

glwagner: ok, switched to `z`

src/Grids/pressure_level_vertical_discretization.jl
glwagner: hmm true. Including `loc` makes the function signature uniform with `fractional_z_index` but perhaps that isn't important here.

src/Grids/pressure_level_vertical_discretization.jl
glwagner: Let me figure out if this already works

glwagner: Not needed at the moment. Looking at the ERA5 path, the only `Region` types that produce a metadata-derived grid here are `BoundingBox` (full 3-D source, hits the `NTuple{3, Any}` form) and `Column` (`Flat-Flat-Bounded` source, hits the `NTuple{1, Any}` form). There's no built-in transect/slice region that would land in `_fractional_indices((y, z), grid, ::Nothing, ℓy, ℓz)` or the zonal analog, so the 2-tuple form has no caller today.

If a `LineRegion` / `Transect` is added later it's a one-line dispatch (`kk = column_fractional_z_index(z, ii, 1, grid)` with `ii` from `fractional_y_index(...)` or similar). I'd add it then so we don't carry dead code.

glwagner: ^^ that was claude

glwagner: hmm this may be right. in principle we can build a grid with one Flat dimension, but we don't have a way to use that right now with ERA5 data I don't think

src/Grids/pressure_level_vertical_discretization.jl
glwagner: Fixed in 1108f2e1. Override is on the concrete `LatitudeLongitudeGrid{…<:PressureLevelVerticalDiscretization}` combination — strictly more specific than either the LLG default or the `PressureLevelGrid` alias, so no ambiguity. The override prints the horizontal axes / size / halo / Lz and delegates the z-line to PLVD's own `show`. Reproduction at the REPL:

```julia
julia> grid = LatitudeLongitudeGrid(CPU(); size=(2, 2, 5), longitude=(0,1), latitude=(0,1), z=plvd)
2×2×5 LatitudeLongitudeGrid{Float64, Bounded, Bounded, Bounded} on CPU with 2×2×3 halo
├── topology: (Bounded, Bounded, Bounded)
├── size: (Nx=2, Ny=2, Nz=5)
├── halo: (Hx=2, Hy=2, Hz=3)
├── Lz (column-mean span): 4000.0 m
└── PressureLevelVerticalDiscretization with 5 levels, g = 9.81 m/s²
```

Regression test in `test_pressure_level_grid.jl` asserts both `sprint(show, grid)` and `sprint(show, MIME"text/plain"(), grid)` complete without `FieldError`.

ewquon (approved):

glwagner:
merging this because the CI errors are unrelated.

PR #242, opened 2026-05-15
Fix veros compatibility

simone-silvestri:
This should fix veros' tests @navidcy


simone-silvestri:
actually, probably not needed since https://github.com/team-ocean/veros/commit/77baf7b0f79848c62abe09b00ee2c4cbfc28ef6a passed

PR #243, opened 2026-05-15
Reinitialize download progress when dowloading a new file

simone-silvestri:
At the moment, the download_progress uses global variables 
```
next_fraction = Ref(0.0)
download_start_time = Ref(time_ns())
```
that persist during downloads in the same session such that we loose the download_progress information after downloading the first file. This PR wraps them in a `DownloadProgress` struct that gets correctly reinitialized at each download

### DIFF
--- a/src/DataWrangling/DataWrangling.jl
+++ b/src/DataWrangling/DataWrangling.jl
@@ -36,10 +36,12 @@ import Oceananigans.Fields: set!
 ##### Downloading utilities
 #####
 
-Base.@kwdef struct DownloadProgress
-    next_fraction :: Base.RefValue{Float64} = Ref(0.0)
-    download_start_time :: Base.RefValue{UInt64} = Ref(time_ns())
+mutable struct DownloadProgress
+    next_fraction :: Float64
+    download_start_time :: UInt64
 end
+
+DownloadProgress() = DownloadProgress(0.0, time_ns())
 
 """
     DownloadProgress(total, now; filename="")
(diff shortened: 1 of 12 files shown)


src/DataWrangling/DataWrangling.jl
```diff
@@ -36,36 +36,38 @@ import Oceananigans.Fields: set!
 ##### Downloading utilities
 #####
 
-next_fraction = Ref(0.0)
-download_start_time = Ref(time_ns())
+Base.@kwdef struct DownloadProgress
+    next_fraction :: Base.RefValue{Float64} = Ref(0.0)
+    download_start_time :: Base.RefValue{UInt64} = Ref(time_ns())
+end
```
giordano: Why not making this a mutable struct?
```suggestion
Base.@kwdef mutable struct DownloadProgress
    next_fraction :: Float64 = 0.0
    download_start_time :: UInt64 = time_ns()
end
```

simone-silvestri: makes sense!

giordano (approved):

src/DataWrangling/DataWrangling.jl
```diff
@@ -36,36 +36,40 @@ import Oceananigans.Fields: set!
 ##### Downloading utilities
 #####
 
-next_fraction = Ref(0.0)
-download_start_time = Ref(time_ns())
+mutable struct DownloadProgress
```
simone-silvestri:
```suggestion
mutable struct DownloadProgress <: Function
```

giordano:
All green, ready to go?

PR #244, opened 2026-05-15
Wire in freshwater fluxes

simone-silvestri:
a freshwater flux should add / remove volume so it is added as a forcing in the \eta equation. This is possible to have only when we have a free-surface following vertical discretization.

The salt fluxes change their interpretation in this case, they are no more virtual fluxes but corrections that account for the incoming / outgoing salt in the added / removed volume. Since rain and runoff have $S = 0$ the correction stays $F * S_o$ (Should be $F * (S_o - S_i)$. 

Depends on https://github.com/CliMA/Oceananigans.jl/pull/5592

Update: 

It required a bit more fiddling to get to a real conservative code: when we have freshwater fluxes, since the volume changes, we need to add a _virtual_ flux which balances the volume change and keeps the tracer content constant. 
The _real_ tracer flux associated with a freshwater flux (both for salinity and for temperature becomes)
```julia
#     Σᵢ (cᴺ − cᵢ) Jʷᵢ  = cᴺ · (Σᵢ Jʷᵢ)  −  Σᵢ cᵢ Jʷᵢ
#                         └─ carrying ─┘  └─ content ─┘
```
where `cᴺ` is the tracer value at the surface of the ocean (the "virtual" part) and `cᵢ` is the tracer value of the incoming freshwater (in case of salinity it is zero most of the times, only for sea ice there might be a value).
To be machine-precision conserving `cᴺ` needs to be evaluated at the stepping time, thefore we split the "carrying" and "content" part. The former is easily computed via a `DiscreteBoundaryFunction` which has access to the live `cᴺ`, while the second is an explicit flux which is computed at flux assembly. 

For salinity, all sources except ice are assumed to have S = 0, while for temperature we use a T=0 for everything except for P-E where we assume that both evaporation and rain are at the SST. A further improvement would be to introduce a `T_rain` and `S_rain` but it might be for another PR, this design allows to add new specific terms to the "content" part of the flux pretty easily (see how rain is treated)


src/EarthSystemModels/InterfaceComputations/component_interfaces.jl
```diff
@@ -130,30 +130,33 @@ struct SeaIceOceanFluxes{C, FX, FY}
     interface_heat :: C
     frazil_heat    :: C
     salt           :: C
+    freshwater     :: C
```
glwagner, line 133: does this include precipitation or just runoff?

src/Oceans/ocean_simulation.jl
```diff
@@ -373,6 +373,15 @@ function hydrostatic_ocean_simulation(grid;
...
     top_ocean_heat_flux          = Jᵀ = Field{Center, Center, Nothing}(grid)
     top_salt_flux                = Jˢ = Field{Center, Center, Nothing}(grid)
+    top_freshwater_volume_flux   = Fη = Field{Center, Center, Nothing}(grid)
+    
+    if grid isa MutableGridOfSomeKind
+        if :η ∈ keys(forcing)
+            forcing = merge(forcing, (η = (Fη, forcing.η),))
+        else
+            forcing = merge(forcing, (η = Fη,))
+        end
```
glwagner:
```suggestion
        if :η ∈ keys(forcing)
            Fη = (Fη, forcing.η)
        end
        forcing = merge(forcing, (; η = Fη))
```

src/EarthSystemModels/InterfaceComputations/sea_ice_ocean_fluxes.jl
```diff
@@ -126,6 +126,7 @@ end
     𝒬ᶠʳᶻ = fluxes.frazil_heat
     𝒬ⁱⁿ = fluxes.interface_heat
     Jˢ = fluxes.salt
+    Fη = fluxes.freshwater
```
glwagner: Should we come up with notation for this?

Physically a "free surface forcing" is weird to me, although we could use this.

J^V for volume flux or J^m for mass flux could make more sense. or J^w for water flux. (J^V is easily confused with J^v, which is the vapor flux). Other suggestions welcome to me.

src/EarthSystemModels/InterfaceComputations/component_interfaces.jl
glwagner: why is this not an AtmosphereOceanFlux as well?

glwagner (approved):
A bit confused how this is a ice-ocean flux but not atmosphere-ocean flux.

Otherwise mechanically it looks good (very simple). Mostly concerned about notation since "Feta" isn't evocative for me

simone-silvestri:
> A bit confused how this is a ice-ocean flux but not atmosphere-ocean flux.

There is also the atmosphere-ocean component but that does not need to be stored because it is assembled on the fly in the flux assembler, since turbulent fluxes return automatically a water vapor flux which is added to the precipitation. 

I ll change the notation, I like `J^w`

src/EarthSystemModels/InterfaceComputations/component_interfaces.jl
simone-silvestri: in the AtmosphereOceanFlux we already have the water_vapor flux which is (part of the) freshwater flux. We combine it with precipitation in `assemble_net_fluxes!`

xkykai (approved):
Nice! Should we add a few simple tests? Feel free to ignore as well

simone-silvestri:
> Nice! Should we add a few simple tests? Feel free to ignore as well

Sure, what should we test? That rain increases the volume? and maybe that it dilutes the salt content?

glwagner:
> > Nice! Should we add a few simple tests? Feel free to ignore as well
> 
> Sure, what should we test? That rain increases the volume? and maybe that it dilutes the salt content?

both would be ideal i suppose

xkykai:
> > Nice! Should we add a few simple tests? Feel free to ignore as well
> 
> Sure, what should we test? That rain increases the volume? and maybe that it dilutes the salt content?

Yes that is a great idea. Add some freshwater into a volume and see that volume and salt is conserved is nice.

simone-silvestri:
Added a nice conservation test. This has very loose tolerance because of https://github.com/CliMA/Oceananigans.jl/pull/5788. We can tighten them after we have updated Oceananigans

src/Oceans/ocean_simulation.jl
```diff
@@ -378,15 +414,39 @@ function hydrostatic_ocean_simulation(grid;
...
+    end
+
     # Merge user-supplied additional fluxes with defaults
     default_additional_fluxes = (u=nothing, v=nothing, T=nothing, S=nothing)
     additional = merge(default_additional_fluxes, additional_surface_fluxes)
 
+    # Freshwater tracer exchange content `Σᵢ cᵢ Jʷᵢ` (the carrying flux is the shared `Jʷ`, so the
+    # live `cᴺ Jʷ` cancels the z-star ambient carry and leaves the freshwater's own content). Pure
+    # freshwater carries no salt; the assembler fills the heat content with the enthalpy of the
+    # atmosphere freshwater entering at the surface temperature (rain − evaporation at SST).
```
simone-silvestri:
```suggestion
    # Freshwater heat content is `Σᵢ Tᵢ Jʷᵢ`, the Freshwater salinity content is assumed to be 0 for the moment (no salinity for incoming freshwater)
```

src/Oceans/Oceans.jl
```diff
@@ -177,7 +179,20 @@ function EarthSystemModels.InterfaceComputations.net_fluxes(ocean::OceananigansM
...
     ocean_surface_tracer_fluxes = NamedTuple(name => net_flux(tracers[name].boundary_conditions.top.condition) for name in keys(tracers))
-    return merge(ocean_surface_tracer_fluxes, net_ocean_surface_fluxes)
+
+    # The freshwater volume flux lives in the salinity top BC (and, on z-star grids, is the same
+    # field forcing η), so the assembler and the BC always write/read one shared field.
+    freshwater_volume_flux = extract_freshwater_flux(ocean.model.tracers.S.boundary_conditions.top.condition)
+
+    # Freshwater heat exchange shares the volume flux `η` as its carrying flux; only the
+    # source-temperature-weighted `content = Σᵢ Tᵢ Jʷᵢ` is a distinct assembler-written field.
+    heat_exchange = freshwater_exchange(ocean.model.tracers.T.boundary_conditions.top.condition)
```
simone-silvestri:
```suggestion
    freshwater_volume_flux = extract_freshwater_flux(ocean.model.tracers.S.boundary_conditions.top.condition)
    heat_exchange = freshwater_exchange(ocean.model.tracers.T.boundary_conditions.top.condition)
```

PR #250, opened 2026-05-16
Format NumericalEarth module docstring

navidcy:
The docstring for the top-level module is the README. But Documenter.jl uses markdown so before this PR there are some glitches in the formatting because the README uses html; see, e.g.:

Screenshot 2026-05-16 at 4 17 31 pm

This PR deals with these:

Screenshot 2026-05-16 at 4 17 41 pm

### DIFF
--- a/src/NumericalEarth.jl
+++ b/src/NumericalEarth.jl
@@ -1,10 +1,37 @@
 module NumericalEarth
 
+function readme_for_module_docs(path::AbstractString)
+    readme = read(path, String)
+
+    # Julia docstrings use Base.Markdown, which escapes raw HTML.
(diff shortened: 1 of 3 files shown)


src/NumericalEarth.jl
```diff
@@ -1,10 +1,37 @@
 module NumericalEarth
 
+function readme_for_module_docs(path::AbstractString)
```
giordano: Move the function inside the let block below? Otherwise the function is unnecessarily baked into the pkgimage, it's not needed for anything else

navidcy: Good point!

giordano (approved):

PR #251, opened 2026-05-16
Fix `global_climate_simulation` example

simone-silvestri:
@navidcy this should fix the problem with the output in the speedyweather coupled simulation. It is a workaround for an upstream bug in speedyweather's initialize that tries to rewrite in a closed buffer (I will open a PR upstream for the fix). 
So the solution here is to pass the output_interval directly in the `atmosphere_simulation`.


giordano:
Looks like https://github.com/CliMA/Oceananigans.jl/pull/5565 broke this package

simone-silvestri:
I m preparing a fix

navidcy (approved):
I confirm that locally this fixed the issue. Thanks!

PR #252, opened 2026-05-16
Correct orca grid for Oceananigans 0.107.7

simone-silvestri:
PR https://github.com/CliMA/Oceananigans.jl/pull/5565 removed the `continnue_south!` function used here in the ORCAGrid. This PR removes it and the metrics at `j == 1` are copied by the metrics at `j == 2`. This PR also adds a `major_basins` kwarg to remove the inland seas (only the Caspian is there in ORCA1)

PR #253, opened 2026-05-16
Update Oceananigans version in Project.toml

navidcy:
Following https://github.com/NumericalEarth/NumericalEarth.jl/pull/252


giordano (approved):
I agree, but I'm also a bit confused by compat bounds here: Oceananigans 0.108 isn't out yet.

PR #256, opened 2026-05-17
Update imports and figure size in simulation example

navidcy:
I'm not sure how the Veros extension was triggered without the CondaPkg.

It may be because of that that Veros was erroring in the docs build in #237?


giordano (approved):

navidcy:
I'm gonna merge this and update #237

PR #258, opened 2026-05-18
Support setting sea ice thermodynamics to `nothing`

taimoorsohail:
This PR adds support to set sea ice thermodynamics to `nothing`. 

## Summary

- Add `sea_ice_ocean_salinity_flux` as a `ComponentInterfaces` keyword.
- Thread the salinity-flux toggle into sea ice-ocean interface construction.
- Extend `IceBathHeatFlux` and `ThreeEquationHeatFlux` to carry `sea_ice_heat_flux` and `salinity_flux` switches.
- Allow `sea_ice_ocean_salinity_flux = nothing` to zero sea ice-ocean salt flux, matching the existing heat-flux behaviour.


navidcy:
@taimoorsohail ping us when this is ready to review

taimoorsohail:
@simone-silvestri I need some advice on this PR. Setting `sea_ice_ocean_heat_flux=nothing` bypasses the ocean/sea-ice thermodynamics altogether at the moment, because if there is no heat flux, there is no melting/freezing, and there is thus no implied salt flux either. This leads to a physically weird outcome where there is no freezing limiter and the water can supercool without freezing. 

We could borrow the `FreezingLimitedOceanTemperature` work from `OceanOnlyModel`, this would ensure that there is basal melting/freezing due to air-sea heat fluxes, and salinity fluxes associated with that melting/freezing. But then that begs the question if the user will want that level of surgical precision to turn off heat and salt fluxes independently. 

I propose we support a more simple `sea_ice_ocean_thermodynamics=nothing` kwarg, just like we set `dynamics=nothing`, NOT `x_dynamics = nothing`, `y_dynamics=nothing` etc. This means the user cannot set salinity and heat fluxes to `nothing` independently, but perhaps that is not necessary?  Note we would still need the freezing limiter that is in `OceanOnlyModel` in this case.

PR #261, opened 2026-05-19
Quality control using ExplicitImports (part 5)

### LINKED ISSUE #229: Cleaning imports

navidcy:
Add more quality assurance tests using ExplicitImports.

Progress towards https://github.com/NumericalEarth/NumericalEarth.jl/issues/229

Because it's a PR that changes a lot of files, I'm merging it in parts to avoid huge merge conflicts.


test/test_quality_assurance.jl
giordano: If all submodules are covered now, I believe that you can remove the loop over all the submodules ([`ExplicitImports.jl` is recursive by default](https://juliatesting.github.io/ExplicitImports.jl/stable/api/#Non-recursive-variants)) and just check `NumericalEarth` (which will also be quicker), that's [what we do in `Breeze.jl`](https://github.com/NumericalEarth/Breeze.jl/blob/d86b470ad9b053a787138b911bb63180170d232a/test/quality_assurance.jl#L58). I _think_ it'll also pick up extensions, as long as they're loaded (which requires loading the trigger packages, which you're doing above)

giordano (approved):
I think the test can be simplified, but looks good otherwise.

test/test_quality_assurance.jl
giordano: This is where I verified that loading the trigger packages is sufficient to test the extensions as well: https://github.com/NumericalEarth/Breeze.jl/pull/201#issuecomment-3571026063

giordano: And in fact also here you're getting the errors for the extensions [when analyzing the `NumericalEarth` module](https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/26072037219/job/76655316641?pr=261#step:12:2099)

glwagner (approved):

test/test_quality_assurance.jl
navidcy:
```Julia
Explicit Imports [NumericalEarth.DataWrangling]: Error During Test at /Users/navid/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/Documents/Research/NumericalEarth.jl-v4/test/test_quality_assurance.jl:63
  Test threw exception
  Expression: ExplicitImports.check_all_qualified_accesses_via_owners(mod) === nothing
  QualifiedAccessesFromNonOwnerException
  Module `NumericalEarth.DataWrangling` has qualified accesses to names via modules other than their owner as determined via `Base.which`:
  - `LibCURL ` has owner `LibCURL` but it was accessed from `Downloads.Curl` at `/Users/navid/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/Documents/Research/NumericalEarth.jl-v4/src/DataWrangling/DataWrangling.jl:105:52`
```

does this means that the only way around is to add LibCURL as a direct dependency?

giordano: Yes

giordano: But I'm confused by the error you showed: what symbol is it referring to?

navidcy: Perhaps something like this:

```Julia
@test ExplicitImports.check_all_qualified_accesses_via_owners(mod, ignore = (:CURLOPT_NETRC_FILE, :CURLOPT_SSL_VERIFYPEER, :LibCURL)) === nothing
```

giordano: LibCURL is an stdlib, what's the problem with depending on it?

navidcy: Oh I thought it's an extra package

navidcy:
Thanks @giordano! Have another look if you like!

ext/NumericalEarthVerosExt/veros_ocean_simulation.jl
giordano:
```suggestion
function Fields.set!(field::CFField2D, pyarray::Py, k=pyconvert(Int, pyarray.shape[2]))
```

ext/NumericalEarthVerosExt/veros_ocean_simulation.jl
```diff
@@ -1,25 +1,10 @@
 using CondaPkg
 
 using Oceananigans.Grids: topology
-using Oceananigans.Simulations
-using OffsetArrays: OffsetArray
+using Oceananigans.Fields: set!
```
giordano, line 2:
```suggestion
using Oceananigans.Fields: Fields, set!
```

navidcy: I think this isn't needed now

giordano (approved):
This cuts down 2 minutes from the test_quality_assurance test 🙁 but there's still an error in the data download job

navidcy:
Working on it!

PR #267, opened 2026-05-21
Regularize halo-fold coupling flux fields

taimoorsohail:
## Summary
Regularize sea-ice/ocean coupling fields at halo and fold boundaries before they are reused in tripolar-grid flux calculations.

## Problem
The branch `ts/try-halo-fold-diagnosis` fixes a halo-fold issue in `NumericalEarth.jl`. Several interface and assembled boundary-flux fields were being consumed before their halo or fold values were refreshed, which can lead to inconsistent reads across fold-adjacent cells on tripolar grids.

## Fix
Call `fill_halo_regions!` on the relevant momentum, tracer, and assembled net-flux fields at the points where they are about to be reused by later kernels or by the ocean boundary-condition machinery. This keeps the fold treatment delegated to each field's own zipper boundary condition instead of hard-coding a pivot choice into the coupling code.

## Files changed
- `src/EarthSystemModels/InterfaceComputations/sea_ice_ocean_fluxes.jl`
- `src/Oceans/assemble_net_ocean_fluxes.jl`

## Changes in each file
- `src/EarthSystemModels/InterfaceComputations/sea_ice_ocean_fluxes.jl`: import `fill_halo_regions!` and refresh the sea-ice momentum flux halos before computing scalar ice-ocean fluxes.
- `src/EarthSystemModels/InterfaceComputations/sea_ice_ocean_fluxes.jl`: refresh the source ocean tracer halos before the coupling kernel reads fold-adjacent cells.
- `src/EarthSystemModels/InterfaceComputations/sea_ice_ocean_fluxes.jl`: refresh the computed scalar surface flux fields, plus `Tˢⁱ` and `Sˢⁱ` for the three-equation formulation, before downstream assemblers consume them.
- `src/Oceans/assemble_net_ocean_fluxes.jl`: import `fill_halo_regions!` and refresh the assembled net ocean boundary flux fields (`T`, `S`, `u`, `v`) before they are used as ocean tracer and momentum boundary conditions.

## Validation
Not run locally per repository instructions; validation will come from GitHub-managed CI.


taimoorsohail:
@simone-silvestri Codex found this bug in the (lack of halo filling) in some of the terms in the fluxes in NumericalEarth.jl. 

While I think it is a legitimate bug (hence the PR), it didn't make a difference to the numerical instabilities around the halo/fold region in the 1/6th run.

simone-silvestri (approved):

navidcy:
Is this the culprit that we were seeing that suspiciously horizontal line of tracer you think?

taimoorsohail:
Yes that's what @simone-silvestri and I figured...

simone-silvestri:
Shall we merge?

simone-silvestri:
> Is this the culprit that we were seeing that suspiciously horizontal line of tracer you think?

Probably only for the RightFaceFolded grid, so I am not sure about a RightCenterFolded

Issue #265, opened 2026-05-20
CI doesn't really work -- Download issues


glwagner:
I think that CI succeeds 1 out of 10 times. It seems to be because of the downloading. Should we stop downloading in CI?

We've tried to implement fallbacks to predownloaded artifacts, but we still seem to get failures.

One idea is to set up downloading tests more on a nightly basis, and have some kind of dashboard to let us know the uptime of various resources we are trying to support via downloading. That removes downloading from the per-commit CI.

navidcy:
I was thinking whether we want to split DataWrangling out to a different package

glwagner:
My vote for now is to keep it... I think fewer packages is better / more productive. But we do have to figure out testing.

glwagner:
@giordano if we are using nightly tests, what is the canonical way to stay on top of the test status? I imagine if we run data downloading tests nightly, many links will fail with some regularity. We then want to compute some kind of "uptime" statistic. If data is never downloadable, we may as well stop supporting it. But this would only be known after some time of collecting statistics, so we need a way of doing that, and also monitoring it.

I also ran into an issue where downloading from NumericalEarthArtifacts failed! So it's not just external providers that are an issue...

giordano:
Can you point me to the download failures? What are the errors exactly? 

> [@giordano](https://github.com/giordano) if we are using nightly tests, what is the canonical way to stay on top of the test status? I imagine if we run data downloading tests nightly, many links will fail with some regularity. We then want to compute some kind of "uptime" statistic. If data is never downloadable, we may as well stop supporting it. But this would only be known after some time of collecting statistics, so we need a way of doing that, and also monitoring it.

I'm not aware of ready solutions, but sounds like you want a sort of dashboard of the results collected daily.

glwagner:
> I'm not aware of ready solutions, but sounds like you want a sort of dashboard of the results collected daily.

Another idea is to simply post results to a slack channel.

A bot can probably compute statistics from github right? We could run both nightly (eg collect statistics + post nightly job status) and post results to slack

glwagner:
I opened a PR with no changes to see what would happen, and got this failures:

https://buildkite.com/clima/oceananigans-distributed/builds/14665/list?jid=019e4b2a-504f-4742-8c34-f6d104aa2314&tab=output

^^ that's wrong I meant to write:

[...]

PR #150, opened 2026-03-31
[CI] Make space on runners in GPU jobs

giordano:

PR #272, opened 2026-05-21
[CI] Delete all CUDA toolkits on the host

giordano:
Ref: https://github.com/NumericalEarth/NumericalEarth.jl/issues/265#issuecomment-4513006126. I suspect we don't need any CUDA toolkit on the host, so we can delete all of them.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,7 +73,7 @@ jobs:
           echo " --> df -a /"
           df -a /
       - uses: actions/checkout@v6
-      - uses: julia-actions/setup-julia@v2
+      - uses: julia-actions/setup-julia@v3
         with:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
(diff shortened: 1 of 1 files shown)


navidcy (approved):

giordano:
This is now deleting stuff again
Before cleanup: https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/26255063659/job/77275655003?pr=272#step:7:18
```
Filesystem     Type      Size  Used Avail Use% Mounted on
overlay        overlay   100G   73G   28G  73% /
```
After cleanup: https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/26255063659/job/77275655003?pr=272#step:9:18
```
Filesystem     Type      Size  Used Avail Use% Mounted on
overlay        overlay   100G   64G   37G  64% /
```
We barely cleaned 10 GB, we used to clean 30 GB two months ago, when we were deleting only the toolkits 12.x: https://github.com/NumericalEarth/NumericalEarth.jl/pull/150#issuecomment-4164480927. It's also interesting that before cleanup here we had 28 GB available, against the 16 GB in https://github.com/NumericalEarth/NumericalEarth.jl/issues/265#issuecomment-4513006126. This all seems to be a bit inconsistent.

giordano:
I'm going to merge this PR because it does achieve the goal of recovering some space on disk, I'll later look into other options to save more space.

PR #259, opened 2026-05-18
MetadataSet (#235): core + adoption sweep

glwagner:
Draft PR for issue #235. Covers Stages 1 and 3 of the staged rollout — the core type, the auto-routing `set!`, the global alias registry, the `download_dataset → download` rename, the ERA5 batched specialization, the docs additions, and the example/test adoption sweep. Stage 2 (PrescribedAtmosphere refactor) follows in a separate PR.

## Stage 1 — Core type + machinery

**Type + accessors (`64ccbd70`)**
- `MetadataSet{V,D,R,N,F}` in `src/DataWrangling/metadata.jl`, mirroring the `Metadata` field layout (`name → names`, `filename → filenames`).
- Varargs constructor: `MetadataSet(:var1, :var2, ...; dataset, dates [or date], region, dir)`.
- Property and indexed access: `mset.temperature`, `mset[:temperature]`, `mset[1]` all yield a `Metadata`/`Metadatum`. Struct fields stay reachable (`mset.dataset`, `mset.dir`).
- Variable-axis iteration: `for m in mset` walks variables. `keys`/`length`/`firstindex`/`lastindex`/`NamedTuple(mset)`/`metadata_path(mset)` implemented.

**Global alias registry (`64ccbd70`)**
- `variable_aliases::Dict{Symbol,Symbol}` at top of `src/DataWrangling/DataWrangling.jl`. 28 entries, every value traceable to `docs/src/appendix/notation.md` (or [Breeze's notation.md](https://github.com/CliMA/Breeze.jl/blob/main/docs/src/appendix/notation.md) for microphysics).
- Synonyms retained as domain disambiguators (`:u_velocity` / `:eastward_velocity` / `:eastward_wind` all → `:u`).

**Notation rows (`64ccbd70`)**
- `docs/src/appendix/notation.md`: new "Net surface freshwater fluxes" subsection with `Jʳⁿ` and `Jˢⁿ`.

**`MetadatumSet` removed (`fa42b037`)**
- Single-date vs multi-date dispatch happens at the element level (`mset[name]` is a `Metadata` or `Metadatum`).

**Auto-routing `set!` (`8a098a27`)**
- `set!(model, mset::MetadataSet)` — one method, generates kwargs from `variable_aliases` and forwards to `set!(model; kw...)`. Variables absent from the map are silently skipped.
- `set!(::NamedTuple, mset)` — explicit per-variable form with no alias translation.

**`download_dataset → download` rename (`f8fbe840`)**
[...]

PR #274, opened 2026-05-22
Filter set!(model, ::MetadataSet) by per-model accepted kwargs

glwagner:
## Summary
- Fixes the docs-build failure introduced by #259 in `examples/one_degree_simulation.jl`: `set!(ocean.model, ecco_set)` and `set!(sea_ice.model, ecco_set)` both threw `ArgumentError: name h not found in model.velocities, model.tracers, or model.free_surface` because the generic `set!(model, ::MetadataSet)` forwarded *every* glossary-mapped variable to the model, but real `HydrostaticFreeSurfaceModel`/`SeaIceModel` throw on unknown kwargs.
- Add direct dispatches `set!(::HydrostaticFreeSurfaceModel, ::MetadataSet)` and `set!(::SeaIceModel, ::MetadataSet)` that filter the routed kwargs to those the model actually consumes.
- Keep the generic `set!(model, ::MetadataSet)` fallback simple: deliver every glossary-mapped variable. Permissive stub/ad-hoc models continue to work; strict models override to filter.

## Design
Plain multiple dispatch — no `accepted_metadata_names` hook in between. Per-model logic lives next to its dispatch, and a future model can do more than just filter (transform, validate, multi-pass) without having to extend a second function. A small `_set_glossary!` helper keeps the kwargs-construction code DRY across the two dispatches.

## Why this got through CI
The existing `set!(model, mset)` tests in `test/test_metadata_set.jl` use a permissive `StubModel` whose `set!` accepts arbitrary kwargs, so the dispatch's failure to filter wasn't exercised. Docs build *did* fail on the #259 merge commit but the run shows as `cancelled`, so it wasn't a hard block.

## Test plan
- [x] `julia --project=. test/test_metadata_set.jl` passes locally, including a new `RestrictiveStubModel` regression test that mimics the real models' kwarg-strictness with a shared 4-variable MetadataSet and demonstrates the override pattern.
- [x] Smoke test against real models confirms `which(set!, (HydrostaticFreeSurfaceModel, MetadataSet))` and `which(set!, (SeaIceModel, MetadataSet))` route to the specialised methods, not the fallback.
- [ ] CI docs build passes (this is the failing case from main).

### DIFF
--- a/src/DataWrangling/metadata.jl
+++ b/src/DataWrangling/metadata.jl
@@ -499,14 +499,16 @@ end
 """
     set!(model, mset::MetadataSet)
 
-Set fields of `model` from the variables in `mset`, auto-routing verbose
-dataset variable names to short model field-names via the global
-[`variable_glossary`](@ref) registry.
+Generic fallback: route every glossary-mapped variable in `mset` to
+`set!(model; kwargs...)`. Variables not in [`variable_glossary`](@ref) are
+silently skipped.
 
-Variables in `mset` that the target `model` does not consume are silently
-skipped — this lets partial application across coupled-model components work
-naturally. For example, a single 4-variable set can drive both an ocean and a
-sea-ice model:
+This fallback works for permissive models that accept arbitrary kwargs but
+will fail on models (e.g. `HydrostaticFreeSurfaceModel`, `SeaIceModel`) that
+throw `ArgumentError` on unknown kwargs. Those models override this method to
+filter the routed kwargs to those they actually consume — see the methods
+below. The overrides let a single multi-component MetadataSet drive both an
+ocean and a sea-ice model:
 
 ```julia
 mset = MetadataSet(:temperature, :salinity,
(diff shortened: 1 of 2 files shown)


navidcy:
Yeah! Thanks! I was about to open an issue!
I'll try with this PR!

glwagner:
i want to merge bc this is urgent (0.5.0 is broken). can you approve?

glwagner:
i'll merge because tests pass.

navidcy:
Hm... Did the test capture the case in the one_degree_model?
I can't test because ECCO server is down...
When I put some show methods in the `set!(model::HydrostaticFreeSurfaceModel, mset::MetadataSet)` method I got:

```Julia
set!(ocean.model,   ecco_set)   # picks up :temperature, :salinity → T, S
short = (propertynames(model.velocities)..., propertynames(model.tracers)..., propertynames(model.free_surface)...) = (:u, :v, :w, :T, :S, :e, :displacement, :barotropic_velocities, :filtered_state, :gravitational_acceleration, :kernel_parameters, :substepping, :timestepper)
names = (:eastward_velocity, :v_velocity, :temperature, :u_velocity, :eastward_wind, :northward_wind, :air_temperature, :northward_velocity, :salinity)
[ Info: Downloading ECCO data: temperature in /home/constantinon/.julia/scratchspaces/904d977b-046a-4731-8b86-9235c0d1ef02/ECCO/v4...
[ Info:  ... downloaded 274.0 B (100% complete, 552.080 ms)
[ Info:  ... downloaded 274.0 B (100% complete, 552.161 ms)
[ Info:  ... downloaded 274.0 B (100% complete, 552.193 ms)
[ Info:  ... downloaded 274.0 B (100% complete, 552.210 ms)
ERROR: RequestError: HTTP/2 403 while requesting https://ecco.jpl.nasa.gov/drive/files/Version4/Release4/interp_monthly/THETA/1993/THETA_1993_01.nc
```

Seems that `names` include also velocities?

navidcy:
> i want to merge bc this is urgent (0.5.0 is broken). can you approve?

I was trying to test!

navidcy:
I guess the docs built on `main` will show us.
https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/26263261928/job/77301118895

glwagner:
ah sorry! ~~yes, it looks like the MetadataSet for ECCO is including velocities, although we may not want to set them. ~~

I don't think it does:

https://github.com/NumericalEarth/NumericalEarth.jl/blob/05507eeadee0b809ec5606d98ec12f5b69690ddd/examples/one_degree_simulation.jl#L86-L88

navidcy:
What could be a clean solution for that?

glwagner:
As far as I can tell the `MetadataSet` does not include velocities. What is your conclusion? I'm not sure what you are showing up there

glwagner:
Ok so looking at the code, what you have shown is the list of _all possible names_ that can be set. It is not the list of names that _are_ set.

navidcy:
Hm... I just went here

https://github.com/NumericalEarth/NumericalEarth.jl/blob/05507eeadee0b809ec5606d98ec12f5b69690ddd/src/DataWrangling/metadata.jl#L530-L536

and put a `@show` in front of `names`. Then `names` is provided to `set!`. So these are the variables that can be set, or aren't they? Am I missing something?

glwagner:
This is what `names` says:

```julia
names = Tuple(n for (n, s) in variable_glossary if s in short)
```

this builds a tuple that contains every name in `variable_glossary` (a mapping of long to short names, containing all possible names that can be set from `MetadataSet`), for which the short name "`s`" is contained in `short` (the list of valid names for this particular model). That is what `if s in short` means.

glwagner:
Another way to write it:

```julia
    valid_long_names = Tuple(long_name for (long_name, short_name) in variable_glossary
                             if short_name in short)
    return set!(model, mset, valid_long_names)
```

glwagner:
```julia
function Fields.set!(model::HydrostaticFreeSurfaceModel, mset::MetadataSet)
    hfsm_short_names = (propertynames(model.velocities)..., propertynames(model.tracers)..., :η)
    valid_long_names = Tuple(long_name for (long_name, short_name) in variable_glossary if short_name in hfsm_short_names)
    return set!(model, mset, valid_long_names)
end
```

navidcy:
But e.g. `:u` is in `hfsm_short_names` and also in `valid_long_names`.... So I don't understand why it won't be set.

Anyway, I'll try myself and see what happens when the ECCO server heals; there's no need to try to "compile" code in our heads.

glwagner:
> But e.g. :u is in hfsm_short_names and also in valid_long_names.... So I don't understand why it won't be set.

It is correct that `:u` is in both the short names for HFSM, and is one of the valid long names.

However, it is not one of the variables contained in `mset`. Therefore, it will not be set. Does that make sense?

glwagner:
one way to understand this could be to write `@show mset`? would that help?

glwagner:
I believe this concept will apply to any dataset, so maybe try another one other than ECCO to see

navidcy:
> one way to understand this could be to write `@show mset`? would that help?

Right! I was forgetting the `mset` only includes the variables we constructed it with and _not all variables_ of the current dataset!

navidcy:
The near_global_simulation.jl seems to error now?

https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/26263261928/job/77301118895#step:7:61

navidcy:
We can deal with it within #277

navidcy:
https://github.com/NumericalEarth/NumericalEarth.jl/pull/277/commits/e1d223e3face5258461901411d35dec12da52416 deals with it; the error was because MetadataSet without a date kwarg defaults to all dates

PR #280, opened 2026-05-22
[CI] Delete some more unneeded tools in GPU jobs

giordano:
Ref #265.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,7 @@ jobs:
         with:
           files: lcov.info
           token: ${{ secrets.CODECOV_TOKEN }}
-      - name: Show available at the end
+      - name: Show available storage at the end
         if: always()
         timeout-minutes: 2
         run: |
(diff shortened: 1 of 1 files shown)


giordano:
Ok, maybe I'll avoid mounting all of root. Also, I have an idea to try and further (slightly) reduce the size of the Docker image

.github/workflows/ci.yml
```diff
@@ -88,6 +88,14 @@ jobs:
         with:
           files: lcov.info
           token: ${{ secrets.CODECOV_TOKEN }}
+      - name: Show available at the end
```
giordano: Reminder: put this at the end of the GPU job as well

navidcy (approved):

giordano:
[Before](https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/26310838846/job/77459077752?pr=280#step:7:18)
```
Filesystem     Type      Size  Used Avail Use% Mounted on
overlay        overlay   100G   86G   15G  86% /
```
[After](https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/26310838846/job/77459077752?pr=280#step:9:18)
```
Filesystem     Type      Size  Used Avail Use% Mounted on
overlay        overlay   100G   45G   56G  45% /
```
With the `df` commands at the end we should also be able to track how much space we still have at the end, and understand what partition is being filled.

Side note, every time we get a VM with a different software stack on the host, this is annoying 😩

PR #277, opened 2026-05-22
Delete unused examples + clean up examples

glwagner:
## Summary

- Prunes ten outdated standalone examples now covered by tests, doctests, or the surviving end-to-end simulations.
- Modernizes `global_climate_simulation.jl`, `near_global_ocean_simulation.jl`, and `one_degree_simulation.jl`:
  - pass Fields / AbstractOperations / Observables straight to `heatmap!` (Oceananigans Makie extension auto-NaNs immersed land cells),
  - rename component-specific velocities/temperatures to unicode subscripts (`uᵒᶜ`, `uᵃᵗ`, `uˢⁱ`, etc.),
  - in `one_degree_simulation`, drive both ocean and sea-ice init from one `MetadataSet` and add a `JRA55PrescribedLand` component.
- Replaces the illustrative ```` ```julia ```` block in the `set!(model, ::MetadataSet, names)` docstring with a runnable ```` ```jldoctest ```` that builds a real `ocean_simulation`, calls `set!(ocean.model, mset)` with a multi-component ECCO4Monthly `MetadataSet`, and reports rounded `min`/`max`/`mean` of the resulting temperature field. The sea-ice route is mentioned in a comment alongside the call (no existing test/doctest exercises the ECCO sea-ice download path, so we don't run it here).

PR #283, opened 2026-05-23
Metadataset improvements and example fixes

navidcy:
This PR pulls some of the changes in #277.

This is to ensure Docs examples can built before we debug the visualisation changes. Building all docs is ~7hrs so it's hard to iterate.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -42,7 +42,7 @@ examples = [
 
 # Developer examples from docs/src/developers/ directory
 developer_examples = [
-    Example("EarthSystemModel interface", "slab_ocean", false),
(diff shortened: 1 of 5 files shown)


glwagner (approved):

glwagner:
> This PR pulls some of the changes in #277.
> 
> This is to ensure Docs examples can built before we debug the visualisation changes. Building all docs is ~7hrs so it's hard to iterate.

Just note for the future, a good way to iterate on examples is to do them one by one (rather than building all of them every time). This can be done by commenting the examples out in `make.jl` and using the tag "build all examples".

PR #286, opened 2026-05-24
Update TagBot key for PRs and Documenter key for documentation deployment

navidcy:


giordano (approved):
🚀

.github/workflows/docs.yml
```diff
@@ -111,5 +111,5 @@ jobs:
       - name: Deploy documentation
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          DOCUMENTER_KEY: ${{ secrets.DOCUMENTER_KEY }}
+          DOCUMENTER_KEY: ${{ secrets.NUMERICALEARTHDOCUMENTATION_DOCUMENTER_KEY }}
```
giordano, line 114: Wait a second, this secret isn't in this repository

giordano: @navidcy Yes, I see the secret is set in https://github.com/NumericalEarth/NumericalEarthDocumentation, but that's wrong: it must be in this repository, otherwise this line isn't doing anything

navidcy: Done! Thanks!

PR #292, opened 2026-05-26
Revive Veros ocean example

navidcy:
Installing Veros via the GitHub action workflow seems to have done the job!


navidcy:
https://numericalearth.github.io/NumericalEarthDocumentation/previews/PR292/literated/veros_ocean_forced_simulation/

navidcy:
Although next time the Docs built will overwrite this preview with one that doesn't have Veros (since the Veros example is `false` by default.)

But it works now, I think we should merge this back.

A proof that the preview worked with Veros below:

Screenshot 2026-05-27 at 2 12 37 pm

giordano (approved):

simone-silvestri (approved):

PR #305, opened 2026-05-27
Patch for Oceananigans v0.109

navidcy:
Oceananigans https://github.com/CliMA/Oceananigans.jl/pull/5604 changed the argument signature of `maybe_prepare_first_time_step!` which NumericalEarth extends. This PR fixes this.


docs/make.jl
```diff
@@ -42,7 +42,7 @@ examples = [
     Example("Near-global ocean simulation", "near_global_ocean_simulation", false),
     Example("Global climate simulation", "global_climate_simulation", false),
     Example("Veros ocean simulation", "veros_ocean_forced_simulation", false),
-    Example("Breeze over four oceans", "breeze_over_four_oceans", false),
+    # Example("Breeze over four oceans", "breeze_over_four_oceans", false),
```
taimoorsohail: Why commenting this?

docs/make.jl
```diff
@@ -211,7 +211,7 @@ makedocs(; sitename = "NumericalEarth.jl",
          clean = true,
          warnonly = [:cross_references, :missing_docs],
          checkdocs = :exports,
-         linkcheck = true,
+         linkcheck = false,
```
taimoorsohail: I'm guessing this is because the ECCO server is down?

taimoorsohail (approved):

glwagner (commented):
I think we have to remove the `[sources]` before merging

docs/make.jl
navidcy: Yes, and things keep failing.

giordano: Can filter out the unreliable server, instead of completely disabling the check.

navidcy: True...

My bad. I was adding one by one in the linkcheck_filter but then I got worn out by the constant fail of yet another link of another dataset and turned to `false`.

giordano: If it's only one, or few, server failing (I haven't seen the error), can match all the links in one go, the ignore pattern is a regular expression

docs/Project.toml
```diff
@@ -23,13 +22,12 @@ SpeedyWeather = "9e226e20-d153-4fed-8a5b-493def4f21a9"
...
+ConservativeRegridding = {rev = "ncc/add-oceananigans-v0.109", url = "https://github.com/JuliaGeo/ConservativeRegridding.jl"}
 NumericalEarth = {path = ".."}
 
 [compat]
 CUDA = "~6.0"
 CairoMakie = "0.15"
-ClimaSeaIce = "0.5.0"
 Documenter = "1"
 DocumenterCitations = "1.3"
 Literate = "2.2"
```
simone-silvestri:
```suggestion
NumericalEarth = {path = ".."}

[compat]
CUDA = "~6.0"
CairoMakie = "0.15"
ConservativeRegridding = "0.2.3"
Documenter = "1"
DocumenterCitations = "1.3"
Literate = "2.2"
```

Project.toml
```diff
@@ -61,7 +61,7 @@ ArchGDAL = "0.10"
 Breeze = "0.4, 0.5"
 CDSAPI = "2.2.1"
 CFTime = "0.1, 0.2"
-ClimaSeaIce = "0.5"
+ClimaSeaIce = "0.5.2"
```
simone-silvestri:
```suggestion
ClimaSeaIce = "0.5.3"
```
while we are at it :)

glwagner (approved):

navidcy:
Oceananigans 0.109.1 proper patch didn't solve the issue with checkpointer. I don't know what it might be. Perhaps the PR with set!(::Field, ::Field)? Perhaps the implicit w advection and the maybe_prepare_timestep?

navidcy:
I tried to dig into the checkpointer test breaking:

- I went as far back as NumericalEarth v0.5.0 and still the issue remains. So it's probably not something NumericalEarth changed but some change in an upstream package?
- The test fails at
https://github.com/NumericalEarth/NumericalEarth.jl/blob/1b7881ea3d3b88581e25df05bfc6e2467c60de7f/test/test_checkpointer.jl#L118-L124
and *the biggest discrepancies* come form `u` and `ui` (just mentioning in case this is illuminating for somebody).

navidcy:
What puzzles me is that even if there were upstream changes the initial simulation will include them and the checkpointed one would also include them. So I don't understand...

Perhaps something changed in the way we set or restore and something doesn't get restored leading to discrepancies in u fields (or velocities in general)?

glwagner:
Claude thinks its the fix on Oceananigans 5506 (108.1)

navidcy:
OK, new data:

For the [last commit](0c6cdfacffd6b57db604bd40e765a17cc8d1a089) where the checkpointer test was passing, ClimaSeaIce v0.5.0 was used. If I rerun on that commit with ClimaSeaIce v0.5.1 test fails.

This is the diff between the two ClimaSeaIces: https://github.com/CliMA/ClimaSeaIce.jl/compare/v0.5.0...v0.5.1

Is this illuminating?

navidcy:
> Claude thinks its the fix on Oceananigans 5506 (108.1)

Really Claude?

https://github.com/CliMA/Oceananigans.jl/pull/5506 ??

navidcy:
I'm not even going to investigate Claude's suggestion.

And if it turns out they are right, I'll stop coding perhaps?

glwagner:
Is the checkpointer test actually exercising something in NumericalEarth? Why does it appear to have a test that Oceananigans doesn't catch?

glwagner:
I think I mistranscribed the PR number. Please don't stop coding.

glwagner:
Or perhaps there is not a checkpointer test in ClimaSeaIce

navidcy:
> Is the checkpointer test actually exercising something in NumericalEarth? Why does it appear to have a test that Oceananigans doesn't catch?

Yes. It checks that esm model and all components can be checkpointed and restored

navidcy:
There is a checkpointed test in ClimaSeaIce but I’ll look into it closely

navidcy:
locally test_checkpointer passes now after https://github.com/CliMA/ClimaSeaIce.jl/pull/151 and ClimaSeaIce 0.5.5; thanks @glwagner

navidcy:
If CI passes here feel free to merge this!

PR #310, opened 2026-05-29
Fix radiation clock

simone-silvestri:
Recently, in OMIP simulations, the simulation would crash at about 1600 days of evolution.
This crash is not related to the simulation exploding, rather because of the clocks de-synching, which happens easily if we have clocks with different floating types across 1600 days. This on the GPU becomes weirdly enough a NaN, probably because `@inbounds` allows the GPU to pick garbage memory.

This is the simulation with `--check-bounds=yes` on the CPU.
```julia
[ Info: time: 1638.625 days, iteration: 95937, Δt: 30 minutes, max(h): 3.74e+00 m, max(ℵ): 1.00e+00 extrema(T, S): (-1.89, 34.56) ᵒC, (5.14, 49.49) psu maximum(u): (1.60e+00, 1.65e+00, 1.87e-03) m/s, wall time: 18.384 seconds
[ Info: time: 1638.646 days, iteration: 95938, Δt: 30 minutes, max(h): 3.74e+00 m, max(ℵ): 1.00e+00 extrema(T, S): (-1.89, 34.56) ᵒC, (5.14, 49.49) psu maximum(u): (1.60e+00, 1.65e+00, 1.92e-03) m/s, wall time: 18.412 seconds
[ Info: time: 1638.667 days, iteration: 95939, Δt: 30 minutes, max(h): 3.75e+00 m, max(ℵ): 1.00e+00 extrema(T, S): (-1.89, 34.58) ᵒC, (5.15, 49.49) psu maximum(u): (1.60e+00, 1.65e+00, 1.96e-03) m/s, wall time: 18.391 seconds
[ Info: time: 1638.688 days, iteration: 95940, Δt: 30 minutes, max(h): 3.76e+00 m, max(ℵ): 1.00e+00 extrema(T, S): (-1.89, 34.62) ᵒC, (5.15, 49.49) psu maximum(u): (1.60e+00, 1.65e+00, 1.93e-03) m/s, wall time: 18.532 seconds
[ Info: time: 1638.708 days, iteration: 95941, Δt: 30 minutes, max(h): 3.77e+00 m, max(ℵ): 1.00e+00 extrema(T, S): (-1.89, 34.68) ᵒC, (5.15, 49.48) psu maximum(u): (1.60e+00, 1.64e+00, 1.88e-03) m/s, wall time: 18.566 seconds
[...]

PR #321, opened 2026-06-02
Pass clocks in simulations

simone-silvestri:
This is an initial PR that allows passing clocks to simulations. So we can do something like
```julia
start_date = DateTime(1958, 1, 1)
ocean = ocean_simulation(grid, clock = Clock(time=start_date))
sea_ice = sea_ice_simulation(grid, clock = Clock(time=start_date))
atmos = JRA55PrescribedAtmosphere(; start_date)
```

### DIFF
--- a/src/Oceans/nonhydrostatic_ocean_simulation.jl
+++ b/src/Oceans/nonhydrostatic_ocean_simulation.jl
@@ -12,6 +12,7 @@ using SeawaterPolynomials.TEOS10: TEOS10EquationOfState
                                     advection = WENO(order=9),
                                     forcing = NamedTuple(),
                                     boundary_conditions::NamedTuple = NamedTuple(),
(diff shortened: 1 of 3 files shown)


navidcy:
Has this broke the slab ocean--breeze example or was it already broken?

https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/26837662766/job/79135845297?pr=321#step:8:64

simone-silvestri:
I think that was already broken after PR #310 because the slab ocean and breeze use different clock types. I will open a PR to fix it

simone-silvestri:
This PR is required to switch climacoupler to the new climaocean version @hgpeterson

hgpeterson (approved):
Sounds good @simone-silvestri. Was there a PR for this already in ClimaOcean and this is just a PR to move that here?

PR #148, opened 2026-03-30
Add Terrarium extension module with rudimentary coupling

bgroenks96:
This is an initial attempt at creating an extension module for [Terrarium](https://github.com/NumericalEarth/Terrarium.jl).

I think for now we can get away with just assuming that the land grid always matches the `exchange_grid` thus avoiding any need for regridding.

Prerequisities:
- [x] Register Terrarium
- [x] Add `Lands` submodule and component in `EarthSystemModel`
- [x] Merge NumericalEarth/Terrarium.jl#155

The following is a rough implementation plan generated by Claude (with feedback):

# Implement `NumericalEarthTerrariumExt` coupling module

> Status: **implemented (baseline) + ERA5 example scoped**. Terrarium's `LandModel` couples into NumericalEarth's `EarthSystemModel` as the land component with surface turbulent fluxes computed by NumericalEarth (Monin-Obukhov) and consumed by Terrarium; radiation is diagnosed locally by Terrarium; the water budget stays internal to Terrarium. End-to-end verified on CPU (`AtmosphereLandModel` steps; ground-heat-flux residual ≈ −249 W/m²). Remaining: the ERA5-forced example (scoped below), albedo export to NumericalEarth radiation, and the conductance hybrid (Phase 2).
>
[...]

PR #323, opened 2026-06-03
(0.6.0) Conservative variably saturated SlabLand + evaporation-front humidity

glwagner:
## Summary

- Adds `VariablySaturatedBucketHydrology` + `WaterCoupledForceRestoreEnergy`
  + `EvaporationFrontHumidity` to replace the clamped bucket, the
  constant-`C` force-restore, and the prescribed-`β(𝒮)` surface humidity.
  All three plug into the existing `SlabLand` and `ComponentInterfaces`
  scaffolding via per-closure declarations — legacy `BucketHydrology +
  SlabEnergy + FractionalHumidity` configurations stay unchanged.
- Storage variable is the augmented liquid fraction
  `ϑˡ = θˡ + Sₛ·max(Π, 0)`, so `Mˡᵃ > Mˡᵃ⁺` is admitted as positive
  pressure head rather than clamped. The surface humidity `qⁱⁿ` is solved
  by a dry-layer vapor-flux balance against an unresolved evaporation
  front at saturation-dependent depth `δᵛ(𝒮)`, collapsing exactly to a
  saturated skin when `𝒮 ≥ 𝒮ᶜ`.
- Converts `breeze_over_slab_land.jl` and `era5_forced_slab_land.jl` to
  the new closures; adds a tutorial-style docs page; extends
  `notation.md` with 22 new symbols.

See `docs/src/land/evaporation_front_slab_land.md` for the SlabLand
tutorial (continuous balances, depth-integrated reduction, the Picard
fixed-point algorithm, parameter guidance, numerical-robustness notes,
and a worked configuration).

### What changed at the closure level

| Closure | Role |
|---|---|
| `VariablySaturatedBucketHydrology` | Conservative `dMˡᵃ/dt = Jˡ_b − Jˡ_s − Jᵛ − Rᴹ_lat` budget; augmented liquid fraction storage |
| `WaterCoupledForceRestoreEnergy` | Force-restore with `Mˡᵃ`-dependent `C(Mˡᵃ) = C_dry + cˡ Mˡᵃ`; conservative `dTˡᵃ/dt` (adding water at slab T leaves T invariant) |
| `VanGenuchtenRetention` / `VanGenuchtenConductivity` | Retention curve `Π(𝒮)` and Mualem `K(𝒮)` |
| `NoDeepLiquidFlux`, `FreeDrainageFlux`, `DarcyDeepLiquidFlux`, `LinearReservoirDrainage` | Bottom-boundary liquid flux closures |
| `NoRunoff`, `InfiltrationCapacityRunoff` | Surface/subsurface runoff |
| `EvaporationFrontHumidity` | Dry-layer vapor balance for `qⁱⁿ`; sub-closures `StorageBasedEvaporationFrontDepth`, `DryLayerVaporPistonVelocity`, `ConstantTortuosity` / `MillingtonQuirk`, `UnitWaterActivity` |

`SlabLand` gains a closure-extensible `diagnostics::NamedTuple` slot and
the flux assembly now populates signed flux fields (`vapor_flux`,
`surface_energy_flux`, `liquid_precipitation_flux`) alongside the legacy
positive-part fields (`evaporation`, `precipitation`, `net_energy_flux`).
Backward-compatible — legacy configurations get an empty diagnostics
tuple and only the legacy flux fields are written/read.

### Verified analytically

- No-flux + evaporation + capacity-limited runoff + free drainage
  conservation
- Force-restore relaxation `T(t) = Tᵈᵉᵉᵖ + (T₀ − Tᵈᵉᵉᵖ)·exp(−Λt/C)`
- Advective drainage at slab T → ΔT = 0 (bit-exact)
- Vapor divider `qⁱⁿ = (Gᵉ qᵉ Δq + Jᵃ qᵃᵗ) / (Gᵉ Δq + Jᵃ)` (ε_machine)
- Wet-branch `qⁱⁿ = qᵛ⁺(Tⁱⁿ)`
- Coupled Breeze wet/dry contrast: ~10² ratio in `Jᵛ`

### Out of scope (deferred to follow-ups)

Listed in `docs/src/land/evaporation_front_slab_land.md` §9:

- Snow, sea ice, vegetation, multi-layer soil columns, river routing.
- `Eˡᵃ` as a first-class state variable (needed for phase change).
- `MatricPotentialActivity` (Kelvin-equation suction-driven vapor reduction).
- Land-side `SkinTemperature(DiffusiveFlux)` solve so `Tⁱⁿ ≠ Tˡᵃ`.
- More runoff closures, Brooks–Corey retention, two-node front energy
  solve, implicit deep Darcy, subgrid tile blending.

## Test plan

- [x] `test/test_slab_land.jl` — 35 existing assertions
- [x] `test/test_breeze_coupling.jl` — 14 existing assertions
- [x] `test/test_variably_saturated_bucket_hydrology.jl` — diagnostics +
      conservation (10)
- [x] `test/test_water_coupled_force_restore_energy.jl` — analytic
      force-restore + advective invariance (3)
- [x] `test/test_evaporation_front_humidity.jl` — vapor divider, wet
      branch, Tᵉ interpolation, edge cases (5)
- [x] `test/test_breeze_evaporation_front_land.jl` — coupled wiring +
      drydown (16)
- [ ] CI on CPU
- [ ] CI on GPU
- [ ] Documentation builds; both converted Breeze + ERA5 examples render


bgroenks96:
A few initial comments before going through everything:

> Out of scope (deferred to follow-ups)

Most of what is listed here should be out-of-scope for the `SlabLand` model generally, I think, and it would make more sense to direct some effort into making progress on #148.

For the Van Genuchten retention curve, it might be nice if you can make use of [FreezeCurves.jl](https://cryogrid.github.io/FreezeCurves.jl/dev/) like Terrarium does, to avoid more code duplication.

examples/breeze_over_slab_land.jl
```diff
@@ -224,18 +258,31 @@ set_to_mean!(reference_state, atmos.model, rescale_densities = true)
...
+# a Fickian vapor-flux balance between the saturated soil at depth `δᵛ` and
+# the atmosphere. The wet center has `δᵛ = 0` (saturated skin); the dry edges
+# have `δᵛ → δᵛ_max` and the dry-layer piston velocity `wᵈ = Dᵛ_eff/δᵛ`
+# limits evaporation self-consistently.
+interface_specific_humidity = EvaporationFrontHumidity(;
+    evaporation_front_depth = StorageBasedEvaporationFrontDepth(
+        maximum_front_depth = 0.05,
+        critical_saturation = 0.5,
+        front_depth_exponent = 2),
+    vapor_exchange = DryLayerVaporPistonVelocity(
```
bgroenks96, line 266: What does "dry-layer piston velocity" mean?

src/EarthSystemModels/InterfaceComputations/evaporation_front_humidity.jl
```diff
@@ -0,0 +1,297 @@
...
+
+#####
+##### Evaporation-front depth diagnostics
+#####
+
+"""
+    StorageBasedEvaporationFrontDepth(maximum_front_depth, critical_saturation,
+                                      front_depth_exponent)
+
+Diagnostic evaporation-front depth `δᵛ` as a function of land saturation `𝒮`:
```
bgroenks96: The notation for saturation is not included in the table above. It's also a bit confusing to use `\mathcal{S}` when you use `S_s` for specific storage. I would suggest using consistent (but distinct) symbols for all volumetric contents (upper bounded by constituent fraction) and saturation fractions (relative to constituent).

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
```diff
@@ -0,0 +1,297 @@
...
+
+Parameters of the dry-layer vapor piston velocity `wᵈ = Dᵛ_eff / max(δᵛ, δᵛ_min)`.
+The tortuosity model is a singleton type — [`ConstantTortuosity`](@ref) or
+[`MillingtonQuirk`](@ref) — dispatched on by
+`effective_vapor_diffusivity`.
+"""
+struct DryLayerVaporPistonVelocity{FT, T}
+    minimum_front_depth   :: FT
+    molecular_diffusivity :: FT
+    tortuosity_model      :: T
```
bgroenks96: Since "model" already has a specific meaning in the Oceananigans/NumericalEarth ecosystem, I would suggest maybe not referring to this as such.

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
```diff
@@ -0,0 +1,297 @@
...
+[`DryLayerVaporPistonVelocity`](@ref) when the soil air space is not modeled
+explicitly.
+"""
+struct ConstantTortuosity end
+
+"""
+    MillingtonQuirk()
+
+Millington–Quirk tortuosity: `Dᵛ_eff = Dᵛ₀ · θᵍ^(10/3) / ν²` where
+`θᵍ = ν − θˡ` is the gas-filled pore fraction. Reduces vapor diffusivity in
```
bgroenks96, line 124: I don't understand what the purpose of this is. Are you trying to explicitly simulate water vapor diffusion between the pore space and the atmosphere? That's a level of detail far beyond slab land model... AFAIK most spatially resolving land/hydrology models don't even do this. Can you provide a reference?

ClimaLand does not do this, as far as I am aware. A [quick search](https://github.com/search?q=repo%3ACliMA%2FClimaLand.jl%20tortuosity&type=code) of the codebase seems to indicate that tortuosity is only explicitly considered in the CO2 model.

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
```diff
@@ -0,0 +1,297 @@
...
+  reduction via the Kelvin equation, only relevant at extreme dryness — is a
+  follow-up).
+"""
+struct EvaporationFrontHumidity{EFD, VEX, FT, A, Φ}
+    evaporation_front_depth :: EFD
+    vapor_exchange          :: VEX
+    thermal_exchange_depth  :: FT
+    porosity                :: FT
+    water_activity          :: A
+    phase                   :: Φ
```
bgroenks96: Why is this here? Evaporation is by definition liquid, no? From solid phase it would be sublimation.

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
```diff
@@ -0,0 +1,297 @@
...
+saturated-skin BC when the slab is wet enough (`𝒮 ≥ 𝒮ᶜ`).
+
+* `evaporation_front_depth` — depth diagnostic, e.g.
+  [`StorageBasedEvaporationFrontDepth`](@ref).
+* `vapor_exchange` — `δᵛ_min`, `Dᵛ₀`, tortuosity (a
+  [`DryLayerVaporPistonVelocity`](@ref)).
+* `thermal_exchange_depth` — `ℓᵀ` (m), the same depth used by
+  `SkinTemperature(DiffusiveFlux)` on the temperature side. Controls the
+  interpolation `Tᵉ = Tⁱⁿ + χ(Tˡᵃ − Tⁱⁿ)` with `χ = clip(δᵛ/ℓᵀ, 0, 1)`.
+* `porosity` — `ν`, soil porosity (matches the hydrology closure; needed
```
bgroenks96, line 207: So this parameter is being duplicated across components?

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
```diff
@@ -0,0 +1,297 @@
...
+
+@inline function effective_vapor_diffusivity(::MillingtonQuirk, D₀, ν, θˡ)
+    FT = typeof(θˡ)
+    νF = convert(FT, ν)
+    θᵍ = max(νF - θˡ, zero(FT))
+    return convert(FT, D₀) * θᵍ^(FT(10//3)) / νF^2
+end
+
+#####
+##### Interface-state hooks. Reads saturation, bulk land temperature, and
```
bgroenks96: Why are these `####` comments? Is this some kind of custom docstring tooling that I am unaware of?

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
```diff
@@ -0,0 +1,297 @@
...
+##### Humidity solver — the headline of the closure.
+#####
+##### Sign convention for fluxes here matches `SkinHumidity` exactly: each
+##### `Jᵃ = -ρᵃᵗ u★ q★` (a positive Jᵃ means vapor moves upward into the
+##### atmosphere; q★ < 0 when evaporating). The dry-layer flux Jᵉ from the
+##### front up to the interface is taken as `Gᵉ(qᵉ - qⁱⁿ)`, so a wetter front
+##### (qᵉ > qⁱⁿ) drives vapor upward. The balance Jᵉ = Jᵃ closes for qⁱⁿ.
+#####
+@inline function compute_interface_humidity(q::EvaporationFrontHumidity, Tₛ, Ψₛ, Ψₐ, Ψᵢ, ℙₐ)
+    ℂᵃᵗ = ℙₐ.thermodynamics_parameters
```
bgroenks96, line 318: I think it's very weird to assign mathematical notation to a data structure like this. At least I was surprised. When I first read the argument line, I expected this to be pressure or something.

I guess the idea is that it's set notation, and it's kind of a "set" of parameters?

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
```diff
@@ -0,0 +1,297 @@
...
+    u★  = Ψₛ.fluxes.u★
+    q★  = Ψₛ.fluxes.q★
+    qⁱⁿ⁻ = Ψₛ.specific_humidity
+    Jᵃ   = -ρᵃᵗ * u★ * q★               # positive upward
+    Δq   = qⁱⁿ⁻ - qᵃᵗ
+
+    # Same Δq-multiplied form as SkinHumidity to stay finite when Δq → 0:
+    #   qⁱⁿ (Gᵉ Δq + Jᵃ) = Gᵉ qᵉ Δq + Jᵃ qᵃᵗ
+    D    = Gᵉ * Δq + Jᵃ
+    qⁱⁿ★ = (Gᵉ * qᵉ * Δq + Jᵃ * qᵃᵗ) / D
```
bgroenks96: I don't recognize this formulation. Where is it coming from?

src/Lands/energy_balance/water_coupled_energy.jl
```diff
@@ -0,0 +1,213 @@
...
+    Λ  = energy.deep_conductance
+    τ  = energy.deep_time_scale
+    if Λ === nothing
+        return C / convert(FT, τ)
+    else
+        return property_value(Λ, i, j, 1)
+    end
+end
+
+@inline upwind_deep_internal_energy(Jˡb, Tᵈ, T, Tᵣ, cˡ) =
```
bgroenks96: This is something that we're actually missing at the moment in Terrarium, so would be nice to add it there too.

src/Lands/energy_balance/water_coupled_energy.jl
```diff
@@ -0,0 +1,213 @@
...
+    ifelse(Jˡb > 0, cˡ * (Tᵈ - Tᵣ), cˡ * (T - Tᵣ))
+
+@inline upwind_surface_internal_energy(Jˡs, T, Tˡp, Tᵣ, cˡ) =
+    ifelse(Jˡs > 0, cˡ * (T - Tᵣ), cˡ * (Tˡp - Tᵣ))
+
+#####
+##### Time-step kernel.
+#####
+
+@kernel function _water_coupled_step!(T, JᴱS, Tˡp,
```
bgroenks96: I guess this is just the general style of NumericalEarth, but I don't really like that I have to scroll down and look at the kernel launch function to match up the various symbols here to their English names. I think it would be better to either pack these fields into a data structure and define the mathematical symbols within the function or write their names out explicitly in the argument list.

src/Lands/energy_balance/water_coupled_energy.jl
```diff
@@ -0,0 +1,213 @@
...
+        Jˡs  = Jˡs_field[i, j, 1]
+        Rlat = Rlat_field[i, j, 1]
+        Jᴱs  = JᴱS[i, j, 1]
+        Tˡpij = Tˡp[i, j, 1]
+    end
+
+    Cdry_ij = property_value(Cdry, i, j, 1)
+    C       = Cdry_ij + cˡ * max(Mij, zero(Mij))
+
+    Tᵈ_ij = stateindex(Tᵈ, i, j, 1, grid, time, (Center, Center, Center))
```
bgroenks96: what is `stateindex`?

src/Lands/energy_balance/water_coupled_energy.jl
```diff
@@ -0,0 +1,213 @@
...
+    eb_up  = upwind_deep_internal_energy(Jˡb, Tᵈ_ij, Tij, Tᵣ, cˡ)
+    Jᴱ_adv_b = ifelse(advect_deep, eb_up * Jˡb, zero(FT))
+
+    es_up  = upwind_surface_internal_energy(Jˡs, Tij, Tˡpij, Tᵣ, cˡ)
+    Jᴱ_adv_s = ifelse(advect_surf, es_up * Jˡs, zero(FT))
+
+    eR_lat = cˡ * (Tij - Tᵣ) * Rlat
+
+    dEdt  = Jᴱ_cond + Jᴱ_adv_b - (Jᴱs + Jᴱ_adv_s) - eR_lat
+    dTdt  = (dEdt - cˡ * (Tij - Tᵣ) * dMdt) / C
```
bgroenks96: I would suggest splitting up the time stepping from the tendency calculation to reduce the cognitive load required to process each function individually.

src/Lands/hydrology/deep_liquid_fluxes.jl
```diff
@@ -0,0 +1,110 @@
...
+Zero deep liquid flux. The slab is closed at the bottom.
+"""
+struct NoDeepLiquidFlux end
+
+@inline deep_liquid_flux(::NoDeepLiquidFlux, M, θˡ, 𝒮, Π, K, ψ_D, time) = zero(M)
+
+Base.summary(::NoDeepLiquidFlux) = "NoDeepLiquidFlux"
+
+"""
+    FreeDrainageFlux(ρˡ)
```
bgroenks96: Is there a reason why these boundary conditions are specified this way rather than using `Field` boundary conditions?

bgroenks96:
One thought that occurred to me while reviewing this is that it might be helpful to adopt some additional standards regarding AI use in drafting issues and PRs. I found myself several times wondering which of the design choices here were consciously made by @glwagner in the prompt/plan vs. which originated from the LLM. This is not so important in cases where the specific formulations or conventions are common or obvious, but it definitely matters when they aren't.

I think it would help me at least for this to be made more transparent. I am not sure though what the best way to do this is without creating too much work. Maybe one idea would be to always include your prompt at the top or bottom of the PR?

What do you think @glwagner, @simone-silvestri, @maximilian-gelbrecht, et al?

examples/breeze_over_slab_land.jl
glwagner: The piston velocity is the velocity scale `U` that appears in a bulk formulation like `Jᵛ = U Δq` where `Jᵛ` is moisture flux and `Δq` is the difference between the atmosphere and surface specific humidity.

I believe this flux formulation introduces the notion of a dry surface layer / evaporation front, so that the soil specific humidity is saturated at some depth, separated from the surface by a "dry layer". In this case the surface specific humidity must be computed by balancing fluxes, similar to the way we compute skin temperature through an energy balance.

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
glwagner: I think it is important that this works for wet or frozen land. We can change the name "evaporation".

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
glwagner: This is not part of `SlabLand` though. This is a surface parameterization for computing moisture fluxes between an atmosphere and any land model (including a multilayer land model). This interface parameterization can be used with Terrarium or ClimaLand.

I think we can simplify the model for vapor diffusion. But I don't have an answer for you re: "reference". We don't have references for many things here.

src/Lands/energy_balance/water_coupled_energy.jl
glwagner: Can you provide a more explicit suggestion? I'm not sure what change is being requested.

src/Lands/hydrology/deep_liquid_fluxes.jl
glwagner: In the slab models the fluxes are packaged externally, I think because we have `Field` with nothing location in the vertical or grids that are `Flat` in the vertical. Flux models are more like forcing here.

src/Lands/energy_balance/water_coupled_energy.jl
glwagner: `stateindex` is a generalization of `getindex` that provides a way to either access a static field index with `field[i, j, k]` or to interpolate a FieldTimeSeries in time. I believe it also handles `Number` and possibly other types.

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
glwagner: agree

src/EarthSystemModels/InterfaceComputations/evaporation_front_humidity.jl
glwagner: Hmm yes `\mathcal{S}` should appear in the table. I don't think we should use `S_s` for specific storage. It conflicts with salinity `S`.

glwagner:
> One thought that occurred to me while reviewing this is that it might be helpful to adopt some additional standards regarding AI use in drafting issues and PRs. I found myself several times wondering which of the design choices here were consciously made by @glwagner in the prompt/plan vs. which originated from the LLM. This is not so important in cases where the specific formulations or conventions are common or obvious, but it definitely matters when they aren't.
> 
> I think it would help me at least for this to be made more transparent. I am not sure though what the best way to do this is without creating too much work. Maybe one idea would be to always include your prompt at the top or bottom of the PR?
> 
> What do you think @glwagner, @simone-silvestri, @maximilian-gelbrecht, et al?

The main difficulty is that this was not generated by a single prompt, but quite a few prompts over several days and a lot of iteration.

What is the additional information that you need to write an effective review?

examples/breeze_over_slab_land.jl
bgroenks96: Oh, that's cool. Would be great to have something like that in Terrarium too then. Does this surface water balance in the end also include the precipitation flux? Or is that separate? I guess conceptually in my head it's separate, but it depends a bit on what exactly this dry surface layer represents.

src/EarthSystemModels/InterfaceComputations/evaporation_front_humidity.jl
bgroenks96: L148 of `notation.md` in this PR:

```julia
| ``S_s`` | `specific_storage` | specific storage | Saturated pressure-storage coefficient (m⁻¹) |
```

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
bgroenks96: Ah, ok I missed that. If it's just a general part of the interface flux computation, then that's fine. But I would want to see a reference (or maybe a derivation, if possible) I think to better understand this formulation and what kinds of assumptions it makes. I haven't really seen it before, which is why I am asking.

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
bgroenks96: So in principle this should also handle sublimation?

src/Lands/energy_balance/water_coupled_energy.jl
bgroenks96: For this PR, I think you can ignore this comment because it's out of scope. This style is consistent with the rest of NumericalEarth. It's just a general comment about this pattern to consider. I can also create a separate issue about it.

bgroenks96:
> The main difficulty is that this was not generated by a single prompt, but quite a few prompts over several days and a lot of iteration.

Yes, that's what I kind of expected. That is indeed difficult. But maybe there was an initial design or implementation plan?

> What is the additional information that you need to write an effective review?

Basically I need to know what to pay closer attention to. I trust you @glwagner to make reasonable design decisions, check equations carefully, keep track of references, etc. But I don't trust Claude to do that. So if some equations or formulations were invented by Claude without any clear basis, then I think those parts should be subject to greater scrutiny.

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
glwagner: The idea is that `ℂ.a`, `ℂ.b` can read like `ℂ_a`; eg if `ℂ` means "parameter" and then subscripts or superscripts give meaning. At some point we had ambitions to use `ℂ` to denote parameters, although I am not sure we will continue with that in the future.

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
glwagner: We loosely use comment blocks with three lines of five `#` each to divide code into sections

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
glwagner: Will have to think about this, but one possibility is to try to derive parameters during materialization within `EarthSystemModel`. We already do this to some extent in other cases (for example to extract the ocean reference density)

src/Lands/energy_balance/water_coupled_energy.jl
glwagner: I changed the function signature to bundle together fields into NamedTuples

src/Lands/energy_balance/water_coupled_energy.jl
glwagner: I have defined `_tendency` functions that split this up (although keeping the pattern of having one kernel, for performance reasons)

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
glwagner: I think we should omit "tortuosity" modeling from this PR, and use simple constant vapor diffusion coefficient for this dry layer parameterization, what do you think?

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
glwagner: That would make sense to me. For example, soil with a dry surface layer which is frozen at some depth. I guess in that case the vapor flux is very small, but it would be consistent to treat it in the same way as wet soil with a dry surface layer?

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
glwagner: Here is a reference on different models for the vapor diffusion coefficient in soil:

https://agupubs.onlinelibrary.wiley.com/doi/pdf/10.1029/2011WR011653

bgroenks96: That sounds good to me!

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
glwagner: I believe this is the balances of fluxes, but it isn't obvious (or explained well in the comment) so I will dig into this

examples/breeze_over_slab_land.jl
glwagner: I have to think about this, but I believe it should not depend on precipitation (which is liquid water, whereas this model balances vapor fluxes). That said, if it is precipitating significantly, the surface should become saturated whch would make the surface vapor pressure equal to the saturation specific humidity. This would be good to show in a simple test / example.

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
bgroenks96: When you say "dry surface layer", you mean an infinitesimally small layer of solid material at the surface, yes? And you mean that this dry surface layer would be sitting on top of a frozen layer? I think in that case the humidity flux should be calculated in the same way but with different vapor pressures coefficients depending on the skin temperature (frozen vs. unfrozen). See [this function](https://github.com/NumericalEarth/Terrarium.jl/blob/34fa7f26eb3878095847f13c1e648f2094464601/src/processes/physics_utils.jl#L72) in Terrarium.

I guess it should also work the same way for snow too.

glwagner: For a snow surface there is no dry layer. The dry layer does not have to be infinitesimally thin, necessarily, but there is an assumption that the layer does not have "memory" (which is why an instantaneous flux balance can be used)

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
glwagner: We dug into this properly — references are now in the docstrings, and the tutorial has a new §4.3 "Where this sits in the literature" laying out the full lineage.

The short version: In-soil vapor diffusion contributes meaningfully only inside a thin, very dry surface layer (the coupled liquid–vapor theory goes back to Philip & de Vries 1957; Tang & Riley 2013, HESS, put the crossover at water-filled pore space below ~25%). Everywhere wetter, capillary liquid transport dominates. So Richards-equation models carry only liquid water in the column and reintroduce the dry-layer vapor limitation as a *surface* condition — a soil resistance, a β factor, or a reduced surface humidity (see the model-concept classification in Vanderborght et al. 2017, WRR).

`DryLayerHumidity` is that surface condition, assembled from standard pieces:

- the dry-surface-layer resistance `r_soil = δ/(D_v τ)` with δ diagnosed from near-surface moisture — Yamanaka et al. (1997, WRR) and Swenson & Lawrence (2014, JGR), which is the CLM5 scheme;
- optional Millington & Quirk (1961) tortuosity (CLM5 uses a texture-dependent variant);
- closure by a vapor-flux *balance* for the surface humidity rather than a prescribed α or β — Ye & Pielke (1993, J. Appl. Meteor., Eq. 12b), who showed the shortcuts fail in complementary ways (α always overestimates evaporation from unsaturated soil; β is fine by day but breaks at night). GEOtop (Endrizzi et al. 2014, GMD) adopts the Ye–Pielke parameterization directly even though it resolves 3-D Richards flow.

On ClimaLand specifically: it does use exactly this construction for bare-soil evaporation — see `soil_conductance` and `dry_soil_layer_thickness` in `src/standalone/Soil/energy_hydrology.jl`: `g_soil = D_vapor/DSL` with the DSL thickness diagnosed from top-layer saturation (Sakaguchi & Zeng 2009 lineage; max depth 15 mm after Shokri & Or 2011), combined in series with the aerodynamic conductance plus a Kelvin-equation humidity factor. The tortuosity search comes up empty only because ClimaLand sets τ = 1 in the evaporation scheme (Millington–Quirk appears in its CO₂ model). Its series formula `q_sfc = (g_liq/g_h · q_sat + q_air)/(1 + g_liq/g_h)` is algebraically the same expression our fixed point converges to — we just take the atmospheric conductance from the similarity-theory iterate instead of a fixed exchange coefficient.

glwagner:
> Basically I need to know what to pay closer attention to. I trust you @glwagner to make reasonable design decisions, check equations carefully, keep track of references, etc. But I don't trust Claude to do that. So if some equations or formulations were invented by Claude without any clear basis, then I think those parts should be subject to greater scrutiny.

The development process is really the outcome of a proposal-review loop, so it isn't really uniformly possible to filter for things that we "can't trust because Claude did it".

I think as a result, we should take as close a look as possible to all parts of the review. If possible, I would approach the review more from a practical perspective: quickly review everything that you _possibly_ can, and for things that can't be reviewed by reading the code (and for which validation isn't provided), propose or ask for empirical validation.

In terms of checking equations, I think we usually take the approach to merge first and validate over a longer period of time. The process of validation will involve a mix of experiments / simulations, plus reading docs and checking math.

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
glwagner: It is the two-conductances-in-series solution of the surface vapor-flux balance — eq. (12b) of Ye & Pielke (1993, J. Appl. Meteor.) with their pore relative humidity h_s = 1, and algebraically the same expression ClimaLand evaluates in `update_q_vap_sfc_at_a_point` with its `g_liq/g_h` ratio. The form looks unfamiliar only because of a numerical trick, so I have now written out the full derivation in the section banner above this function (cd185c0a). Summary:

The dry layer transmits a Fick flux `Jᵉ = Gᵉ(qᵉ − qⁱⁿ)` with `Gᵉ = ρᵃᵗ Dᵛ_eff/δᵛ`, while similarity theory carries vapor away above the interface. Over one Picard iteration we linearize the similarity flux as a conductance law anchored at the previous iterate, `Jᵃ(q) ≈ Gᵃ(q − qᵃᵗ)` with `Gᵃ = Jᵃ/Δq`, `Δq = qⁱⁿ⁻ − qᵃᵗ`, chosen so it reproduces the flux `Jᵃ = −ρᵃᵗ u★ q★` the solver actually computed. The interface stores no vapor, so `Jᵉ = Jᵃ` gives the series solution

`qⁱⁿ = (Gᵉ qᵉ + Gᵃ qᵃᵗ)/(Gᵉ + Gᵃ).`

Substituting `Gᵃ = Jᵃ/Δq` and multiplying through by `Δq` gives the coded form `qⁱⁿ = (Gᵉ qᵉ Δq + Jᵃ qᵃᵗ)/(Gᵉ Δq + Jᵃ)` — that step is purely numerical, removing the 0/0 ambiguity of `Gᵃ` as `Δq → 0` (same trick `SkinHumidity` uses). Limits behave as expected: `δᵛ → 0` gives `qⁱⁿ → qᵉ` (saturated skin), a deep front gives `qⁱⁿ → qᵃᵗ` (no evaporation), and at the Picard fixed point the linearization is exact, so the converged humidity satisfies the true nonlinear balance.

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
bgroenks96: Ok but there is a lot of actual text in many of those comments? Wouldn't it be better for those to be in docstrings?

glwagner: Very well could be, I am not sure which ones you are referring to

bgroenks96: e.g. L225-243 of this file, and L1-32 of `water_coupled_energy.jl`

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
bgroenks96: Thanks (I guess Claude...), that helps, especially the Vanderborght et al. reference. I also managed to find the relevant section of the [CLM docs](https://escomp.github.io/CTSM/release-clm5.0/tech_note/Fluxes/CLM50_Tech_Note_Fluxes.html) which further clarified it.

I guess I am just used to the simpler formulation of the surface humidity flux in terms of a single transfer coefficient, i.e. $r_a = C_h v^{-1}$, that is either prescribed or derived from Monin-Obukhov. At least this is how it's done in JSBACH, CryoGrid, and PALADYN. The additional resistance term accounting for tortuosity (based on soil texture or not) is new to me. I guess it also doesn't really matter in the wet-surface-layer formulation.

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
bgroenks96: Now that I know that this is supposed to be a generic interface rather than part of `SlabLand`, I don't think this parameter should be here. The porosity is a property of the soil model and should remain as such. It will be a derived variable when considering pedotransfer functions and can even be a prognostic variable if considering soil mechanics. As such, I think it should be passed in as a function parameter to the `DryLayerHumidity` methods rather than treated as a parameter here.

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
glwagner: I'm not sure the resistance formulation is consistent with MO theory.

I prepared this comment with the help of Claude but it was not solely due to Claude. I feel weird that you give credit just to Claude here.

bgroenks96: It just read very much like Claude style and not your usual tone, which felt very weird to me.

glwagner: Ok, looking at this paper: https://agupubs.onlinelibrary.wiley.com/doi/epdf/10.1002/2014JD022314

and specifically equation 4, I think that the typical way of formulating this in terms of soil resistances can be consistent with MO theory, provided that $q_{soil}$ is solved for along with the fluxes during the MO solve.

The inconsistency only arises if a "resistance correction" is applied _after_ the solve, which some models do (I believe).

So I think one can think of this `DryLayerPistonVelocity` as a rewriting of the typical resistance formulation. The advantage is that it reveals the role that a vapor diffusion coefficient has to play in the formulation. It also avoids introducing the "resistance" which could be an advantage or disadvantage depending on your perspective. Basically this way of expressing the concept of a prognostic surface specific humidity is more consistent with how we express the concept of skin temperatures for air-ice and air-sea fluxes.

bgroenks96: It always bothered me that the surface humidity fluxes were handled differently than the ground heat flux in many of these models' formulations, so I'm happy to have something more consistent.

examples/breeze_over_slab_land.jl
bgroenks96: "Surface vapor pressure" here means the vapor pressure of the dry layer? If this is determined in part by the soil moisture content of the uppermost soil layer, then I would agree that this should happen automatically.

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
bgroenks96: @glwagner Any thoughts on this?

src/Lands/hydrology/variably_saturated_hydrology.jl
```diff
@@ -0,0 +1,246 @@
...
+
+@inline function liquid_saturation(h, θˡ)
+    FT  = typeof(θˡ)
+    ν   = convert(FT, h.porosity)
+    θʳ  = convert(FT, h.residual_liquid_fraction)
+    Δ   = ν - θʳ
+    return clamp((θˡ - θʳ) / Δ, zero(FT), one(FT))
+end
+
+@inline function pressure_head_diag(h, M, θˡ, 𝒮)
```
xkykai: Should we rename this function? on my first read I am slightly confused

```suggestion
@inline function diagnostic_pressure_head(h, M, θˡ, 𝒮)
```

src/Radiations/apply_air_land_radiative_fluxes.jl
```diff
@@ -42,9 +42,27 @@ function apply_air_land_radiative_fluxes!(coupled_model, land)
...
+    FT = eltype(grid)
+    if hasproperty(fluxes, :surface_energy_flux)
+        land_energy_flux = fluxes.surface_energy_flux
+        flux_sign = -one(FT)
+    elseif hasproperty(fluxes, :net_energy_flux)
+        land_energy_flux = fluxes.net_energy_flux
+        flux_sign = one(FT)
+    else
+        return nothing
+    end
```
xkykai: why don't we unify and standardize the direction convention to make it consistent? that would remove this part right? seems clunky to me right now

src/Lands/energy_balance/water_coupled_energy.jl
```diff
@@ -0,0 +1,200 @@
...
+
+    eb_up    = upwind_deep_internal_energy(Jˡb, Tᵈ_ij, Tij, Tᵣ, cˡ)
+    Jᴱ_adv_b = ifelse(energy.advect_deep_liquid_energy, eb_up * Jˡb, zero(FT))
+
+    es_up    = upwind_surface_internal_energy(Jˡs, Tij, Tˡpij, Tᵣ, cˡ)
+    Jᴱ_adv_s = ifelse(energy.advect_surface_liquid_energy, es_up * Jˡs, zero(FT))
+
+    eR_lat = cˡ * (Tij - Tᵣ) * Rlat
+
+    dEdt = Jᴱ_cond + Jᴱ_adv_b - (Jᴱs + Jᴱ_adv_s) - eR_lat
```
xkykai: This is missing the energy flux from sensible internal energy of evaporating water right? 
```suggestion
    # Evaporating liquid (Jᵛ > 0) leaves at the slab temperature carrying its sensible
    # internal energy cˡ(Tⁱⁿ − Tᵣ); the latent heat is already inside Jᴱs. This term
    # pairs the vapor mass flux in dM/dt and keeps the update invariant to Tᵣ.
    Jᴱ_adv_v = cˡ * (Tij - Tᵣ) * Jᵛ

    dEdt = Jᴱ_cond + Jᴱ_adv_b - (Jᴱs + Jᴱ_adv_s + Jᴱ_adv_v) - eR_lat
```

src/Lands/energy_balance/water_coupled_energy.jl
```diff
@@ -0,0 +1,200 @@
...
+    FT = eltype(grid)
+
+    @inbounds begin
+        Tij   = prognostic.T[i, j, 1]
+        Mij   = prognostic.M[i, j, 1]
+        dMdt  = diagnostics.water_storage_tendency[i, j, 1]
+        Jˡb   = diagnostics.deep_liquid_flux[i, j, 1]
+        Jˡs   = diagnostics.surface_liquid_flux[i, j, 1]
+        Rlat  = diagnostics.subsurface_runoff[i, j, 1]
+        Jᴱs   = fluxes.surface_energy_flux[i, j, 1]
```
xkykai:
```suggestion
        Jᴱs   = fluxes.surface_energy_flux[i, j, 1]
        Jᵛ    = fluxes.vapor_flux[i, j, 1]
```

src/Lands/energy_balance/water_coupled_energy.jl
```diff
@@ -0,0 +1,200 @@
...
+                                          advect_deep_liquid_energy,
+                                          advect_surface_liquid_energy)
+end
+
+# Consumes the signed surface energy flux `Jᴱ_s` (positive upward, written by
+# the interface) plus the optional precipitation temperature for the surface
+# advective term. The legacy `net_energy_flux` is kept as a deprecated fallback
+# when the new interface isn't writing `surface_energy_flux`.
+flux_variables(::WaterCoupledEnergy) =
+    (:surface_energy_flux, :liquid_precipitation_temperature)
```
xkykai:
```suggestion
flux_variables(::WaterCoupledEnergy) =
    (:surface_energy_flux, :vapor_flux, :liquid_precipitation_temperature)
```

src/Lands/energy_balance/water_coupled_energy.jl
```diff
@@ -0,0 +1,200 @@
...
+#####    so adding/removing water *at the slab temperature* leaves `Tˡᵃ`
+#####    unchanged. `dMˡᵃ/dt` is consumed from `land.diagnostics.water_storage_tendency`
+#####    written by the hydrology step.
+#####
+##### Deep liquid advective energy uses upwind values: capillary rise brings in
+##### internal energy at `T = Tᵈᵉᵉᵖ`; drainage exports at `T = Tˡᵃ`. The surface
+##### advective term (precipitation/evaporation) is optional and uses bulk
+##### `Tˡᵃ` as the up-flux proxy and `liquid_precipitation_temperature` as the
+##### down-flux value. (Vapor latent heat is *not* an advective liquid term —
+##### it is already inside `surface_energy_flux` via the interface balance.)
```
xkykai:
```suggestion
##### advective term (precipitation/evaporation) is optional and uses bulk
##### `Tˡᵃ` as the up-flux proxy and `liquid_precipitation_temperature` as the
##### down-flux value. The vapor flux carries sensible internal energy out of the
##### slab as an advective term `cˡ(Tˡᵃ − Tᵣ) Jᵛ` — evaporating liquid leaves at the
##### slab temperature; its *latent* heat is separate, already inside
##### `surface_energy_flux` via the interface balance.
```

test/test_water_coupled_energy.jl
```diff
@@ -0,0 +1,103 @@
...
+        for _ in 1:50
+            time_step!(land, 1000.0)
+        end
+
+        T_actual = @allowscalar interior(land.temperature)[1, 1, 1]
+        M_actual = @allowscalar interior(land.water_storage)[1, 1, 1]
+        @test isapprox(T_actual, 290.0; atol = 1e-10)
+        @test M_actual < 200  # water did leave
+    end
+end
```
xkykai, line 101:
```julia
end

@testset "WaterCoupledEnergy evaporation is reference-temperature invariant" begin
    for arch in test_architectures
        grid = RectilinearGrid(arch; size = 1, x = (0, 1), y = (0, 1),
                               z = (-1, 0), topology = (Flat, Flat, Bounded))

        hydrology = VariablySaturatedHydrology(eltype(grid);
            slab_depth = 1.0, porosity = 0.4, storage_height = 1000,
            critical_saturation = 0.5,
            retention_curve = VanGenuchtenRetention(α = 1.0, n = 2.0),
            hydraulic_conductivity = VanGenuchtenConductivity(K_saturated = 1e-6, n = 2.0),
            deep_liquid_flux = NoDeepLiquidFlux(), runoff = NoRunoff())

        # Identical forcing, two arbitrary internal-energy references. The physical
        # temperature evolution cannot depend on Tᵣ; it does so iff the vapor mass
        # flux is unpaired in the energy budget.
        function evaporate_with_reference(Tᵣ)
            energy = WaterCoupledEnergy(eltype(grid);
                dry_heat_capacity = 1e6, liquid_heat_capacity = 4186,
                reference_temperature = Tᵣ,
                deep_temperature = 290.0, deep_conductance = 0.0,
                advect_deep_liquid_energy = true,
                advect_surface_liquid_energy = false)
            land = SlabLand(grid; hydrology, energy)
            set!(land; T = 300.0, M = 200.0)
            fill!(land.fluxes.surface_energy_flux, 50.0)  # latent cooling, positive upward
            fill!(land.fluxes.vapor_flux, 2e-5)           # evaporation, positive upward
            fill!(land.fluxes.liquid_precipitation_flux, 0)
            for _ in 1:100
                time_step!(land, 100.0)
            end
            return @allowscalar interior(land.temperature)[1, 1, 1]
        end

        @test isapprox(evaporate_with_reference(0.0),
                       evaporate_with_reference(273.15); atol = 1e-8)
    end
end
```

src/EarthSystemModels/InterfaceComputations/interface_states.jl
```diff
@@ -62,6 +62,15 @@ ImpureSaturationSpecificHumidity(phase) = ImpureSaturationSpecificHumidity(phase
...
     εᵈᵛ⁻¹ = 1 / AtmosphericThermodynamics.Parameters.Rv_over_Rd(ℂᵃᵗ)
+
+    # Guard against an unphysically warm interface temperature. The denominator
+    # `p - (1 - εᵈᵛ⁻¹) pᵛ⁺` vanishes and then turns negative once pᵛ⁺ exceeds
+    # p / (1 - εᵈᵛ⁻¹) ≈ 2.6 p, returning a *negative* specific humidity that drives
+    # a runaway spurious-condensation instability in the coupled fluxes. This only
+    # happens for super-boiling temperatures (e.g. an ocean T mistakenly supplied in
+    # Kelvin, read as °C and converted to ~566 K); in the physical regime pᵛ⁺ ≪ p and
+    # the cap is inert. Capping pᵛ⁺ below p keeps qₛ ∈ [0, 1) and saturating.
+    pᵛ⁺   = min(pᵛ⁺, convert(CT, 0.999) * p)
```
xkykai, line 70: shouldn't we throw an error in such cases rather than let it run silently?

src/Lands/hydrology/variably_saturated_hydrology.jl
```diff
@@ -0,0 +1,246 @@
...
+    hˢˢ = convert(FT, h.storage_height)
+    Mˡᵃ⁺ = ρˡ * ν * hˡᵃ
+    # Unsaturated branch: Π = Π_m(𝒮). Saturated branch: Π = (M − M⁺) hˢˢ/(ρˡ hˡᵃ).
+    return ifelse(M < Mˡᵃ⁺,
+                  pressure_head(h.retention_curve, 𝒮),
+                  (M - Mˡᵃ⁺) * hˢˢ / (ρˡ * hˡᵃ))
+end
+
+@inline moisture_availability(h, 𝒮) =
+    clamp(𝒮 / convert(typeof(𝒮), h.critical_saturation), zero(𝒮), one(𝒮))
```
xkykai: is `moisture_availability` going to be used anywhere? right now it is unused

```suggestion

```

src/Radiations/apply_air_land_radiative_fluxes.jl
glwagner: Fluxes should be positive upwards right?

src/Lands/energy_balance/water_coupled_energy.jl
glwagner: Do you mean the latent heat flux?

src/EarthSystemModels/InterfaceComputations/interface_states.jl
glwagner: This is a hot inner loop though, so we can't. But we could return NaN (also note to self, the notation in the comment for saturation specific humidity is wrong)

glwagner:
@xkykai I can respond to comments and resolve conflicts if you like

xkykai:
> @xkykai I can respond to comments and resolve conflicts if you like

sure, thanks a lot! let me know when you want me to take over/continue the fixes

src/Lands/energy_balance/water_coupled_energy.jl
xkykai: latent is in Jᴱs right? I mean the sensible internal energy that's lost when the water is evaporated away. This keeps the slab temperature invariant when water at the slab temperature evaporates, where the cooling comes purely from the latent heat in Jᴱs.

src/EarthSystemModels/InterfaceComputations/dry_layer_humidity.jl
xkykai: why don't we remove `porosity` here and access it through `Ψₛ.hydrology.porosity` in `compute_interface_humidity` which is integrated into the soil model instead

src/EarthSystemModels/InterfaceComputations/interface_states.jl
xkykai: Should we throw `NaN` then? so it doesn't fail silently?

src/Lands/energy_balance/water_coupled_energy.jl
xkykai: Here's a demonstration on why we need the additional term which represents the sensible internal energy of evaporated water:

## Setup

The slab stores internal energy in two reservoirs — the dry matrix and the liquid water:

$$
E = C_{\text{dry}} (T - T_r) + c^l M (T - T_r), \qquad C = C_{\text{dry}} + c^l M ,
$$

so the temperature is recovered as $T = T_r + E / C$.

## What evaporation does to the slab

Evaporating a mass $\delta M$ removes energy from the slab through **two** channels:

1. **Latent heat**, $\mathcal{L} \delta M$, drawn off to drive the phase change. This is already accounted for inside the surface energy flux $J^E_s$ (the $\mathcal{Q}^v$ term).
2. **Sensible internal energy**, $c^l (T - T_r) \delta M$. That $\delta M$ was *liquid at temperature $T$*; the energy it was holding leaves the control volume with the vapor.

Correspondingly the energy budget must change by

$$
\delta E = - \underbrace{\mathcal{L} \delta M}_{\text{latent}} - \underbrace{c^l (T - T_r) \delta M}_{\text{sensible}} ,
\qquad
C \to C_{\text{new}} = C - c^l \delta M .
$$

The heat-capacity update already drops the water from the **denominator** of $T = T_r + E/C$. If we do not also drop the sensible piece from the **numerator** $E$, the same energy divides over a smaller heat capacity and the slab spuriously warms.

## Pure-water slab example

With $C_{\text{dry}} = 0$, so $C = c^l M$, the spurious term collapses to

$$
\Delta T_{\text{spurious}} = (T - T_r) \frac{\delta M}{M - \delta M} .
$$

which means evaporating 1% of the water heats the slab spuriously by ≈ 1% of $(T - T_r)$.

examples/breeze_over_slab_land.jl
```diff
@@ -77,33 +77,66 @@ land_grid = RectilinearGrid(arch;
...
+# bulk soil drainage), and an infiltration-capacity runoff closure for any
+# precipitation that exceeds the soil capacity.
+
+hydrology = VariablySaturatedHydrology(eltype(land_grid);
+    slab_depth = 1.0,
+    porosity = 0.4,
+    residual_liquid_fraction = 0.05,
+    storage_height = 1000,
+    retention_curve = VanGenuchtenRetention(α = 1.0, n = 2.0),
+    hydraulic_conductivity = VanGenuchtenConductivity(K_saturated = 1e-7, n = 2.0),
```
glwagner, line 94: Can we find more descriptive names for these properties? "Van Genuchten" doesn't ring a bell for me, although it could be good to mention the citation in the docstring.

examples/breeze_over_slab_land.jl
```diff
@@ -224,18 +257,31 @@ set_to_mean!(reference_state, atmos.model, rescale_densities = true)
...
+# limits evaporation self-consistently.
+interface_specific_humidity = DryLayerHumidity(;
+    dry_layer_depth = StorageBasedDryLayerDepth(
+        maximum_dry_layer_depth = 0.05,
+        dry_layer_onset_saturation = 0.5,
+        dry_layer_exponent = 2),
+    vapor_exchange = DryLayerVaporPistonVelocity(
+        minimum_dry_layer_depth = 1e-4,
+        molecular_diffusivity = 2.5e-5,
+        tortuosity_model = MillingtonQuirk()),
```
glwagner: Perhaps a more descriptive name for this kind of tortuosity would also make sense.

examples/breeze_over_slab_land.jl
xkykai: > Can we find more descriptive names for these properties? "Van Genuchten" doesn't ring a bell for me

"van Genuchten" refers to the retention curve and the hydraulic conductivity, which seems to be commonplace in land models (eg. on page 5 of the ClimaLand paper they refer to them as "ClimaLand supports the van Genuchten (1980) and Brooks and Corey (1964) parameterizations for hydraulic conductivity K and matric potential ψm as functions of volumetric liquid water content in the vadose zone..."). So I think the name is ok. `retention_curve` and `hydraulic_conductivity` seems descriptive enough. 

> it could be good to mention the citation in the docstring

agree, I will add it

xkykai: but we could also say something like `SigmoidalRetention` for the `VanGenuchtenRetention`. But I don't think that is descriptive enough for to distinguish itself with other models. On the other hand we could have  `PoreBundleConductivity` for `VanGenuchtenConductivity` if we were to convey the spirit of the models. Let me know what you think of these.

examples/breeze_over_slab_land.jl
xkykai: perhaps something like `AirConnectivityTortuosity`? But I feel that `MillingtonQuirk` might be more descriptive to some people (not all though), so depending on the philosophy of the naming scheme happy to go either way.

src/Lands/energy_balance/water_coupled_energy.jl
glwagner: I think you're right we need to include the energy flux from evaporation explicitly. Internal phase changes have no effect on the total energy of the slab, but external changes do. (I guess we do not have internal phase changes yet, but we will if we include ice / ice fraction).

src/EarthSystemModels/InterfaceComputations/interface_states.jl
glwagner: Is this calculation embedded in the solve? If it is inside a solver, then we can allow unphysical temperatures to temprarily persist while we are iterating to find the solution. If this is outside the solve, NaN makes sense

examples/breeze_over_slab_land.jl
bgroenks96: I can confirm that van Genuchten is quite ubiquitous in soil hydrological modeling. Almost everyone will understand what you're talking about (although clarifying between the VG retention vs. conductivity curves would be good). I am usually not (anymore) a fan of naming code components after author names, but this is a clear exception. Same with Brooks-Corey.

examples/breeze_over_slab_land.jl
glwagner: for most other things (eg `ValueBoundaryCondition` vs `Dirichlet`) we are going with descriptive names.

Why "air connectivity"?

xkykai: > Why "air connectivity"?

Because the argument I think goes like the tortuosity depends on how many connected paths there are for vapor travelling through the soil. So as the soil moisture increases, the tortuosity rapidly decreases.

or it could be `MoistureDependentTortuosity`?

xkykai:
@glwagner I believe this is ready!

I was wondering if you intend to publish https://github.com/NumericalEarth/NumericalEarth.jl/blob/7d4b04a0a62f8e9c1e341c242780f23e196b8b7e/docs/src/land/evaporation_front_slab_land.md 

and 

https://github.com/NumericalEarth/NumericalEarth.jl/blob/7d4b04a0a62f8e9c1e341c242780f23e196b8b7e/docs/src/land/follow_up_roadmap.md

? I'm not sure so I didn't touch them

examples/breeze_over_slab_land.jl
glwagner: "Surface vapor pressure" refers to the vapor pressure of the interface itself. This (or rather the interface specific humidity) is the value which is used in bulk formula / MOST to compute surface fluxes. It actually differs from the specific humidity of the dry layer. In this model the "dry layer specific humidity" is not constant. It varies linearly, from the bulk specific humidity (the specific humidity of the uppermost layer) to the surface.

glwagner: This is the same kind of model we would invoke for a skin temperature. The bulk temperature is a prognostic of the component model (eg hydrostatic ocean, slab ocean, slab land, terrarium land). The skin temperature is the temperature of the interface, which enters into bulk formula. A typical skin temperature model models the interface energy balance; then the skin temperature is the solution of a balance of fluxes, which is computed as part of the MOST solve. The underlying model for the energy balance prescribes the structure of the temperature, but usually you would assume it varies linearly from the bulk temperature below to the skin temperature.

examples/breeze_over_slab_land.jl
glwagner: There's still the problem that I have no idea what it means

examples/breeze_over_slab_land.jl
bgroenks96: Got it, thanks.

examples/breeze_over_slab_land.jl
glwagner: Basically the goal of coming up with descriptive names is to increase the accessibility of the code beyond people who are already indoctrinated in the field. So while true that experts in soil hydrology will have no problem with "van Genuchten", its also the case that experts need less guidance anyways. So the probably more important consideration is how to provide something for people/students who are not soil hydrology experts here.

glwagner: Our idea is to use "SigmoidBasedRetention", and then we will explain that this conductivity model is based on a sigmoidal relationship between the pressure head and water content, where the connection between those is described by Mualem (1976) and Van Genutchen (1980).

examples/breeze_over_slab_land.jl
glwagner: maybe `PowerLawTortuosity`! and change the kwarg to `tortuosity` from `tortuosity_model`

glwagner:
> @glwagner I believe this is ready!
> 
> I was wondering if you intend to publish https://github.com/NumericalEarth/NumericalEarth.jl/blob/7d4b04a0a62f8e9c1e341c242780f23e196b8b7e/docs/src/land/evaporation_front_slab_land.md
> 
> and
> 
> https://github.com/NumericalEarth/NumericalEarth.jl/blob/7d4b04a0a62f8e9c1e341c242780f23e196b8b7e/docs/src/land/follow_up_roadmap.md
> 
> ? I'm not sure so I didn't touch them

lets delete those?

examples/breeze_over_slab_land.jl
bgroenks96: The problem with `SigmoidBasedRetention` is that it provides strictly speaking *less* information. Assuming you mean the general form of the sigmoid, i.e.
```math
y = a\sigma(f(x)) + b
```
then there are multiple soil water retention curves that could be considered "sigmoid-based" (the Campbell and Haverkamp models come to mind). The reason why many of these empirical curves are either S-shaped or power laws (e.g. Brooks-Corey) is because they are related to the CDF of the pore size distribution of the soil; smaller pores are harder to dry than larger pores due to stronger capillary effects. If by "sigmoid-based" you mean the logistic function, then it would be wrong because VG is not identical to the logistic function. 

I also don't really see how `SigmoidBasedRetention` is that much more "beginner friendly". It still assumes that the reader is familiar with what "retention" means in this context, which means they need to know what a soil water retention curve is, and if they know that, they almost certainly know van Genuchten and Brooks-Corey. It's literally on the Wikipedia page.

If you still insist on `SigmoidBasedRetention`, then the implementation should be generalized to cover all sigmoidal retention curves, not just VG. Otherwise it's misleading.

glwagner: `SigmoidRetention` means that the retention curve varies smoothly between two values, that's it. I don't know what x, y, a, b, and sigma are in your formula above. What specifically do you propose as a generalization? What wikipedia page?

glwagner: Our objective is not to provide more or less information. I don't think we need to provide an ultra generic implementation, the idea is just to use a name that is at least minimally descriptive in nature. As an example, we have a type in Oceananigans called "ScalarDiffusivity". However, it is not necessarily possible to implement any mathematically viable scalar diffusivity. Instead, it is simply the case that using this type does imply that the diffusivity is a scalar. Or perhaps even more generally, we use the term "hydrostatic free surface model". However, this does not mean that all possible hydrostatic models are supported. We only support height and z* coordinates, for example.

For `SigmoidRetention`, it is nice to support a family of curves that are sigmoids, but I do not think we need to support all sigmoids.

xkykai: Let's go with `SigmoidRetention` and `SigmoidRetentionConductivity`?

bgroenks96: > SigmoidRetention means that the retention curve varies smoothly between two values, that's it. 

Ok, that's even more general then. By that definition, I think almost all SWRCs would qualify..?

> What specifically do you propose as a generalization?

I don't. I prefer to stick with `VanGenuchtenRetention`.

> What wikipedia page?

[This one](https://en.wikipedia.org/wiki/Water_retention_curve).

>  it is nice to support a family of curves that are sigmoids, but I do not think we need to support all sigmoids.

Ok, I shouldn't have said *all*. My point was just that, if it has a more generic name, it should actually *be* more generic. It should not be called `SigmoidRetention` and then actually just be *only* the van Genuchten curve.

bgroenks96: > Let's go with SigmoidRetention and SigmoidRetentionConductivity

I strongly object to that for the current implementation that is written in `hydraulic_functions.jl`. But if I am outvoted, then so be it.

xkykai: Let's use `VanGenuchtenConductivity` and `VanGenuchtenRetention` for now!

xkykai (approved):
I'll approve, then check through and polish

examples/breeze_over_slab_land.jl
bgroenks96: btw, FWIW @xkykai, you could also just use/alias the existing names and functionss in [FreezeCurves jl](https//dithub.com/CryoGrid/FreezeCurves.jl).

xkykai: Can do! But I think I'll work it in in a separate PR

PR #327, opened 2026-06-04
Fix breeze over land example

simone-silvestri:
cc @navidcy


navidcy (approved):
Hm... Again this hardcoded Float64 in the Clock, right?

Should we change the AtmosphereModel constructor in Breeze to have default `clock = clock = Clock{eltype(grid)}(grid)`?

Or change the Oceananigans method

```Julia
Clock(grid::AbstractGrid{FT}) where {FT} = Clock{Float64}(; time=0, kernel_time_type=FT)
```
?

PR #332, opened 2026-06-06
Add SoilGrids 2.0 dataset to DataWrangling

### LINKED ISSUE #328: Add SoilGridsV2 global soil properties dataset
As a first step for facilitating global land simulations, we should add a global dataset for soil properties.

So far as I am aware, the state-of-the-art here is still [SoilGrids v2](https://www.soilgrids.org/) which has a native resolution of 250 m over land. I have preprocessed the data from its native GeoTIFF format into ZARR at 1 km resolution and NetCDF at ~10 km resolution for the following variables at all five vertical layers:
- Sand content (`sand`)
- Silt content (`silt`)
- Clay content (`clay`)
- Bulk density (`bdod`)
- Organic carbon density (`ocd`)
- Soil organic carbon (`soc`)

I will add a new `Metadata` implementation for this dataset that downloads the NetCDF file from a file sharing server provided by TU Munich.

One slight wrinkle is that the dataset currently also includes layers for each statistic: i.e. the mean, 5%, 50%, and 95% quantiles. I figured we could just add this as an option when instantiating the `Field` from the `Metadatum`.

bgroenks96:
This PR adds the SoilGrids 2.0 dataset of [Poggio et al. 2021](https://soil.copernicus.org/articles/7/217/2021/) to `DataWrangling` following the existing conventions.

One slight deviation is the custom override of the `Field` constructor to allow for the late specification of which layer to use: mean or 5th, 50th, and 95th percentiles. We could alternatively make this a property of the dataset `struct` rather than a keyword argument. That would probably be a bit more user friendly.

The dataset used here is a postprocessed version of the original data. The postprocessing code is in python, but I can also include it here if needed.

### Additional changes
- ~~Adds an optional position argument to `retrieve_data` that lets the caller specify the index for the trailing dimension.~~
- Improves missing data handling (needed for SoilGrids dataset)

Resolves #328


src/DataWrangling/SoilGrids/SoilGridsV2.jl
```diff
@@ -0,0 +1,92 @@
...
+    Downloads.download(metadatum)
+
+    grid = native_grid(metadatum, arch; halo)
+    LX, LY, LZ = location(metadatum)
+    field = Field{LX, LY, LZ}(grid)
+
+    # Retrieve data from file according to metadatum type
+    data = retrieve_data(metadatum, Int(stat) + 1)
+    # Replace fill value
+    data = replace(data, -32768 => 0)
```
bgroenks96: @glwagner I am not quite sure what to do here. The dataset is saved in `int16`  format for compression with `-32768` as the fill value for non-land grid cells. All of the variables have precision of three significant digits; for the sand, silt, and clay contents, we should apply a unit conversion by dividing by 1000. Not sure what the best place is to do this here.

src/DataWrangling/SoilGrids/SoilGridsV2.jl
```diff
@@ -0,0 +1,92 @@
...
+download_SoilGridsV2_cache::String = ""
+function __init__()
+    return global download_SoilGridsV2_cache = @get_scratch!("SoilGridsV2")
+end
+
+struct SoilGridsV2 <: AbstractStaticDataset end
+
+# Variable name mappings from NumericalEarth names to SoilGridsV2 variable names
+SoilGridsV2_dataset_variable_names = Dict(
+    :sand_content => "sand",
```
bgroenks96: TODO: Add these variables to the notation doc page

simone-silvestri:
> One slight deviation is the custom override of the Field constructor to allow for the late specification of which layer to use: mean or 5th, 50th, and 95th percentiles.

I guess a property of the dataset would be preferred. It is better to extend `retrieve_data` rather than the `Field` constructor if possible

src/DataWrangling/SoilGrids/SoilGridsV2.jl
simone-silvestri: You can extend `conversion_units(::Metadatum)` to perform the conversion.

simone-silvestri: For the missing value, I think we need a `missing_value(::Metadatum)` function which defaults to `missing`. This will be used in `nan_convert_missing` (need to change that function too) that is called in `set_region_data!`

src/DataWrangling/DataWrangling.jl
```diff
@@ -267,38 +271,38 @@ for the microphysics symbols).
...
-    :air_temperature                      => :T,
-    :salinity                             => :S,
-    :u_velocity                           => :u,
-    :v_velocity                           => :v,
-    :eastward_velocity                    => :u,
-    :northward_velocity                   => :v,
-    :eastward_wind                        => :u,
-    :northward_wind                       => :v,
-    :sea_level_pressure                   => :p,
+    :temperature => :T,
```
simone-silvestri: can we keep the current formatting?

bgroenks96: Do you have a an automatic formatting configuration? I think my editor is just applying runic by default.

simone-silvestri: no, we format by hand mostly :) we were talking about adding a formatter but haven't gotten to a decision yet.

simone-silvestri: I personally really like having aligned code, for example at `=` signs and operations across lines. It makes it easier to modify multiple lines in cursor / vscode.

src/DataWrangling/DataWrangling.jl
```diff
@@ -422,4 +423,4 @@ function Downloads.download(metadata::Metadata)
     error("No download method for $metadata is available (is the backend package loaded?)")
 end
 
-end # module
+end # module
```
simone-silvestri:
```suggestion
end # module

```

src/DataWrangling/metadata_field.jl
```diff
@@ -471,4 +471,4 @@ function Oceananigans.Fields.Field(mset::MetadataSet, arch=CPU(); kw...)
     end
     names = getfield(mset, :names)
     return NamedTuple{names}(map(n -> Field(mset[n], arch; kw...), names))
-end
+end
```
simone-silvestri:
```suggestion
end

```

src/DataWrangling/SoilGrids/SoilGridsV2.jl
bgroenks96: @simone-silvestri I don't think I understand how the unit conversions work. Do the types correspond to the source units of the data? and then `convert_units` converts them to whatever is the standard units for `NumericalEarth`?

src/DataWrangling/DataWrangling.jl
bgroenks96: Fixed in [d0b866c](https://github.com/NumericalEarth/NumericalEarth.jl/pull/332/commits/d0b866cc592ba885037e23cd4f15d8467121bb76)

src/DataWrangling/SoilGrids/SoilGridsV2.jl
bgroenks96: Done in [05b2d43](https://github.com/NumericalEarth/NumericalEarth.jl/pull/332/commits/05b2d4306fe59aa4b00f01c1e382d8fc07d43e94)

src/DataWrangling/SoilGrids/SoilGridsV2.jl
bgroenks96: Unit conversions added in [4a1def2](https://github.com/NumericalEarth/NumericalEarth.jl/pull/332/commits/4a1def2767d26b82f80d2d1c27e9eca2b048465e)

src/DataWrangling/DataWrangling.jl
bgroenks96: Fixed in [d0b866c](https://github.com/NumericalEarth/NumericalEarth.jl/pull/332/commits/d0b866cc592ba885037e23cd4f15d8467121bb76).

bgroenks96:
@simone-silvestri The SoilGrids statistic is now configured via the struct in [8ae24f1](https://github.com/NumericalEarth/NumericalEarth.jl/pull/332/commits/8ae24f18ffb6b6e5540ede5b6e1e4f761b8dab06) and we override `retrieve_data` instead of `Field`.

simone-silvestri (approved):
Looks good. I would add a small unit test to make sure the metadata constructor and the associated methods work as intended. then LGTM

src/DataWrangling/SoilGrids/SoilGridsV2.jl
```diff
@@ -0,0 +1,103 @@
...
+const SoilGridsV2_url = "https://syncandshare.lrz.de/dl/fiVMyHskjL3FNbceuUFJev/soilgrids2_clenshaw989_10km.nc"
+
+# Dataset methods
+DataWrangling.available_variables(::SoilGridsV2) = SoilGridsV2_dataset_variable_names
+DataWrangling.default_download_directory(::SoilGridsV2) = download_SoilGridsV2_cache
+DataWrangling.reversed_vertical_axis(::SoilGridsV2) = true
+DataWrangling.reversed_latitude_axis(::SoilGridsV2) = true
+DataWrangling.longitude_interfaces(::SoilGridsV2) = (-180, 180)
+DataWrangling.latitude_interfaces(::SoilGridsV2) = (-90, 90)
+DataWrangling.z_interfaces(::SoilGridsV2) = [-200.0, -100.0, -60.0, -30.0, -15.0, -5.0, 0.0]
```
glwagner: is float64 necessary here?

src/DataWrangling/SoilGrids/SoilGridsV2.jl
```diff
@@ -0,0 +1,103 @@
...
+using Scratch: Scratch, @get_scratch!
+
+using ..DataWrangling: DataWrangling, DownloadProgress, AbstractStaticDataset, Metadatum,
+    metadata_path, metadata_url, native_grid, retrieve_data, set_metadata_field!
+
+import Oceananigans
+
+@enum SoilGridsStat mean Q5 Q50 Q95
+
+download_SoilGridsV2_cache::String = ""
```
glwagner:
```suggestion
download_SoilGrids2_cache::String = ""
```

It doesn't look like the paper uses "V2", instead it is written "SoilGrids 2.0"

src/DataWrangling/SoilGrids/SoilGridsV2.jl
```diff
@@ -0,0 +1,103 @@
...
+using Oceananigans: CPU, Field, Center, location, fill_halo_regions!
+using Oceananigans.DistributedComputations: @root
+using Scratch: Scratch, @get_scratch!
+
+using ..DataWrangling: DataWrangling, DownloadProgress, AbstractStaticDataset, Metadatum,
+    metadata_path, metadata_url, native_grid, retrieve_data, set_metadata_field!
+
+import Oceananigans
+
+@enum SoilGridsStat mean Q5 Q50 Q95
```
glwagner: what does "Stat" stand for? It is "statistic"? Perhaps we can use the full word?

src/DataWrangling/SoilGrids/SoilGridsV2.jl
```diff
@@ -0,0 +1,103 @@
...
+
+download_SoilGridsV2_cache::String = ""
+function __init__()
+    return global download_SoilGridsV2_cache = @get_scratch!("SoilGridsV2")
+end
+
+@kwdef struct SoilGridsV2 <: AbstractStaticDataset
+    stat::SoilGridsStat = mean
```
glwagner: what does this indicate?

src/DataWrangling/SoilGrids/SoilGridsV2.jl
bgroenks96: I suppose not. Will the type be autoconverted?

src/DataWrangling/SoilGrids/SoilGridsV2.jl
bgroenks96: Fixed in [c16c240](https://github.com/NumericalEarth/NumericalEarth.jl/pull/332/commits/c16c2401e79dbc99ddbdcfa3255f81d0ada26e85)

src/DataWrangling/SoilGrids/SoilGridsV2.jl
bgroenks96: Fixed in [2a64be4](https://github.com/NumericalEarth/NumericalEarth.jl/pull/332/commits/2a64be41e57c19e1f90fedb241d6742858d6a49e)

src/DataWrangling/SoilGrids/SoilGridsV2.jl
bgroenks96: Does [2a64be4](https://github.com/NumericalEarth/NumericalEarth.jl/pull/332/commits/2a64be41e57c19e1f90fedb241d6742858d6a49e) clarify this sufficiently?

src/DataWrangling/SoilGrids/SoilGridsV2.jl
bgroenks96: Changed to integers in [2a64be4](https://github.com/NumericalEarth/NumericalEarth.jl/pull/332/commits/2a64be41e57c19e1f90fedb241d6742858d6a49e)

bgroenks96:
@simone-silvestri I need to deal with the missing values, but I am not sure that I understand `nan_convert_missing`. It takes as an argument the actual data value, yes? So we would need to check whether that value equals `missing_value(metadatum)`?

Is the point of this dispatch
```julia
@inline nan_convert_missing(FT, d::Number) = convert(FT, d)
```
to handle non-missing data? Or missing values defined with other numbers?

simone-silvestri:
the `nan_convert_missing` is there to genuinely convert missing values from the dataset in NaNs that we know how to handle.
We can rework this function to accept a `missing_value` and convert the actual missing values, probably something like
```julia
@inline nan_convert_missing(FT, d::Number, missing_value) = ifelse(d == missing_value, NaN, convert(FT, d))
```
the method you cite is there to handle non-missing data indeed

src/DataWrangling/metadata.jl
```diff
@@ -757,9 +761,21 @@ struct MillimetersPerHour end     # liquid precipitation rate in mm/hr → kg/m
...
+These units will be used to apply automatic conversions to standard units for `NumericalEarth`.
+"""
 conversion_units(metadatum) = nothing
 
+"""
+    missing_value(metadatum)
+
+Return the value used by the underlying dataset to represent missing data. Defaults to `missing`.
+"""
+missing_value(metadatum) = missing
```
bgroenks96, line 777: @simone-silvestri Added `missing_value` as you suggested

src/DataWrangling/set_region_data.jl
```diff
@@ -206,12 +206,13 @@ function set_region_data!(target::Field, data, λc, φc, metadata;
...
-    arch   = architecture(grid)
-    data   = on_architecture(arch, data)
-    launch!(arch, grid, :xyz, _set_region_kernel!, interior(target), data, region, mangling, conversion, FT)
+    region      = region_info(metadata.region, target, λc, φc)
+    FT          = eltype(target)
+    grid        = target.grid
+    arch        = architecture(grid)
+    data        = on_architecture(arch, data)
+    missing_val = missing_value(metadata)
+    launch!(arch, grid, :xyz, _set_region_kernel!, interior(target), data, region, mangling, conversion, missing_val, FT)
```
bgroenks96, line 218: @simone-silvestri missing value is passed through to `_set_region_kernel!` and upstack methods

test/test_soilgrids.jl
```diff
@@ -0,0 +1,27 @@
+using NumericalEarth
+using NumericalEarth.DataWrangling
+using NumericalEarth.SoilGrids
+using Oceananigans
+
+import Downloads
+
+@testset "Metadata tests for SoilGrids" begin
```
bgroenks96, line 8: @simone-silvestri Is this good enough for now? Or should I add some more tests?

src/DataWrangling/metadata_field.jl
```diff
@@ -371,8 +371,8 @@ function centers_to_interfaces(z_centers)
     return z_faces
 end
 
-@inline nan_convert_missing(FT, ::Missing) = convert(FT, NaN)
-@inline nan_convert_missing(FT, d::Number) = convert(FT, d)
+# Convert missing values to NaN
+@inline nan_convert_missing(FT, x, missing_val) = ifelse(x == missing_val, convert(FT, NaN), convert(FT, x))
```
bgroenks96: @simone-silvestri Does this look good to you?

test/test_soilgrids.jl
simone-silvestri: looks good!

src/DataWrangling/metadata_field.jl
simone-silvestri: I think we need `===`:

```julia
julia> missing == missing
missing

julia> missing === missing
true
```

simone-silvestri (approved):
a part for the correction in `nan_convert_missing` (`===` instead of `==`), looks good to merge for me.

src/DataWrangling/metadata_field.jl
bgroenks96: Hmm, that works for `missing` but

```julia
julia> -32768 === -32768.0
false
```

but the updated code does
```julia
@inline nan_convert_missing(FT, x, ::Missing) = ifelse(ismissing(x), convert(FT, NaN), convert(FT, x))
@inline nan_convert_missing(FT, x, missing_val) = ifelse(ismissing(x) || x == missing_val, convert(FT, NaN), convert(FT, x))
```
which I think should solve it.

src/DataWrangling/SoilGrids/SoilGridsV2.jl
glwagner: It could be worth a docstring with example (which maybe I missed). I still don't completely understand, although I can try to guess. I think the "statistic" refers to a spatial average? So "mean" is the mean over a grid cell, whereas "Q5" (50, 95) are the 5, 50, 95% quantiles of the quantity?

Then the user interface being proposed looks something like

```julia
mean_dataset = SoilGrids2(mean)
mean_meta = Metadata(:temperature, mean_dataset)

q5_dataset = SoilGrids(Q5)
q5_meta = Metadata(:temperature, q5_dataset)
```

My main concern is that this interface is a bit surprising. For example, for the rest of the data we have, the "dataset" refers to the entire corpus of data, and the "variable name" refers to a specific piece of that corpus. Thus within this pattern, one might expect something like

```julia
q5_meta = Metadata(:q5_temperature, SoilGrids2())
```

I understand why you would choose the former versus this latter, though (eg it is less code).

glwagner: It also isn't clear to me just for lack of Julia knowledge whether the struct `SoilGrids2` is concretely typed? Since `SoildGridsStatistic` is not a concrete type?

bgroenks96:
@simone-silvestri I guess the failing CI is not relevant to this PR, right?

src/DataWrangling/SoilGrids/SoilGridsV2.jl
bgroenks96: > I still don't completely understand, although I can try to guess. I think the "statistic" refers to a spatial average? So "mean" is the mean over a grid cell, whereas "Q5" (50, 95) are the 5, 50, 95% quantiles of the quantity?

Not quite. SoilGrids provides probabilistic predictions of soil properties, so the quantiles capture uncertainty in the predicted data values for each grid cell (this neglects spatial autocorrelation, which is a limitation, but I digress). The main reason a user might choose something other than the mean is if they want to investigate the sensitivity of the simulations to this uncertainty.

> My main concern is that this interface is a bit surprising. For example, for the rest of the data we have, the "dataset" refers to the entire corpus of data, and the "variable name" refers to a specific piece of that corpus

I see your point, but I don't think I would agree that these statistics represent new variables. It's really just several realizations of the same underlying dataset. There is no use case that I can think of where we would actually refer to `q5_sand_content` in the numerical model.

> It also isn't clear to me just for lack of Julia knowledge whether the struct SoilGrids2 is concretely typed? Since SoildGridsStatistic is not a concrete type?

Enums are concrete types.

bgroenks96: I will add some more detail to the docstrings.

simone-silvestri:
they do seem related to this PR (at least some of them):
```julia
ERROR: LoadError: MethodError: no method matching ifelse(::Missing, ::Float32, ::Float32)
The function `ifelse` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  ifelse(::Bool, ::Any, ::Any)
   @ Base essentials.jl:799
  ifelse(::Static.False, ::Any, ::Any)
   @ Static /usr/local/share/julia/packages/Static/TjBVO/src/Static.jl:114
  ifelse(::Static.True, ::Any, ::Any)
   @ Static /usr/local/share/julia/packages/Static/TjBVO/src/Static.jl:113
  ...
```
looks like we are still hitting `missing == missing`

src/DataWrangling/metadata_field.jl
```diff
@@ -371,8 +371,10 @@ function centers_to_interfaces(z_centers)
     return z_faces
 end
 
-@inline nan_convert_missing(FT, ::Missing) = convert(FT, NaN)
-@inline nan_convert_missing(FT, d::Number) = convert(FT, d)
+# Convert missing values to NaN
+@inline nan_convert_missing(FT, x::Number) = convert(FT, x)
+@inline nan_convert_missing(FT, x::Missing) = convert(FT, NaN)
+@inline nan_convert_missing(FT, x, missing_val) = ifelse(ismissing(x) || x == missing_val, convert(FT, NaN), nan_convert_missing(FT, x))
```
simone-silvestri:
```suggestion
@inline nan_convert_missing(FT, x::Missing, missing_val) = convert(FT, NaN)
@inline nan_convert_missing(FT, x, missing_val) = ifelse(ismissing(x) || x == missing_val, convert(FT, NaN), nan_convert_missing(FT, x))
```

bgroenks96: I think this won't work if `missing_val` is `missing` and `x` isn't because then `x == missing_val` is also `missing`. But [a955acf](https://github.com/NumericalEarth/NumericalEarth.jl/pull/332/commits/a955acf985e8412ecb5fd6f8212a7910e3c98f05) should finally fix this...

src/DataWrangling/metadata_field.jl
```diff
@@ -371,8 +371,11 @@ function centers_to_interfaces(z_centers)
     return z_faces
 end
 
+# Convert missing values to NaN
+@inline nan_convert_missing(FT, x::Number) = convert(FT, x)
 @inline nan_convert_missing(FT, ::Missing) = convert(FT, NaN)
```
simone-silvestri, line 376:
```suggestion
```
I think these are not used anymore

src/DataWrangling/SoilGrids/SoilGrids.jl
```diff
@@ -0,0 +1,121 @@
+module SoilGrids
+
+export SoilGrids2
+
+using Downloads: Downloads
+using Oceananigans: CPU, Field, Center, location, fill_halo_regions!
```
simone-silvestri:
```suggestion
using Oceananigans: Center
```

src/DataWrangling/metadata_field.jl
bgroenks96: They are used in lines 377 and 378. They aren't strictly speaking necessary, but I think it simplifies the logic a bit by encoding it through dispatch rather than another `ifelse`.

simone-silvestri: ok, I think tests should pass now!

PR #334, opened 2026-06-06
Allow using the internal model's diffusivity in the `SkinTemperature`

simone-silvestri:
This allows us to assess the skin temperature consistently with what the ocean does in the interior. In this case it will be something like:
```julia
ocean = ocean_simulation(grid; ...)
interior_flux = DiffusiveFlux(InteriorDiffusivity(), Δzᶜᶜᶠ(i, j, k, grid) / 2))
interfaces = ComponentInterfaces(atmosphere, ocean; atmosphere_ocean_interface_temperature = SkinTemperature(interior_flux))
```

### DIFF
--- a/src/EarthSystemModels/InterfaceComputations/interface_states.jl
+++ b/src/EarthSystemModels/InterfaceComputations/interface_states.jl
@@ -348,9 +348,51 @@ end
 
 SkinTemperature(internal_flux; max_ΔT=5) = SkinTemperature(internal_flux, max_ΔT)
 
+"""
+    DiffusiveFlux(δ, κ)
+
+Internal flux ``J = - κ (Tₛ - Tᵢ) / δ`` between the interior temperature ``Tᵢ``,
+located a distance ``δ`` below the interface (typically half the spacing of the topmost
+interior cell), and the interface temperature ``Tₛ``. The diffusivity `κ` (m² s⁻¹) is
+either a prescribed constant or an [`InteriorDiffusivity`](@ref) assessed from the
+interior model.
+"""
 struct DiffusiveFlux{Z, K}
     δ :: Z # Boundary layer thickness, as a first guess we will use half the grid spacing
     κ :: K # diffusivity in m² s⁻¹
+end
+
+"""
+    InteriorDiffusivity(FT = Oceananigans.defaults.FloatType; minimum_diffusivity = 1.4e-7)
+
+Diffusivity for a [`DiffusiveFlux`](@ref) that is assessed from the interior model (for example the near-surface 
+vertical diffusivity predicted by the ocean turbulence closure) instead of being prescribed. The assessed value is floored by
+`minimum_diffusivity`, which defaults to the molecular thermal diffusivity of seawater, guarding stably-stratified conditions 
+in which modeled diffusivities vanish.
+"""
+struct InteriorDiffusivity{FT}
+    minimum_diffusivity :: FT
+end
+
+InteriorDiffusivity(FT::DataType = Oceananigans.defaults.FloatType; minimum_diffusivity = 1.4e-7) =
+    InteriorDiffusivity(convert(FT, minimum_diffusivity))
+
+@inline internal_diffusivity(κ::Number, Ψᵢ) = κ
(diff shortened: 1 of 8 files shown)


src/Oceans/Oceans.jl
```diff
@@ -110,7 +112,55 @@ function EarthSystemModels.InterfaceComputations.ComponentExchanger(ocean::Ocean
...
+                                                         Oceananigans.fields(model))
+
+    return ComponentExchanger((; u, v, T, S, κ), nothing)
+end
+
+#####
+##### Near-surface vertical diffusivity assessment
+#####
+
+# Total vertical tracer diffusivity at the surface. Salls back to zero for closures without vertical diffusivity
```
xkykai:
```suggestion
# Total vertical tracer diffusivity at the surface. Falls back to zero for closures without vertical diffusivity
```

src/Oceans/Oceans.jl
```diff
@@ -110,7 +112,55 @@ function EarthSystemModels.InterfaceComputations.ComponentExchanger(ocean::Ocean
...
+@inline Σκzᶜᶜᶠ(i, j, k, grid, closure::Tuple{<:Any, <:Any, <:Any}, K::Tuple{<:Any, <:Any, <:Any}, id, clock, model_fields) =
+    κzᶜᶜᶠ(i, j, k, grid, closure[1], K[1], id, clock, model_fields) +
+    κzᶜᶜᶠ(i, j, k, grid, closure[2], K[2], id, clock, model_fields) +
+    κzᶜᶜᶠ(i, j, k, grid, closure[3], K[3], id, clock, model_fields) 
+
+@inline Σκzᶜᶜᶠ(i, j, k, grid, closure::Tuple{<:Any, <:Any, <:Any, <:Any}, K::Tuple{<:Any, <:Any, <:Any, <:Any}, id, clock, model_fields) =
+    κzᶜᶜᶠ(i, j, k, grid, closure[1], K[1], id, clock, model_fields) +
+    κzᶜᶜᶠ(i, j, k, grid, closure[2], K[2], id, clock, model_fields) +
+    κzᶜᶜᶠ(i, j, k, grid, closure[3], K[3], id, clock, model_fields) +
+    κzᶜᶜᶠ(i, j, k, grid, closure[4], K[4], id, clock, model_fields)
```
xkykai, line 155: why are these needed since we have the recursive call?

src/EarthSystemModels/InterfaceComputations/interface_states.jl
```diff
@@ -403,10 +436,11 @@ end
     # Flux balance: T★ = (Tᵢ κ - (Jᵀ + Ωc Tᵃᵗ) δ) / (κ - Ωc δ)
     # where Ωc = 𝒬ᵀ λ / ΔT. Multiply through by ΔT to avoid Inf when ΔT → 0.
     Ωᵀ = 𝒬ᵀ * λ  # unnormalized sensible heat coefficient (= Ωc * ΔT)
-    D  = F.κ * ΔT - Ωᵀ * F.δ
-    T★ = (Ψᵢ.T * F.κ * ΔT - (Jᵀ * ΔT + Ωᵀ * Tᵃᵗ) * F.δ) / D
-
-    return ifelse(D == 0, Ψₛ.temperature, T★)
+    D  = κ * ΔT - Ωᵀ * F.δ
+    T★ = (Ψᵢ.T * κ * ΔT - (Jᵀ * ΔT + Ωᵀ * Tᵃᵗ) * F.δ) / D
+    T★ = ifelse(D == 0, Ψₛ.temperature, T★)
```
xkykai, line 442: out of curiosity what does this mean?

xkykai (approved):

src/Oceans/Oceans.jl
simone-silvestri: for performance it is better to use explicit calls so we code up methods up to a certain amount of closures, if we use more we fall back to the recursive call

src/EarthSystemModels/InterfaceComputations/interface_states.jl
simone-silvestri: This is the explicit formulation of the balance temperature computed as a balance of fluxes (ocean and atmosphere side), where the sensible heat flux from the atmosphere is treated implicitly

PR #336, opened 2026-06-08
Fix veros' docs fragility

simone-silvestri:
Every time the python ecosystem is updated, the veros simulation breaks.
In this PR I am fixing the version of veros to 1.6.2 and making sure we install the correct requirements for the package.
This should solve the fragility issue


navidcy:
Yes!
Why didn't the docs built run? That's the one that breaks... right?

simone-silvestri:
not sure, they do not target changes in the extensions? Let me change that

navidcy (approved):
if the veros example builds in the docs then merge!

PR #337, opened 2026-06-09
Fix `JRA55` prescribed atmosphere seam BCs

taimoorsohail:
## What changed

This PR fixes the tripolar seam boundary conditions for regridded prescribed-atmosphere velocities and applies the same treatment consistently across prescribed atmospheres.

The main change is that exchange-grid `u` and `v` fields now use velocity-aware boundary conditions, so on tripolar grids their north-fold halo extension flips sign as expected. The boundary-condition setup is factored into a shared helper and used both for prescribed-atmosphere velocity fields and for the regridded exchange-state buffers.

In addition, this PR fixes mixed `DateTime` / `CFTime` comparisons in `compute_native_date_range`, which were causing metadata date-range selection failures for some prescribed datasets, and includes a small Ocean Station Papa naming/doc cleanup.

## Why it changed

The original issue was that `JRA55PrescribedAtmosphere` velocities behaved correctly on their native grid, but after interpolation onto a tripolar exchange grid the regridded `u` and `v` fields did not carry the expected signed seam behavior. That was because those exchange buffers were allocated with default boundary conditions instead of velocity-specific ones.

While generalizing that fix, a related metadata date-comparison issue surfaced and was fixed as well.

## Impact

- Restores the expected tripolar seam sign flip for regridded prescribed-atmosphere velocities.
- Applies the fix consistently across JRA55, ECCO, ERA5, and Ocean Station Papa prescribed atmospheres.
- Prevents mixed `DateTime` / `CFTime` failures in metadata date-range selection.

## Validation

- Added regression coverage for regridded prescribed-atmosphere exchange-state velocities on a tripolar grid for JRA55, ECCO, ERA5, and Ocean Station Papa.
- Added a regression test for `compute_native_date_range` with native `DateTime` values and `CFTime` bounds.

[edited above by @navidcy]


simone-silvestri:
Looks good. Maybe we should do the same to every atmosphere? (ECCO, ERA5, ...). I would define a function `velocity_boundary_conditions` which we can use in all the atmosphere constructors

taimoorsohail:
MWE that reproduces the issue on main: 

```Julia
using CairoMakie
using NumericalEarth
using Oceananigans
using Oceananigans.BoundaryConditions: fill_halo_regions!
using Oceananigans.TimeSteppers: update_state!

const Nx = 120
const Ny = 60
const Nz = 1
const figdir = expanduser("/home/tsohail/uom/ocean-ensembles/figures/")

grid = TripolarGrid(CPU();
                    size = (Nx, Ny, Nz),
                    z = (-1, 0),
                    halo = (7, 7, 7))

                        free_surface = SplitExplicitFreeSurface(grid, substeps = 20))

atmosphere = JRA55PrescribedAtmosphere(CPU(); time_indices_in_memory = 2)
radiation  = JRA55PrescribedRadiation(CPU(); time_indices_in_memory = 2)

coupled_model = OceanOnlyModel(ocean; atmosphere, radiation)

u_field_atmos = coupled_model.interfaces.exchanger.atmosphere.state.u
v_field_atmos = coupled_model.interfaces.exchanger.atmosphere.state.v

u_field_ocean = ocean.model.velocities.u
v_field_ocean = ocean.model.velocities.v

set!(u_field_ocean, 1)
set!(v_field_ocean, 0)

fill_halo_regions!(u_field_ocean)
fill_halo_regions!(v_field_ocean)
fill_halo_regions!(u_field_atmos)
fill_halo_regions!(v_field_atmos)

fig = Figure(size = (1200, 900))

ax1 = Axis(fig[1, 1],
           title = "JRA55 atmosphere u at the tripolar seam",
           xlabel = "Seam index",
           ylabel = "Velocity")

lines!(ax1, u_field_atmos[:, Ny]; label = "JRA55 Atmosphere", linewidth = 3, color = :dodgerblue)
lines!(ax1, u_field_ocean[:, Ny, 1]; label = "Ocean", linewidth = 3, color = :orangered)

outname = figdir * "jra55_tripolar_seam.png"
axislegend(ax1, position = :rb)

save(outname, fig)

@info "Saved seam plot" outname

```

Gives:

jra55_tripolar_seam

simone-silvestri:
This is definitely the correct path to go (probably we should think of how to make these boundary conditions automatically satisfied). However, I don't think the exchanger halos are ever explicitly filled with `fill_halo_regions!` at the moment (we interpolate the atmospheric state "inside" the halo automatically without calling the fill halo method). So if there is a problem of stability this will probably not solve it.

simone-silvestri (approved):

src/Atmospheres/prescribed_atmosphere.jl
```diff
@@ -30,8 +30,16 @@ function Base.show(io::IO, pa::PrescribedAtmosphere)
...
 
 function default_atmosphere_velocities(grid, times)
-    ua = FieldTimeSeries{Center, Center, Nothing}(grid, times)
-    va = FieldTimeSeries{Center, Center, Nothing}(grid, times)
+    velocity_bcs = FieldBoundaryConditions(grid, (Center(), Center(), nothing))
+
+    if grid isa OrthogonalSphericalShellGrids.TripolarGrid
+        north_bc = OrthogonalSphericalShellGrids.north_fold_boundary_condition(grid)(-1)
+        velocity_bcs = FieldBoundaryConditions(grid, (Center(), Center(), nothing);
+                                               north = north_bc)
```
navidcy: if we don't prescribe all the other non-north bcs what are they? the defaults? nothing?

simone-silvestri: they should be the defaults

navidcy (approved):

navidcy:
If CI CPU & GPU passes feel free to merge! (No need to wait for Docs)

navidcy:
I'm not sure what's happening with the ERA5Prescribed... I think it's a download issue? @simone-silvestri?

navidcy:
The Ocean Station Papa is a dataset at a single location; no meaning testing for the tripolar seam...

navidcy:
test_prescribed_atmosphere.jl pass locally for me now! let's see

taimoorsohail:
There is some download issue in the tests at the moment @navidcy ... is the server down or some bug?

navidcy:
There were a few bugs before that were revealed after I fixed the download issues. So let's try to see if CI passes.
GPU CI passes so that's a very good sign!

PR #338, opened 2026-06-09
Fix clock in `nonhydrostatic_ocean_simulation`

simone-silvestri:
it was not added to the function signature


navidcy (approved):

navidcy:
This was something that #321 missed, right?

PR #342, opened 2026-06-10
Use `AdaptiveVerticallyImplicitDiscretization` as a default for momentum and tracer advection

simone-silvestri:


taimoorsohail (approved):

taimoorsohail:
I have patched Oceananigans so this checkpointing error won't recur: https://github.com/CliMA/Oceananigans.jl/pull/5690

taimoorsohail:
@simone-silvestri -- why is the checkpointing test failing here then? If the state == 0 fix has already been implemented in the latest Oceananigans version? I am confused...

simone-silvestri:
let's see if the new tests pass, the previous failing tests were using Oceananigans 0.110.1

PR #343, opened 2026-06-10
Fix north halo filling

taimoorsohail:
Summary: fix north-halo handling for coupled flux fields on tripolar grids by adding north-fold-aware vector boundary conditions and explicitly filling interface and net-flux halos before downstream use.

Validation: not run locally per repo workflow; GitHub CI will validate the change.


src/EarthSystemModels/InterfaceComputations/component_interfaces.jl
```diff
@@ -48,6 +50,13 @@ end
 @inline computed_fluxes(interface::AtmosphereInterface)  = interface.fluxes
 @inline computed_fluxes(interface::SeaIceOceanInterface) = interface.fluxes
 
+vector_component_boundary_conditions(grid, loc) = FieldBoundaryConditions(grid, loc)
+
+function vector_component_boundary_conditions(grid::OrthogonalSphericalShellGrids.TripolarGrid, loc)
```
simone-silvestri:
```suggestion
function vector_component_boundary_conditions(grid::OrthogonalSphericalShellGrids.TripolarGridOfSomeKind, loc)
```
to include an immersed boundary

src/EarthSystemModels/time_step_earth_system_model.jl
```diff
@@ -61,19 +80,30 @@ function Oceananigans.TimeSteppers.update_state!(coupled_model::EarthSystemModel
     InterfaceComputations.correct_state!(exchanger.sea_ice,    grid)
     InterfaceComputations.correct_state!(exchanger.ocean,      grid)
 
+    !isnothing(exchanger.atmosphere) && fill_halo_regions!(exchanger.atmosphere.state)
+
```
simone-silvestri:
```suggestion
```

src/EarthSystemModels/time_step_earth_system_model.jl
```diff
@@ -61,19 +80,30 @@ function Oceananigans.TimeSteppers.update_state!(coupled_model::EarthSystemModel
...
     compute_atmosphere_ocean_fluxes!(coupled_model)
     compute_atmosphere_sea_ice_fluxes!(coupled_model)
     compute_atmosphere_land_fluxes!(coupled_model)
     compute_sea_ice_ocean_fluxes!(coupled_model)
 
+    fill_interface_flux_halos!(isnothing(coupled_model.interfaces.atmosphere_ocean_interface) ? nothing : coupled_model.interfaces.atmosphere_ocean_interface.fluxes)
+    fill_interface_flux_halos!(isnothing(coupled_model.interfaces.atmosphere_sea_ice_interface) ? nothing : coupled_model.interfaces.atmosphere_sea_ice_interface.fluxes)
+    fill_interface_flux_halos!(isnothing(coupled_model.interfaces.sea_ice_ocean_interface) ? nothing : coupled_model.interfaces.sea_ice_ocean_interface.fluxes)
+    fill_interface_flux_halos!(isnothing(coupled_model.interfaces.atmosphere_land_interface) ? nothing : coupled_model.interfaces.atmosphere_land_interface.fluxes)
+
```
simone-silvestri:
```suggestion
```

src/EarthSystemModels/time_step_earth_system_model.jl
```diff
@@ -61,19 +80,30 @@ function Oceananigans.TimeSteppers.update_state!(coupled_model::EarthSystemModel
...
     update_net_fluxes!(coupled_model, radiation)
     update_net_fluxes!(coupled_model, atmosphere)
     update_net_fluxes!(coupled_model, land)
     update_net_fluxes!(coupled_model, sea_ice)
     update_net_fluxes!(coupled_model, ocean)
 
+    net_ocean_fluxes = coupled_model.interfaces.net_fluxes.ocean
+    fill_halo_regions!((net_ocean_fluxes.u, net_ocean_fluxes.v, net_ocean_fluxes.T, net_ocean_fluxes.S))
+    fill_halo_regions!(coupled_model.interfaces.net_fluxes.sea_ice)
+
```
simone-silvestri:
```suggestion
```

src/Oceans/ocean_simulation.jl
```diff
@@ -52,6 +52,13 @@ end
 @inline build_top_bc(flux_field, ::Nothing) = FluxBoundaryCondition(flux_field)
 @inline build_top_bc(flux_field, additional) = FluxBoundaryCondition(MultipleFluxes(flux_field, additional); discrete_form=true)
 
+vector_component_boundary_conditions(grid, loc) = FieldBoundaryConditions(grid, loc)
+
+function vector_component_boundary_conditions(grid::TripolarGrid, loc)
```
simone-silvestri: We can use the function in `InterfaceComputations` here?

simone-silvestri:
The only fill halo required should be the one for the sea-ice atmosphere stresses, which is done here:
https://github.com/CliMA/ClimaSeaIce.jl/blob/6f311fd66184b2e41cb4ea00e9653ad3499d33f9/src/SeaIceDynamics/sea_ice_external_stress.jl#L74-L78
So in theory we do not need any fill halo in this PR, just making sure that all the fluxes have the correct BCs though.
This is also good practice even for fluxes where we do not need to fill the halos, because we are not sure how the code will change in the future, so better having correctly defined fields

simone-silvestri (approved):

simone-silvestri:
I will merge, if we see that we need extra fill_halo regions we can open another PR

taimoorsohail:
Thanks @simone-silvestri ! Sorry this fell off my radar a bit :)

PR #350, opened 2026-06-16
Couple a compressible (terrain-following) Breeze atmosphere

glwagner:
## Summary

`AtmosphereOceanModel` could not couple a `CompressibleDynamics` (terrain-following)
Breeze atmosphere — the air–sea coupler's Breeze interface assumed anelastic dynamics in
two spots and errored before the first step. Both are fixed in
`ext/NumericalEarthBreezeExt/breeze_atmosphere_interface.jl`:

1. **`interpolate_state!`** read `atmosphere.dynamics.reference_state.density` /
   `.surface_pressure`. Compressible dynamics have `reference_state === nothing`, giving
   `FieldError: type Nothing has no field density`. Now falls back to the 3-D
   `terrain_reference_density` (surface level `[i,j,1]` is exactly what the interpolation
   kernel indexes) and the scalar `surface_pressure`.
2. **`surface_layer_height`** reads the lowest cell's `zspacing`, which indexes a device
   array on a terrain-following GPU grid → `Scalar indexing is disallowed`. The single
   scalar read is now wrapped in `@allowscalar` (GPUArraysCore), as the prior comment hinted.

The anelastic path is unchanged (the new branch only triggers when `reference_state === nothing`).

## Testing

Verified on an NVIDIA GH200: a terrain-following compressible Breeze atmosphere over real
fjord terrain (Nærøyfjord) coupled to a hydrostatic + CATKE immersed-boundary ocean now
steps stably, with air–sea sensible/latent/momentum fluxes computed at the interface and
the ocean responding to the coupled wind stress. Before this change both a full
hydrostatic ocean and a `SlabOcean` failed identically (so it was the atmosphere side).

### DIFF
--- a/ext/NumericalEarthBreezeExt/breeze_atmosphere_interface.jl
+++ b/ext/NumericalEarthBreezeExt/breeze_atmosphere_interface.jl
@@ -1,4 +1,5 @@
 using Oceananigans.Grids: Center
+using GPUArraysCore: @allowscalar
 using Breeze.AtmosphereModels: thermodynamic_density
 using NumericalEarth.Atmospheres: AtmosphereThermodynamicsParameters
 using NumericalEarth.EarthSystemModels: component_model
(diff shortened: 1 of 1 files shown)


glwagner:
this PR has been superceded, and we are coupled now.

PR #363, opened 2026-06-18
Add GloFASPrescribedLand: ERA5-routed river discharge onto the ocean coastline

glwagner:
## Summary

Adds a `GloFASPrescribedLand` component that forces the ocean with **GloFAS** river discharge, placed on the ocean coastline conserving freshwater volume.

GloFAS (Global Flood Awareness System) forces the LISFLOOD hydrological + channel-routing model with **ERA5 runoff** to produce river discharge already accumulated downstream to river mouths ([Harrigan et al. 2020](https://essd.copernicus.org/articles/12/2043/2020/)). It's the ERA5-consistent analogue of JRA55's pre-routed river freshwater flux. Raw ERA5 surface runoff, by contrast, is unrouted and would be largely discarded when interpolated onto an ocean grid and land-masked.

## What's included

**DataWrangling — GloFAS dataset**
- New `GloFAS` module: `GloFASReanalysis` (v4, 0.05°, daily river discharge, 1979–present), wired into the existing `Metadata → FieldTimeSeries` pipeline.
- CDS extension download via `cems-glofas-historical`, with server-side `area` subsetting for `BoundingBox` regions. GloFAS lives on the **EWDS** endpoint (separate from the ERA5 CDS), so the download overrides CDSAPI's url per call through its `ScopedValues`, reusing the shared ECMWF token — a `~/.cdsapirc` pointed at the ERA5 CDS keeps working. `GLOFAS_CDSAPI_URL` / `GLOFAS_CDSAPI_KEY` override the defaults.

**Lands — conservative coastal routing**
- `RiverRouting` + `build_river_routing`: locate river mouths from the dataset's land/ocean boundary, map each to the nearest active ocean cell, with weight `ρ_freshwater / A_ocean` so the area integral of the deposited flux equals the total discharge × density.
- Scatter `interpolate_state!`: **one thread per destination ocean cell** (no atomics → GPU-safe), with temporal `FieldTimeSeries` interpolation at the exact mouth cell. `PrescribedLand` gains an optional `river_routing` field; the JRA55 (pre-routed, pointwise) path is unchanged and dispatches as before.

**Docs & tests**
- Retitle the ERA5 example to *"ERA5 and GloFAS reanalysis data"* with a new section that downloads Amazon-mouth discharge and visualizes the coastal routing.
- `test/test_river_routing.jl`: outlet detection + volume-conservation (direct and through the real scatter).

## Testing
- `test_river_routing.jl` passes on CPU; all ExplicitImports QA checks pass.
- **Verified end-to-end against the live EWDS API** (Amazon-mouth box): download succeeds, NetCDF parsing is correct (variable `dis24`, `latitude`/`longitude` coords, latitude north-to-south), discharge magnitudes are physical (~196,000 m³/s at the Amazon), and all 281 detected river mouths route onto 157 coastal ocean cells.

## Notes
- Routing requires the target ocean grid to have **active coastal cells** — i.e. enough near-surface vertical resolution that shallow shelf cells aren't masked out by `GridFittedBottom`. The tutorial grid was updated accordingly (a single thick level drops every mouth).
- GloFAS native longitude is kept in the −180–180 convention to match typical regional ocean grids; the CDS files (0–360) are reconciled on load. A global (`region = nothing`) GloFAS forcing against a 0–360 ocean grid would need convention handling in the routing — out of scope here.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -43,7 +43,7 @@ examples = [
     Example("Global climate simulation", "global_climate_simulation", false),
     Example("Veros ocean simulation", "veros_ocean_forced_simulation", false),
     Example("Breeze over four oceans", "breeze_over_four_oceans", false),
-    Example("Exploring ERA5 reanalysis data", "exploring_era5_reanalysis_data", true),
+    Example("ERA5 and GloFAS reanalysis data", "exploring_era5_reanalysis_data", true),
     Example("ERA5-forced slab land", "era5_forced_slab_land", true),
     Example("Breeze over slab land", "breeze_over_slab_land", true),
 ]
--- a/docs/src/NumericalEarth.bib
+++ b/docs/src/NumericalEarth.bib
@@ -153,3 +153,14 @@
   pages = {217–240},
   language = {English}
 }
+
+@article{harrigan2020glofas,
+  title={{GloFAS-ERA5} operational global river discharge reanalysis 1979--present},
+  author={Harrigan, Shaun and Zsoter, Ervin and Alfieri, Lorenzo and Prudhomme, Christel and Salamon, Peter and Wetterhall, Fredrik and Barnard, Christopher and Cloke, Hannah and Pappenberger, Florian},
+  journal={Earth System Science Data},
+  volume={12},
+  number={3},
+  pages={2043--2060},
+  year={2020},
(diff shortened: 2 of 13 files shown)


jagoosw (approved):
Makes sense to me

glwagner:
yay!

PR #366, opened 2026-06-19
Restrict native grid vertical extent for z-restricted downloads

glwagner:
## Summary

A `BoundingBox`/`Column` `z = (z_bottom, z_top)` already restricts the CopernicusMarine/GLORYS *download* (via the already-merged `depth_bounds_kw`). This PR extends that so the **native grid** built from the restricted file follows the file's vertical extent rather than the dataset's full-column `Nz`.

- **GLORYS `region_suffix`** appends a `z_suffix` so a z-restricted download caches to a *distinct* file — `skip_existing` no longer hands back the full-column file. Regions without `z` keep their existing filenames (backward compatible).
- **New `vertical_size` helper**: an explicit z-interface vector sets `Nz` to `length(z) - 1`; a tuple/`nothing` keeps the dataset's full-column `Nz`. Wired into all three `construct_native_grid` methods (`::Nothing`, `BoundingBox`, `Column`).
- **Docstring** updates for the `BoundingBox` `z` keyword.

## Tests

- `vertical_size` unit tests (vector → `length-1`, tuple → full `Nz`)
- `depth_bounds_kw` mapping (negative-downward `z` → positive Copernicus depth bounds; empty bbox → unrestricted)
- z-restricted download builds a grid with shrunken `Nz` matching the file, and the field loads onto it
- distributed-vs-serial GLORYS `set!` check across `Partition(4,1)`, `(1,4)`, `(2,2)`

## Notes

- No `Project.toml` dependency changes.
- Backward compatible: regions without `z` keep prior filenames and full `Nz`.
- The download/distributed tests require CopernicusMarine credentials + network, so they run only in the relevant CI lanes.


test/test_distributed_utils.jl
```diff
@@ -6,10 +6,22 @@ MPI.Init()
 using CFTime
 using Dates
 using NCDatasets
-using NumericalEarth.DataWrangling: metadata_path
+using NumericalEarth.DataWrangling: metadata_path, BoundingBox
+using NumericalEarth.DataWrangling.GLORYS: GLORYSDaily
 using Oceananigans.DistributedComputations
 using Oceananigans.DistributedComputations: reconstruct_global_grid
 
+# h5py/hdf5 are required to write the CopernicusMarine subset; add them once on
```
simone-silvestri: I think this has been fixes in CopernicusMarine v0.1.2. But, given that the new CopernicusMarine does not use python, we probably do not need it

test/test_distributed_utils.jl
```diff
@@ -6,10 +6,22 @@ MPI.Init()
...
+# the root rank to avoid concurrent CondaPkg writes across ranks.
+using CondaPkg
+@root begin
+    CondaPkg.add("h5py"; channel="conda-forge", version=">=3.0,<3.13")
+    CondaPkg.add("hdf5"; channel="conda-forge", version="<2")
+end
+MPI.Barrier(MPI.COMM_WORLD)
```
simone-silvestri:
```suggestion
```

simone-silvestri (approved):

PR #372, opened 2026-06-19
Make radiation→ocean forcing hook live; rename radiative flux functions

glwagner:
## Summary

Two related cleanups around how radiation forces the ocean, prompted by tracing `compute_radiative_forcing!` and `TwoColorRadiation`.

### 1. `compute_radiative_forcing!` — from dead code to a live extension hook

Previously `compute_radiative_forcing!` was defined but **never called**, and its dispatch was backwards: the fallback was a no-op while the `TwoColorRadiation` method did host-side work that the in-kernel `shortwave_radiative_forcing` already does.

It's now the generic, overridable **precompute hook** for radiation models whose in-water absorption requires a vertical pass over ocean tracers (e.g. biogeochemical / chlorophyll-based schemes):

- generic `compute_radiative_forcing!(radiation, coupled_model)` — the overridable entry point (fallback no-op),
- `compute_radiative_forcing!(::TwoColorRadiation, coupled_model)` — explicit no-op (analytic absorption; its surface flux is set per-column in the flux kernel),
- called from `compute_radiation_ocean_fluxes!` right before the kernel `launch!`.

Behavior today is unchanged (no-op for the only existing scheme and for `nothing`), but the extension point is now genuinely wired in. Also fixes notation `Iˢʷ` → `ℐˢʷ` to match the script-I used in the kernel (`ℐꜜˢʷ`/`ℐₜˢʷ`/`ℐꜛˡʷ`).

### 2. Rename the radiative flux family: `apply_air_*` → `compute_radiation_*`

The fluxes are sourced from the `radiation` component (`coupled_model.radiation`), not the atmosphere — these functions never touch `coupled_model.atmosphere`. The new names use the component-pair form parallel to the turbulent `compute_atmosphere_ocean_fluxes!`:

| Old | New |
|---|---|
| `apply_air_sea_radiative_fluxes!` | `compute_radiation_ocean_fluxes!` |
| `apply_air_sea_ice_radiative_fluxes!` | `compute_radiation_sea_ice_fluxes!` |
| `apply_air_land_radiative_fluxes!` | `compute_radiation_land_fluxes!` |

Files renamed (`git mv`, history preserved); all definitions, `_`-prefixed kernels, `(::Any)`/`{<:Nothing}` fallbacks, `include`s, and doc/comment references updated. `grep` confirms no stale `apply_air_*` tokens remain.

Left intentionally unchanged: `air_sea_interface_radiation_state` (describes the radiation *state at the surface interface* — distinct concept, separate scope).

## Test plan

- [ ] CI (no compile check run locally — no Julia REPL available; pure token rename + a no-op hook call, definition/call sites updated symmetrically)


simone-silvestri (approved):

glwagner:
I think this is stale -- @simone-silvestri can you confirm?

PR #374, opened 2026-06-20
Add regression test for PrescribedOcean exchange_grid

glwagner:
Follow-up to #359, which added the `land=nothing` argument to
`exchange_grid(atmosphere, ocean::PrescribedOcean, sea_ice, land)` so a
`PrescribedOcean` can be coupled alongside a land component.

This adds a focused regression test (`test/test_prescribed_ocean.jl`) for that
path. Without the fix, the 4-argument call falls through to the generic
`exchange_grid` method (`ocean.model.grid`) and throws `FieldError`, since
`PrescribedOcean` has no `model` field — confirmed by running the test against
the pre-#359 code.

The test is download-free, runs on both CPU and GPU, and completes in ~10s.
It is auto-discovered by `ParallelTestRunner.find_tests`, so no change to
`runtests.jl` is needed.


test/test_prescribed_ocean.jl
```diff
@@ -0,0 +1,26 @@
+include("runtests_setup.jl")
+
+using NumericalEarth.Oceans: PrescribedOcean
```
navidcy: It's been exported; no need for this line
```suggestion
```

navidcy (approved):

simone-silvestri (approved):

PR #377, opened 2026-06-23
Remove specific_humidity from ERA5 single-level variable maps

jtbuch:
## Problem

`specific_humidity` is listed in the **single-level** ERA5 variable maps (`ERA5_dataset_variable_names` and `ERA5_netcdf_variable_names` in `src/DataWrangling/ERA5/ERA5_single_levels.jl`), but it is **not a product of the
`reanalysis-era5-single-levels` CDS dataset** — it only exists on pressure levels and model levels.

### Reproduction (CDS API directly)

| Request | Dataset | Result |
|---|---|---|
| `2m_temperature` + 10m winds + `surface_pressure` | single-levels | ✅ success |
| `specific_humidity` alone | single-levels | ❌ failed |
| `specific_humidity` (levels 500/850/1000) | pressure-levels | ✅ success |

## Fix

Remove the two `specific_humidity` entries from the single-level maps. It is already correctly present in the pressure-level maps (`ERA5PL_dataset_variable_names` / `ERA5PL_netcdf_variable_names`), so the splitter
now routes it to the pressure-level path, where it downloads fine.

## Verification

With this change applied, an end-to-end ERA5 staging run requesting `[surface_pressure, geopotential, eastward_velocity, northward_velocity, temperature, specific_humidity, specific_cloud_liquid_water_content, specific_cloud_ice_water_content]` succeeds:
- single-level batch → `[surface_pressure, eastward_velocity, northward_velocity, temperature]`
- `specific_humidity` is staged from `ERA5HourlyPressureLevels` alongside `geopotential` and the cloud-water-content fields.


simone-silvestri:
huh, so we cannot have a prescribed ERA5 atmosphere to drive the ocean? probably we need to compute specific humidity as a function of relative humidity in ERA5PrescribedAtmosphere?

simone-silvestri:
Ah stupid me, it is already done 😅

simone-silvestri (approved):

PR #380, opened 2026-06-24
Add ERA5 pure Julia CDS client via package extension

aklocker42:
## Summary

This PR adds optional pure Julia ERA5 data downloading via the Copernicus Climate Data Store (CDS) API, eliminating the need for Python `era5cli` dependency. The implementation uses package extensions (Julia 1.9+) to keep the core package lightweight.

### New Features
- **Pure Julia CDS client** (`ERA5_cds_client.jl`) - Direct API access to CDS without Python
- **ERA5 variable mappings** (`ERA5_variables.jl`) - Maps symbolic names to CDS parameter names
- **ERA5PrescribedLand component** (`ERA5_prescribed_land.jl`) - Land surface component (placeholder)
- **Package extension** (`NumericalEarthCDSClientExt`) - Loads CDS client only when `HTTP` and `JSON3` are available

### Technical Implementation
- Moved `HTTP` and `JSON3` from hard dependencies to weak dependencies (`[weakdeps]`)
- CDS client loads automatically via extension when users `using HTTP, JSON3`
- Maintains backward compatibility - existing Python-based downloader via `CopernicusClimateDataStore` continues to work
- Added compatibility fix for `TimeSeriesInterpolation` across Oceananigans versions

## Usage

**For users who want pure Julia downloads:**
```julia
using NumericalEarth
using HTTP, JSON3  # Activates the CDS client extension

# Configure credentials once: ~/.cdsapirc
# url: https://cds.climate.copernicus.eu/api
# key: <your-api-key>

# Now pure Julia downloads available
atmosphere = ERA5PrescribedAtmosphere(arch; dataset=ERA5Hourly(), ...)
```

**For users who don't need ERA5 or prefer Python:**
- No change needed - `HTTP` and `JSON3` are not loaded
- Existing Python-based downloader still works via `CopernicusClimateDataStore`

## Testing

- CDS client tested with multi-GPU distributed simulations
- Credentials discovery tested (environment variables, `~/.cdsapirc`, era5cli config)
- Extension activation verified with and without `HTTP`/`JSON3`

## Benefits

1. **No Python dependency** for ERA5 downloads
2. **Lightweight** - HTTP/JSON3 only loaded when needed
3. **Standard credentials** - Uses same `~/.cdsapirc` as Python CDS client
4. **Backward compatible** - Existing code continues to work


ext/NumericalEarthCDSClientExt.jl
```diff
@@ -0,0 +1,18 @@
...
+# This allows pure Julia ERA5 downloads without Python dependencies
+
+using NumericalEarth
+using NumericalEarth.DataWrangling.ERA5
+using HTTP
+using JSON3
+
+# Include the CDS client implementation
+include("../src/DataWrangling/ERA5/ERA5_variables.jl")
+include("../src/DataWrangling/ERA5/ERA5_cds_client.jl")
```
glwagner: I don't think this is correct... the functionality should be brought in when we write `using NumericalEarth`

src/DataWrangling/ERA5/ERA5_cds_client.jl
glwagner, line 1: should this file be moved to the extension?

src/DataWrangling/ERA5/ERA5_variables.jl
glwagner, line 1: This file isn't included in the top-level module file `ERA5.jl` but I believe it should be

src/Grids/pressure_level_vertical_discretization.jl
```diff
@@ -8,7 +8,13 @@ using Oceananigans: instantiated_location
...
 using Oceananigans.OutputReaders: FieldTimeSeries
 using Oceananigans.Grids: AbstractVerticalCoordinate, AbstractUnderlyingGrid, Center, Face, Flat, LatitudeLongitudeGrid, topology
-using Oceananigans.OutputReaders: TimeSeriesInterpolation
+# TimeSeriesInterpolation may not exist in all versions - import with fallback
+try
+    using Oceananigans.OutputReaders: TimeSeriesInterpolation
+catch
+    # Not available in this Oceananigans version - define dummy type
+    abstract type TimeSeriesInterpolation end
+end
```
glwagner, line 17: This should be removed

Project.toml
```diff
@@ -42,24 +42,25 @@ CondaPkg = "992eb4ea-22a4-4c89-a5bb-47a3300528ab"
...
 Reactant = "3c362404-f566-11ee-1572-e11a4b42c853"
 SpeedyWeather = "9e226e20-d153-4fed-8a5b-493def4f21a9"
 WorldOceanAtlasTools = "04f20302-f1b9-11e8-29d9-7d841cb0a64a"
-Zarr = "0a941bbe-ad1d-11e8-39d9-ab76183a1d99"
 
 [extensions]
 NumericalEarthArchGDALExt = "ArchGDAL"
 NumericalEarthBreezeExt = "Breeze"
 NumericalEarthCDSAPIExt = "CDSAPI"
+NumericalEarthCDSClientExt = ["HTTP", "JSON3"]
```
glwagner: This is possibly problematic because it produces surprising behavior; ie the extension for CDSClient is loaded when we write `using JSON3, HTTP`. 

I'm wondering if it is possible to add this functionality instead to the `CopernicusClimateDataStore.jl` package? Is that what we need?

glwagner:
Do we know why two extensions are needed, rather than just the single `NumericalEarthCopernicusClimateDataStoreExt` ? I'm also curious what this PR adds on top of what we put into https://github.com/NumericalEarth/NumericalEarth.jl/pull/364

aklocker42:
Thanks for the review Greg!

## Re: What this adds beyond #364

PR #364's `CopernicusClimateDataStore` extension still requires Python under the hood - it uses `CopernicusClimateDataStore.jl`, which installs `era5cli` via `CondaPkg.add_pip()` and shells out to Python CLI.

This PR adds a **truly pure Julia client** - direct HTTP API calls to CDS using HTTP.jl + JSON3.jl. No Python, no CondaPkg, no shell commands. But you're right that having two separate extensions is confusing.

## Better idea: Just add it to CopernicusClimateDataStore.jl

Yeah, you're right that the HTTP/JSON3 weak deps approach is problematic. Loading the extension just because someone does `using HTTP` somewhere is definitely surprising behavior.

After thinking about it, I agree the pure Julia client should just live in `CopernicusClimateDataStore.jl` directly. That way:

**Proposed architecture:**
- **Pure Julia (HTTP/JSON3) as primary** - try this first, eliminates Python for most users
- **era5cli as fallback** - only if pure Julia fails or user opts in
- User just does `using CopernicusClimateDataStore` and it works

Something like:
```julia
function download_era5(...)
    # Try pure Julia first
    try
        return download_via_cds_api(...)  # HTTP + JSON3
    catch e
        @warn "Falling back to era5cli..." exception=e
        return download_via_era5cli(...)  # CondaPkg + Python
    end
end
```

**Why this is cleaner:**
- Single extension in NumericalEarth (just the CopernicusClimateDataStore one)
- No surprising extension loading
- Python-free by default
- Still have era5cli available if needed
- Download method choice lives where it belongs

**Plan:**
- Close/draft this PR
- Open PR in CopernicusClimateDataStore.jl with pure Julia client
- Once merged, NumericalEarth just uses the enhanced package

Sound good?

## Quick answers on the inline comments:

**1. Extension exports** - Moot if we move to CopernicusClimateDataStore.jl 

**2. ERA5_cds_client.jl location** - Yeah, should move to CopernicusClimateDataStore.jl/src/

**3. ERA5_variables.jl include** - Also belongs in CopernicusClimateDataStore.jl (CDS API details)

**4. pressure_level try/catch** - Agreed, will remove. Was just a temporary workaround.

---

Let me know if this approach makes sense and I'll move forward with the CopernicusClimateDataStore.jl PR!

glwagner:
Since we have the "python path" through CDSAPI, I think we can just eliminate python completely from `CopernicusClimateDataStore` !

aklocker42:
@glwagner Ah got it - so users who want Python can just use `cdsapi` directly, no need for NumericalEarth.jl to wrap it. Makes sense!

I'll make these changes to eliminate Python completely:

**Changes:**
1. **Remove Python dependency** - delete `era5cli` wrapper code
2. **Make HTTP/JSON3 required dependencies** - move from `[weakdeps]` to `[deps]`
3. **Simplify `CopernicusClimateDataStore` module** - only the pure Julia CDS client
4. **Update docs** - users configure `~/.cdsapirc` (standard CDS approach)

Ok?

glwagner:
These changes need to be made on the `CopernicusClimateDataStore` package right? 

https://github.com/NumericalEarth/CopernicusClimateDataStore.jl

we can move the discussion over to there (or just open a PR there)

> Ah got it - so users who want Python can just use cdsapi directly, no need for NumericalEarth.jl to wrap it.

Almost, but we still need to have an extension in NumericalEarth to `CDSAPI.jl`. NumericalEarth does not wrap the python itself. `CDSAPI.jl` wraps the python. Unless I am confused.

aklocker42:
# PR #380 Update: Addressed all reviewer feedback

## Changes made to address @glwagner's comments

### ✅ 1. Extension loading behavior
**Fixed**: Pure Julia CDS client moved to `CopernicusClimateDataStore.jl` package with required dependencies (not weak deps).

### ✅ 2. File location  
**Fixed**: Removed `ERA5_cds_client.jl` from NumericalEarth source tree. Implementation now lives in `CopernicusClimateDataStore.jl`.

### ✅ 3. Missing module inclusion
**Fixed**: Added `include("ERA5_variables.jl")` to `ERA5.jl` module.

### ❓ 4. Unnecessary compatibility code
**Not found**: The try/catch for TimeSeriesInterpolation compatibility doesn't exist in the `fix-era5-download-method` branch. May have been already removed or in a different branch.

### ✅ 5. Weak dependencies architecture
**Fixed**:
- Removed `NumericalEarthCDSClientExt` (direct HTTP/JSON3 implementation)
- Removed HTTP and JSON3 from weak dependencies
- Pure Julia path now via `NumericalEarthCopernicusClimateDataStoreExt`

### ✅ 6. Architectural scope
**Fixed**: Pure Julia CDS client contributed to `CopernicusClimateDataStore.jl` package.

## New architecture

Users now have two clean options:

**Option 1: Python wrapper (optional)**
```julia
using CDSAPI  # Triggers NumericalEarthCDSAPIExt
using NumericalEarth
```

**Option 2: Pure Julia (recommended)**
```julia
using CopernicusClimateDataStore  # Triggers NumericalEarthCopernicusClimateDataStoreExt
using NumericalEarth
```

## Related PR

**CopernicusClimateDataStore.jl#5**: Pure Julia CDS API v2 client
- Link: https://github.com/NumericalEarth/CopernicusClimateDataStore.jl/pull/5
- Implements HTTP/JSON3-based download functionality
- Makes HTTP, JSON3, Downloads, Dates, Base64 required dependencies
- Eliminates Python/CondaPkg requirement for CDS downloads
- Tested and working with CDS API v2
## Files changed in this update

**NumericalEarth.jl (`fix-era5-download-method` branch):**
- ✅ `src/DataWrangling/ERA5/ERA5.jl`: Added ERA5_variables.jl include
- ✅ `src/DataWrangling/ERA5/ERA5_cds_client.jl`: **Deleted** (moved to CopernicusClimateDataStore.jl)
- ✅ `ext/NumericalEarthCDSClientExt.jl`: **Deleted** (redundant with CopernicusClimateDataStore.jl extension)
- ✅ `Project.toml`: Removed HTTP/JSON3 from weakdeps; kept CDSAPI and CopernicusClimateDataStore

**Summary**: -295 lines (cleanup), +1 line (ERA5_variables.jl include)

## Testing status

- ✅ CopernicusClimateDataStore.jl pure Julia client tested and working
- ⏳ NumericalEarth extension integration needs testing (requires CopernicusClimateDataStore.jl PR merge first)

src/DataWrangling/ERA5/ERA5.jl
```diff
@@ -132,13 +132,23 @@ end
 
 DataWrangling.inpainted_metadata_path(metadata::ERA5Metadatum) = joinpath(metadata.dir, inpainted_metadata_filename(metadata))
 
+#####
+##### Pure Julia CDS client (replaces Python era5cli)
+#####
+
+# CDS client is loaded via NumericalEarthCDSClientExt extension when HTTP/JSON3 are available
+# The extension exports: download_era5, read_cds_credentials, cds_variable_name
+# Users need to: using HTTP, JSON3  (or just using HTTP if JSON3 is already loaded)
```
glwagner, line 141: I don't think this will be true any longer, since `HTTP` and `JSON3` are dependencies now of `CopernicusClimateDataStore.jl`

src/Grids/pressure_level_vertical_discretization.jl
glwagner, line 17: I think we have to omit this for the PR... if we are missing a compat we can add that, so that `Oceananigans.OutputReaders.TimeSeriesInterpolation` is guaranteed to exist

PR #381, opened 2026-06-24
Fix sea-ice fluxes

simone-silvestri:
The sea ice model requires per-cell fluxes, not per-ice fluxes.
For this reason the fluxes need to be multiplied by concentration.
The only flux that does not need multiplication by concentration is the frazil ice flux which is independent of the ice state


glwagner (approved):

glwagner:
This is also what the ocean model wants right? So the total flux is `total = ice + ocean` now?

PR #383, opened 2026-06-25
Fix radiative forcing for a zstar grid

simone-silvestri:
The formulation was not correct because it assumed that `z == 0` at the surface. 
However, for a zstar grid, especially when we have a very small surface grid cell (for example 1m), `z` at `Nz` might be positive, therefore the radiation computed with `exp(\kappa z)` amplifies instead of reducing the incoming solar flux and the simulation explodes.

This PR fixes the penetrative radiation

### DIFF
--- a/src/Oceans/radiative_forcing.jl
+++ b/src/Oceans/radiative_forcing.jl
@@ -45,9 +45,12 @@ end
 const c = Center()
 const f = Face()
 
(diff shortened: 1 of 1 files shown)


glwagner (approved):

glwagner:
simple

PR #386, opened 2026-06-26
PrescribedAtmosphere refactor + ERA5PrescribedAtmosphere

### LINKED ISSUE #266: Refactor PrescribedAtmosphere: notation, microphysical variables, surface-vs-nesting role
`PrescribedAtmosphere` (`src/Atmospheres/prescribed_atmosphere.jl`) needs a refactor that combines three related threads. Deferred out of the NestedSimulations work on `eq/era5_breeze_land` so the v1 nesting PR can stay scoped.

## 1. Proper notation

Today's struct uses ad-hoc names: `velocities.u`, `tracers.T`, `tracers.q`, etc. These should align with the convention used elsewhere in the codebase (see `examples/era5_breeze.jl`): `u`, `v`, `θˡⁱ`, `qᵗ`, `qᶜ`, `qⁱ`, `ρ`.

## 2. Microphysical variables

`tracers` is currently `(T, q)`. To support Breeze's prognostic moisture partition and downstream microphysics coupling, it should hold at minimum:

- `qᵗ` — total water mixing ratio
- `qᶜ` — cloud liquid mixing ratio
- `qⁱ` — cloud ice mixing ratio

And potentially precipitating species when those land.

## 3. Surface-forcing vs 3D-nesting role distinction

The struct carries surface-only concepts (`pressure`, `freshwater_flux`, `surface_layer_height`, `boundary_layer_height`) that are meaningless when `PrescribedAtmosphere` is used as a 3D parent for limited-area nesting.

Two paths considered (decision postponed to this refactor):

- **Nullable per-field defaults driven by grid dimensionality** — a 3D grid defaults the surface-only fields to `nothing`; user can override. More flexible; leaves design space open.
- **`mode` / `role` kwarg** — `SurfaceForcing` vs `NestingParent` selects defaults. More self-documenting; locks in a named enum.

Leaning is toward the first, but worth revisiting with fresh eyes when this refactor happens.

## Why deferred

The nesting work needs 3D `PrescribedAtmosphere` *now*, but a wholesale rename + tracer-set expansion would otherwise gate that PR. v1 of NestedSimulations adds 3D constructor variants only; this refactor lands separately so it's the only change in its PR.

## Touch points (non-exhaustive)

- `src/Atmospheres/prescribed_atmosphere.jl`
- `src/Atmospheres/prescribed_atmosphere_regridder.jl`
- `src/Atmospheres/interpolate_atmospheric_state.jl`
- Surface-flux callers under `src/EarthSystemModels/InterfaceComputations/`
- Any JRA55 / ERA5 builders that construct `PrescribedAtmosphere`

glwagner:
Closes #266 (PrescribedAtmosphere notation / microphysical variables / surface-vs-nesting role) and adds a native-grid 3D ERA5 parent constructor for nesting. Builds on the `NestedSimulations` work from #233 (now on `main`).

## What's in here

**1. `PrescribedAtmosphere` redesign (#266).** `tracers=(T,q)` → direct **`temperature`** and **`specific_humidity`** fields; new **`microphysical_variables`** container (cloud/precip species, e.g. `(; qᶜˡ, qʳ, qᶜⁱ, qˢ)`, empty by default); `tracers` reserved for gas species (empty by default). Retires the `two_dimensional` kwarg — surface (2D, z-`Nothing`) vs volumetric (3D, z-`Center`, +`w`) default fields are inferred from the grid's vertical topology (`Flat` ⇒ surface forcing; resolved ⇒ a `NestedSimulation` parent).
  - Ocean/sea-ice coupling is **unchanged**: flux kernels read the exchanger state (`.T/.q/.u/.v`), not the struct; and the dataset builders pass fields explicitly, so the default-construction change can't affect them.
  - Updated: the interpolate hub, the ERA5/JRA55/ECCO/OSPapa builders, and the affected tests.

**2. `freshwater_flux` → `precipitation_flux`.** The atmosphere field/kwarg is renamed to what it holds (a `PrescribedPrecipitationFlux`), disambiguating it from the **ocean**'s `net_ocean_freshwater_flux` budget and the **land**'s `freshwater_flux` (river runoff) — both untouched. Its default is now grid-dispatched: a **3D atmosphere defaults to none** (it carries precipitation in `microphysical_variables`), while a **surface atmosphere** keeps the 2D `PrescribedPrecipitationFlux`. So `precipitation_flux = nothing` is no longer needed for ERA5/nested parents.

**3. `ERA5PrescribedAtmosphere(bounding_box, dates)`.** A 3D parent built from ERA5 **pressure-level** reanalysis on ERA5's **native grid** (a `PressureLevelGrid`, geopotential-height aware) — no pre-regridding; a downstream child interpolates onto its own grid on the fly. Holds `velocities`, `temperature`, `specific_humidity`, `microphysical_variables`, and `pressure` (the level coordinate). Because pressure is the native level value, on-the-fly **log-space** interpolation of it (in the follow-up nesting PR) is faithful ln(p) interpolation of ERA5.

**4. Notation cleanup** in `prescribed_atmosphere.jl` and the ERA5/JRA55 builders: shadow locals (`ua/va/wa/Ta/qa/pa/Fr/ff`) → canonical (`u/v/w/T/qᵛ/p`), shared loader closures (`single_level` / `pressure_level` / `jra55_field`), and no mixed verbose+abbrev identifiers (`velocity_bcs`, `north_bc`). Also completes the #266 migration of leftover `atmosphere.tracers.T/.q` accesses in examples/docs/tests.

**5. `examples/breeze_downscaling_era5.jl`** (renamed from `era5_breeze.jl`) — adapted to the new struct; runs the MC3E case at ~12 km over a 2×-expanded domain, **Float32**, 12 h.

## Validation

Loads/precompiles in a dev env; the #266 grid-dispatched layout and the `precipitation_flux` default-by-dimensionality are covered by `test_nested_simulation.jl`. The 3D `ERA5PrescribedAtmosphere(bbox, dates)` method precompiles with the correct signature; the full ERA5 ingest needs CDS data (exercised by the example). A 6 h GPU baseline of the example runs cleanly (all fields finite). Full test suite runs in CI.

## Scope / follow-ups

This is the struct + constructor refactor only. Deferred to the **next (nesting-BC) PR**:
- The transform-aware **`interpolate(f, …)`** shim (log-space interpolation of `p`/`ρ`, FieldTimeSeries-capable) — being upstreamed to Oceananigans (#5726); removed from this PR.
- The **on-the-fly multi-field boundary condition** that derives Breeze prognostics (`ρ, ρθˡⁱ, ρqᵉ`) from raw parent state, eliminating `breeze_prognostic_state`.
- Wiring the example to `ERA5PrescribedAtmosphere(grid)` instead of the hand-rolled parent.


src/DataWrangling/ERA5/ERA5_prescribed_atmosphere.jl
```diff
@@ -93,10 +95,82 @@ function ERA5PrescribedAtmosphere(architecture = CPU();
...
+    pressure_level(name) = FieldTimeSeries(Metadata(name; dataset, dates, region, dir), architecture; kw...)
+
+    ua  = pressure_level(:eastward_velocity)
+    va  = pressure_level(:northward_velocity)
+    Ta  = pressure_level(:temperature)
+    qva = pressure_level(:specific_humidity)
+    qcl = pressure_level(:specific_cloud_liquid_water_content)
+    qrw = pressure_level(:specific_rain_water_content)
+    qci = pressure_level(:specific_cloud_ice_water_content)
+    qsw = pressure_level(:specific_snow_water_content)
```
glwagner: let's clean up notation here to match Breeze/NumericalEarth.

Also "pressure_level(name)" doesn't read intuitively to me.

src/EarthSystemModels/NestedSimulations/transformed_interpolate.jl
```diff
@@ -0,0 +1,95 @@
...
+using Oceananigans.Fields: FractionalIndices, flatten_node, interpolator, _interpolate
+using Oceananigans.Grids: topology, Flat
+using Oceananigans.OutputReaders: FlavorOfFTS, TimeInterpolator, memory_index
+
+# Inverse-transform trait: the inverse applied to the (transform-space) blend. Defined only for
+# transforms with an exact analytic inverse; an unlisted `f` is a deliberate "no inverse known".
+@inline inverse_transform(::typeof(identity)) = identity
+@inline inverse_transform(::typeof(log))      = exp
+@inline inverse_transform(::typeof(log2))     = exp2
+@inline inverse_transform(::typeof(log10))    = exp10
```
glwagner: these should be deleted since we don't need to intrinsically support the inverse transform

examples/breeze_downscaling_era5.jl
```diff
@@ -395,7 +392,7 @@ parent_grid = LatitudeLongitudeGrid(arch;
 # and replace the code below
 
 parent_times = [Float64(Dates.value(d - start_date)) / 1000 for d in dates]
-parent = PrescribedAtmosphere(parent_grid, parent_times; two_dimensional = false, freshwater_flux = nothing, thermodynamics_parameters = nothing)
+parent = PrescribedAtmosphere(parent_grid, parent_times; freshwater_flux = nothing, thermodynamics_parameters = nothing)
```
glwagner: I think `PrescribedAtmosphere` does not need `freshwater_flux` anymore, because this is part of `PrescribedLand` ?

glwagner: I think it's still needed and is a distinct flux from the land one. `PrescribedAtmosphere.freshwater_flux` is the atmospheric *precipitation* (rain/snow): `interpolate_atmospheric_state.jl` reads it via `surface_rainfall_flux`/`surface_snowfall_flux` and writes `Jʳⁿ`/`Jˢⁿ` into the exchanger state, which the ocean/sea-ice coupling consumes for the surface freshwater flux. `PrescribedLand.freshwater_flux` is a separate source — river runoff / icebergs (land→ocean, assembled in `assemble_net_ocean_fluxes.jl`). So both coexist. Here the parent passes `freshwater_flux = nothing` only because it's a 3D nesting parent with no surface coupling. (Happy to revisit if we want to move precipitation out of the atmosphere, but that's a separate change — it's currently consumed.)

glwagner: Now I am thinking we want to change `freshwater_flux` to `precipitation` (less confusing) or even absorb into `microphysical_variables`

kaiyuan-cheng: `freshwater_flux ` indeed sounds ambiguous in the context of atmosphere

src/Atmospheres/prescribed_atmosphere.jl
```diff
@@ -2,10 +2,13 @@
...
 #####
 
-mutable struct PrescribedAtmosphere{FT, G, T, U, P, C, F, TP, TI} <: AbstractPrescribedComponent
+mutable struct PrescribedAtmosphere{FT, G, T, U, Θ, Q, M, P, C, F, TP, TI} <: AbstractPrescribedComponent
     grid :: G
     clock :: Clock{T}
     velocities :: U
+    temperature :: Θ
+    specific_humidity :: Q
+    microphysical_variables :: M
```
kaiyuan-cheng, line 12: Maybe just `hydrometeors` ?

glwagner: As currently written `microphysical_variables` also could hold cloud liquid and cloud ice (where applicable).

We could have separate properties for cloud variables and hydrometeors though. What do you think?

examples/breeze_downscaling_era5.jl
glwagner: I actually changed it to `precipitation_flux`. More specifically, this is the precipitation flux at the surface (perhaps `surface_precipitation_flux` would be even better?)

src/Atmospheres/prescribed_atmosphere.jl
kaiyuan-cheng: I think cloud droplets and ices are part of hydrometeors, am I wrong?

src/Atmospheres/prescribed_atmosphere.jl
```diff
@@ -2,76 +2,89 @@
...
-    if two_dimensional
-        velocity_bcs = velocity_boundary_conditions(grid, (Center(), Center(), nothing))
-        ua = FieldTimeSeries{Center, Center, Nothing}(grid, times; boundary_conditions = velocity_bcs)
-        va = FieldTimeSeries{Center, Center, Nothing}(grid, times; boundary_conditions = velocity_bcs)
-        return (u=ua, v=va)
+# Surface (2D, z-`Nothing`) vs volumetric (3D, z-`Center`) defaults are inferred from the grid:
+# a `Flat` vertical ⇒ surface forcing (ocean / sea-ice coupling); a resolved vertical ⇒ a 3D
+# atmosphere (e.g. a `NestedSimulation` parent). Dataset builders pass their own fields explicitly
+# and so are unaffected; only the `default_*` paths consult this. Override any field via kwarg.
+@inline volumetric_atmosphere(grid) = topology(grid, 3) !== Flat
```
kaiyuan-cheng: 'volumetric' sounds weird. I think `atmosphere(grid)` alone works.

simone-silvestri (approved):

kaiyuan-cheng (approved):

src/Atmospheres/prescribed_atmosphere.jl
glwagner: you would know better 😂 I assumed that a "hydrometeor" is sedimenting, but we model "cloud" properties as non-sedimenting (even though this is technically wrong, all condensates are sedimenting)

Another term is "condensates" which covers all liquid and ice.

Another difference is that `microphysical_variables` can include things like higher moments (number concentrations, etc). However I would classify `condensates` as simply the mass fractions. But I am not sure where that leaves hydrometeors.

src/Atmospheres/prescribed_atmosphere.jl
glwagner: ah nice catch. This function actually has nothing to do with atmospheres at all! I think perhaps `is_three_dimensional(grid)` is what we want

glwagner: Renamed to `is_three_dimensional(grid)` — matches the existing `is_three_dimensional(::Metadata)` predicate convention in `DataWrangling`. Kept it a predicate name rather than `atmosphere(grid)`, since the function returns a `Bool` (`topology(grid, 3) !== Flat`) and `atmosphere(grid)` reads like it constructs an atmosphere. (3c181889)

PR #387, opened 2026-06-26
On-the-fly nesting BCs: ParentStateBoundary + #5726-aligned interpolate (WIP)

glwagner:
**Draft / WIP — stacked on #386** (base is `glw/prescribed-atmosphere-refactor`; rebase onto `main` once #386 merges).

Builds the on-the-fly nesting boundary machinery that derives a child model's prognostics from raw parent state at fill time — eliminating the materialized density-weighted FTS bundle and the per-snapshot `breeze_prognostic_state` precompute.

## In this draft (committed + smoke-validated on CPU)

- **`breeze_prognostic_state` factored** into reusable pointwise maps (`air_density`, `liquid_ice_potential_temperature`, `total_water_specific_humidity`, …) — one definition, used both broadcast over `Field`s and per boundary-face node. No behavior change.
- **`ParentStateBoundary`** (internal carrier): holds a `NamedTuple` of parent-state `FieldTimeSeries` + per-source interpolation transforms (log for `p`/`T`, identity otherwise) + a prognostic map. `getbc` interpolates each source at the boundary-face node and applies the map. Interpolation uses the #5726-style `interpolate(func,…)` (func-space) and un-maps locally via `inverse_transform`, so log-space pressure = `exp(blend of log p)` (faithful ln-p). Smokes: a 2-field transform reconstructs the right face values; a log-space source recovers the geometric interpolation.

> The transform-interpolate shim (`transformed_interpolate.jl`, on #386) now mirrors **CliMA/Oceananigans.jl#5726** (`MappedData` / func-space / caller-applies-inverse). When #5726 releases, that file is deleted and the calls hit upstream verbatim.

## Still to come (this PR will grow)

- Breeze-ext prognostic maps as transforms (`ρ`, `ρu`, `ρθˡⁱ`, `ρqᵉ`) + the model hook `parent_boundary_conditions(grid, parent, ::Type{AtmosphereModel})`.
- The public factory `atmosphere_simulation(grid; parent=…)` dispatching to an internal `nested_atmosphere_simulation(grid, parent; …)` that augments BCs + Davies relaxation from the parent and delegates back to the plain factory (so the default simulation config lives in one place).
- GPU-path validation of the carrier `getbc` (CPU validated here; the kernel path needs a GPU node).

### DIFF
--- a/ext/NumericalEarthBreezeExt/breeze_prognostic_state.jl
+++ b/ext/NumericalEarthBreezeExt/breeze_prognostic_state.jl
@@ -1,9 +1,29 @@
 #####
-##### breeze_prognostic_state: moist thermodynamic state → Breeze prognostics
+##### Breeze prognostic maps: moist thermodynamic state → CompressibleDynamics prognostics
 #####
+#
+# Pointwise `@inline` maps from a moist thermodynamic state (temperature `T`, vapor `qᵛ`, cloud
+# liquid `qᶜˡ`, cloud ice `qᶜⁱ`, pressure `p`) to the variables Breeze integrates. Written on bare
+# scalars + gas-constant/heat-capacity/latent-heat numbers so they are GPU-safe and reusable both
+# broadcast over `Field`s (`breeze_prognostic_state`) and per boundary-face node (the on-the-fly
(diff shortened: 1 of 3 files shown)


glwagner:
(didn't mean to open)

PR #388, opened 2026-06-27
On-the-fly nested-atmosphere boundary conditions + Davies relaxation

glwagner:
Follow-up to #386. Derives a Breeze nested child's lateral boundary conditions **and** interior Davies relaxation **on the fly** from a coarse parent atmosphere's *raw* state — eliminating the materialized `breeze_prognostic_state` parent series. Builds on the `ERA5PrescribedAtmosphere` / `PrescribedAtmosphere` refactor from #386 (now on `main`).

## The abstraction

- **`ParentStateBoundary`** (`NestedSimulations`) — a multi-field boundary condition: at each boundary-face node it interpolates a `NamedTuple` of parent `FieldTimeSeries` (each with its own interpolation transform — `log` for the strictly-positive `p`/`T`, `identity` otherwise) and applies a pointwise `transform` to the resulting state. Carrier-only; knows nothing about which prognostic or dataset.
- **`ParentStateTarget`** — the interior-relaxation analogue: a `Relaxation` `target(x, y, z, t)` using the same `_query_source` machinery at an arbitrary node.
- Both use Oceananigans' **`interpolate(f, …)`** (CliMA/Oceananigans.jl#5726) for the log-space interpolation; the local shim is gone.

## Breeze wiring (in `NumericalEarthBreezeExt`)

- **Prognostic transforms** (`isbits` callables over the pointwise maps): `AirDensity`, `MomentumDensity`, `EnergyDensity`, `MoistureDensity` (density-weighted, for the BCs); `LiquidIcePotentialTemperature`, `TotalWater` (specific, for the relaxation).
- **`atmosphere_simulation(grid; parent_atmosphere, relaxation_rate, relaxation_mask)`**:
  - lateral BCs for `ρ, ρu, ρv, ρe, <moisture>` via `ParentStateBoundary` (ρu/ρv → `NormalFlowBC`, the rest → `ValueBC`), merged per-side with the coupling bottom-flux BCs;
  - interior Davies relaxation keyed `(u, v, θ, qᵉ)` for Breeze's `SpecificForcing` (u/v toward the raw parent velocities, θ/qᵉ toward on-the-fly specific targets).
  - The `parent` kwarg was named `parent_atmosphere` to avoid shadowing `Base.parent` / Oceananigans `parent(field)`.

Workflow: **build parent → `atmosphere_simulation(grid; parent_atmosphere)` → `NestedModel`**.

## Dependency

`interpolate(f, …)` is merged to Oceananigans main but not yet tagged, so this PR points Oceananigans at **main** via `[sources]` (root/test/docs). Swap to a `[compat]` bump once a release is tagged (TODOs in place).

## Status / TODO (draft)

- [ ] Rewrite `examples/breeze_downscaling_era5.jl` to use `atmosphere_simulation(grid; parent_atmosphere, relaxation_rate, …)`, deleting the `parent_series` materialization loop + single-source `parent_boundary_conditions`.
- [ ] Tests for `ParentStateBoundary` / `ParentStateTarget` (CDS-free, toy parent FTS).
- [ ] Benchmark the on-the-fly cost (vs. a precomputed parent series) — precomputing is a possible later optimization (would move the conversion into parent construction).
- Not yet exercised locally (dev env can't precompile here); relying on CI against Oceananigans main.


docs/Project.toml
```diff
@@ -6,16 +6,20 @@ CFTime = "179af706-886a-5703-950a-314cd64e0468"
...
 DataDeps = "124859b0-ceae-595e-8997-d05f6a7a8dfe"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 DocumenterCitations = "daee34ce-89f3-4625-b898-19384cb65244"
+GeoInterface = "cf35fbd7-0cd7-5166-be24-54bfbe79505f"
+JLD2 = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
 Literate = "98b081ad-f1c9-55d3-8b20-4c87d4299306"
 LiveServer = "16fef848-5104-11e9-1b77-fb7a48bbb589"
 NCDatasets = "85f8d34a-cbdd-5861-8df4-14fed0d494ab"
+NaturalEarth = "436b0209-26ab-4e65-94a9-6526d86fea76"
```
ewquon, line 23: 🤪

examples/breeze_downscaling_era5.jl
```diff
@@ -1,973 +1,307 @@
...
-#   cycle count — the deep updrafts are CAPE-driven at grey-zone resolution, not seeded by the
-#   initialization transient. A single balance cycle therefore suffices (`balance_cycles = 1`).
-# - Near-surface initialization transient: the ERA5 winds are set on the terrain-following grid as-is,
-#   in balance with neither surface drag nor the model's pressure field. The lowest cell samples the
-#   free-atmosphere ERA5 wind — strongest over high terrain, where k=1 sits ~1 km above sea level — so
-#   the near-surface |U| sheds ~25% in the first ~0.5 h as surface drag spins up and the flow
-#   geostrophically adjusts (DFI balances ρw, not the horizontal momentum). A balanced / terrain-aware
-#   wind initialization, or a DFI that also balances horizontal momentum, would reduce it; motivates
-#   further development of the initialization routine.
+# - Single nest only (ERA5 → 12 km; coarsened 4× from Fan's 3 km Domain 3 for a fast configuration).
```
ewquon, line 23:
```suggestion
# - Single nest only (ERA5 → 12 km; coarsened 4× from Fan's 3 km Domain 3 for a fast configuration).
# - No map projection (the child is a spherical `LatitudeLongitudeGrid`) — for projected constant-Δx/Δy,
#   `LambertConformalConicGrid` will be available in a future Oceananigans release.
```

ewquon:
Nice overhaul of the `breeze_downscaling_era5.jl` example!

kaiyuan-cheng:
Shall we still keep the `ParentBoundary` approach? It is not used.

examples/breeze_downscaling_era5.jl
glwagner: actually the lambert grid is currently supported. should we switch to that?

docs/Project.toml
glwagner: Natural \ne Numerical

glwagner:
> Shall we still keep the `ParentBoundary` approach? It is not used.

I will look at this

ext/NumericalEarthBreezeExt/breeze_nested_atmosphere.jl
```diff
@@ -0,0 +1,306 @@
...
+    λ₁, λ₂ = extrema(λnodes(grid, Face(), Center(), Center()))
+    φ₁, φ₂ = extrema(φnodes(grid, Center(), Face(), Center()))
+    Nx, Ny, _ = size(grid)
+    w = width * max((λ₂ - λ₁) / Nx, (φ₂ - φ₁) / Ny)
+    return (λ, φ, z) -> begin
+        d = min(λ - λ₁, λ₂ - λ, φ - φ₁, φ₂ - φ)
+        d >= w && return zero(λ)
+        return oftype(λ, (1 + cos(π * d / w)) / 2)
+    end
+end
```
ewquon:
```suggestion
# Ramp shapes (isbits callables) for a nudging zone: weight vs. normalized distance from the wall, s ∈ [0, 1].
# Contract: ramp(0)=1, ramp(1)=0, monotone between.
struct CosineRamp end
struct SmoothStepRamp end

@inline (::CosineRamp)(s)     = (1 + cos(π * s)) / 2
@inline (::SmoothStepRamp)(s) = 1 - s^2 * (3 - 2s)

# Davies mask: 1 at the lateral walls, ramping to 0 over the outermost `width` cells.
function davies_relaxation_mask(grid, width; ramp = CosineRamp())
    λ₁, λ₂ = extrema(λnodes(grid, Face(), Center(), Center()))
    φ₁, φ₂ = extrema(φnodes(grid, Center(), Face(), Center()))
    Nx, Ny, _ = size(grid)
    w = width * max((λ₂ - λ₁) / Nx, (φ₂ - φ₁) / Ny)
    return (λ, φ, z) -> begin
        d = min(λ - λ₁, λ₂ - λ, φ - φ₁, φ₂ - φ)
        s = clamp(d / w, zero(d), one(d))
        return oftype(d, ramp(s))
    end
end
```
Is now a good time to change the ramping function to something more computationally efficient?

ext/NumericalEarthBreezeExt/breeze_nested_atmosphere.jl
```diff
@@ -0,0 +1,306 @@
...
+
+# Cubic-ramp (smoothstep) Rayleigh mask over the top `depth` metres of the domain, for the ρw lid sponge.
+# `z_top` is read once host-side under `@allowscalar`: a `znode` on a terrain-following GPU grid
+# indexes the (device) terrain arrays, which is otherwise disallowed.
+function lid_sponge_mask(grid, depth)
+    z_top = @allowscalar znode(1, 1, size(grid, 3) + 1, grid, Center(), Center(), Face())
+    d = convert(eltype(grid), depth)
+    return (λ, φ, z) -> (s = clamp((z - (z_top - d)) / d, zero(z), one(z)); s * s * (3 - 2s))
+end
```
ewquon:
```suggestion
function lid_sponge_mask(grid, depth; ramp = SmoothStepRamp())
    z_top = @allowscalar znode(1, 1, size(grid, 3) + 1, grid, Center(), Center(), Face())
    d = convert(eltype(grid), depth)
    return (λ, φ, z) -> (s = clamp((z - (z_top - d)) / d, zero(z), one(z)); ramp(s))
end
```
with the change above

ERA5_BREEZE_LONG_RUN_PLAN.md
```diff
@@ -0,0 +1,94 @@
...
+  `ρᵈ, ρθ, ρu, ρv, <moisture>` toward the parent over a cosine ramp across the outermost 5 cells
+  (matches the WRF/MPAS mass-nudging precedent; `ρw` is *not* relaxed, also per precedent).
+- **Lid sponge:** `damping_depth = default_lid_depth(grid) = Lz/4` (~5 km for a ~19.5 km column, base
+  ~15 km) — thin enough to leave the 12–16 km convective layer undamped, deep enough to absorb genuine
+  top reflection now that the IC fix removes the wall-mode source. Lifts both the split-explicit
+  `UpperSponge` and the `ρw` Rayleigh lid.
+- **Advection (implicit LES, no explicit closure):** `momentum_advection = WENO(order=5)`; per-tracer
+  `scalar_advection = (ρθ = WENO(5), ρqᵉ = WENO(5, bounds=(0,1)), ρqʳ = …, ρqˢ = …)` — moisture/precip
+  are positivity-bounded, `ρθ` must **not** be bounded to `[0,1]`; `closure = nothing`.
+- **Time stepping:** adaptive `conjure_time_step_wizard!(cfl = 0.7, max_Δt = 20 s)` on the (slower)
```
ewquon, line 27: max dt = 20 s is pretty conservative (small) for dx ~ 12 km

examples/breeze_downscaling_era5.jl
ewquon: This would motivate a new end-to-end run with a few untested features. In the interest of keeping the scope tight, perhaps we defer?

ext/NumericalEarthBreezeExt/breeze_state_exchanger.jl
```diff
@@ -0,0 +1,199 @@
...
+#####
+#
+# The child (Breeze `CompressibleDynamics`) prognostic variables — dry density `ρᵈ`, momentum densities
+# `ρu`/`ρv`, potential-temperature density `ρθ`, and vapor density `ρqᵛ` — are computed from the raw
+# parent (ERA5) specific state *on the parent grid* and stored as a `FieldTimeSeries` holding just the
+# **two time levels bracketing the child's clock** (memory-O(1) in time). A downstream child boundary
+# condition / forcing then interpolates these precomputed prognostics in space + time. Computing the
+# nonlinear combines on the dense parent grid first (then interpolating) is both cheaper — once per
+# parent time level rather than per child node per RK stage — and more faithful than interpolating the
+# raw fields and combining afterward.
```
ewquon: 👍

ext/NumericalEarthBreezeExt/breeze_state_exchanger.jl
```diff
@@ -0,0 +1,199 @@
...
+#
+# Density weighting matches Breeze's `establish_densities!`/`set!` (dry density is the prognostic):
+#   ρ   = p / (Rᵈ Tᵥ)            (total moist density),   Tᵥ = T (1 + (Rᵛ/Rᵈ − 1) qᵛ)
+#   qᵗ  = qᵛ + qᶜˡ + qᶜⁱ
+#   ρᵈ  = ρ (1 − qᵗ)                                    ← the prognostic (dry) density
+#   ρθ  = ρᵈ · θˡⁱ,   ρu = ρᵈ · u,   ρv = ρᵈ · v         ← DRY-weighted (energy + momentum)
+#   ρqᵛ = ρ · qᵛ                                         ← TOTAL-weighted (moisture mass density)
```
ewquon, line 20: Inconsistent use of ρ vs ρᵈ here.

ext/NumericalEarthBreezeExt/breeze_state_exchanger.jl
```diff
@@ -0,0 +1,199 @@
...
+
+    # Bracketing indices for `time` on the parent's time axis; cycle the derived window if it moved.
+    _, n₁, _ = interpolating_time_indices(p.ρᵈ.time_indexing, p.ρᵈ.times, time)
+    if p.ρᵈ.backend.start != n₁
+        for fts in p
+            fts.backend = new_backend(fts.backend, n₁, length(fts.backend))
+        end
+    end
+
+    compute_child_prognostics!(p, parent, ex.pˢᵗ, ex.constants, ex.condensates)
```
ewquon: Recompute happens regardless of whether the derived window moved or not (contrary to the function descriptino)

src/NestedModels/nested_model.jl
```diff
@@ -0,0 +1,100 @@
...
+    return update_state!(nm.child, callbacks; kwargs...)
+end
+
+# Setting the initial state targets the child's prognostics.
+set!(nm::NestedModel, args...; kwargs...) = set!(nm.child, args...; kwargs...)
+
+# Ordering: refresh the exchanger FIRST — it advances the parent's own FTS windows to bracket the
+# current time and fills the derived 2-level window — so the parent data the child interpolates over
+# `[t, t+Δt]` is resident *before* the child steps. Then step the child, then tick the parent clock to
+# match the child's (adaptive) Δt. `Simulation` calls `time_step!(model, Δt; callbacks=...)`.
```
ewquon:
```suggestion
# `[t, t+Δt]` is resident *before* the child steps. Then step the child, and advance the parent to the
# child's new time with (adaptive) Δt. For a prescribed parent, that reduces to a clock tick +
# FTS-window refresh; a live prognostic parent would instead integrate its own dynamics over
# Δt_parent. `Simulation` calls `time_step!(model, Δt; callbacks=...)`.
```

ewquon (approved):

ext/NumericalEarthBreezeExt/breeze_state_exchanger.jl
glwagner: I think the dycore does actually evolve `ρ · qᵛ = ρᵛ` so this may be correct?

examples/breeze_downscaling_era5.jl
glwagner: that sounds good. yes, we need follow up to couple to SlabLand as well

ext/NumericalEarthBreezeExt/breeze_nested_atmosphere.jl
glwagner: I think so! I doubt the precise structure of the ramping matters? So efficiency is good.

src/NestedModels/interpolated_fts_boundary.jl
kaiyuan-cheng: Type stability?

ewquon: Fixed in 3d1c11a

kaiyuan-cheng (approved):

examples/breeze_downscaling_era5.jl
```diff
@@ -28,7 +28,7 @@ using NumericalEarth
 using Oceananigans
 using Oceananigans.Units          # `minutes` for the output schedule (not re-exported by Oceananigans)
 using Breeze
-using CopernicusClimateDataStore # activates NumericalEarthCopernicusClimateDataStoreExt (ERA5 downloads)
+using CDSAPI                     # activates NumericalEarthCDSAPIExt (ERA5 downloads via the CDS REST API)
```
glwagner: I think there is a preference to use the `CopernicusClimateDataStore` because `CDSAPI` has trouble in HPC / distributed environments due to `PythonCall` (if I understand correctly). But we can use CDSAPI temporarily here ?

kaiyuan-cheng: The reason I switched to CDSAPI is that the merged PR suggests it will resolve our data downloading issues in the CI.

glwagner: Do we need to open an issue / fix `CopernicusClimateDataStore`?

kaiyuan-cheng: I revisited the usage of `CopernicusClimateDataStore` and came up with a fix in `NumericalEarthCopernicusClimateDataStoreExt`.

ewquon:
Successfully ran 3h with latest. Note that `max_Δt` was moot for this period, the actual largest simulated Δt was only ~7 s due to |w| ~6→43 m/s (convection develops but doesn't appear to run away)

src/Atmospheres/Atmospheres.jl
```diff
@@ -41,6 +42,13 @@ function bulk_drag end
 # prognostic fields. Extended by atmosphere models (see NumericalEarthBreezeExt).
 function breeze_prognostic_state end
 
+# Conservative advective Δt cap (s) for the outer time-step wizard: safety · min(Δx, Δy) / jet_speed,
+# assuming a jet_speed ≈ 100 m/s upper bound. On a LatitudeLongitudeGrid the smallest zonal spacing sits
+# at the highest |latitude|, so minimum_xspacing is the binding constraint. Atmosphere twin of the ocean's
+# estimate_maximum_Δt in Oceans.
+estimate_maximum_Δt(grid; jet_speed = 100meters/second, safety = 0.5) =
+    safety * min(minimum_xspacing(grid), minimum_yspacing(grid)) / jet_speed
+
```
kaiyuan-cheng, line 51: I don't think we should hard-code this constraint.

ewquon: We can pull it out of the example if you're not comfortable with it. It's not used anywhere by default.

https://github.com/NumericalEarth/NumericalEarth.jl/blob/7db5d674c08a76ccdae49b87cf2952798d52259e/examples/breeze_downscaling_era5.jl#L163-L164

glwagner: what does this constraint provide on top of the time step wizard? I think this would be useful for generating a constant dt, but we don't need it when we do adaptive stepping, perhaps?

.github/workflows/ci.yml
```diff
@@ -83,7 +83,14 @@ jobs:
...
         with:
-          test_args: '--verbose'
+          # Cap parallelism at 2 workers. ParallelTestRunner's default_njobs
+          # assumes ~2 GiB per worker, but the Breeze-compilation and bathymetry
+          # tests peak at 2.5-3.5 GiB RSS each. On the 16 GiB GitHub-hosted
+          # runner, the default (4 = core count) intermittently OOMs when several
+          # heavy tests coincide, and the runner reports it as
+          # "The operation was canceled." Two workers keep peak memory well
+          # within budget while staying comfortably under the job timeout.
+          test_args: '--verbose --jobs=2'
```
glwagner, line 93: cc @giordano

ext/NumericalEarthBreezeExt/breeze_atmosphere_simulation.jl
```diff
@@ -117,12 +125,75 @@ function NumericalEarth.Atmospheres.atmosphere_simulation(grid;
...
+                            coriolis, forcing, closure, clock,
+                            radiation)
+
+    # Breeze ≥0.7 no longer derives the prognostic density from the reference state inside a bare
+    # `set!(θ, …)` — an uninitialized model carries ρ ≡ 0, which NaNs anything that divides by density
+    # (e.g. the MOST surface-flux coupling). Initialize to a resting, hydrostatically balanced state at
+    # the reference θ so the returned atmosphere is valid; a later user `set!` of θ/velocities (without
+    # a `ρ` entry) preserves this density. Callers that derive the full state themselves (the nested
+    # child) pass `initialize = false`: this set!'s default-anchored state otherwise survives into
+    # `initialize_nested_child!` and destabilizes the adiabatic balance twin (DomainError blowup).
```
glwagner, line 143: I don't completely understand this comment. It's mentioning an "adiabatic balance twin" which is unrelated to this. Is there a problem in the adiabatic balance twin that should be fixed instead?

glwagner: Either way I suggest we simply always initialize with some valid resting state anyways for convenience? We can remove `initialize=true`. A user can always re-set the state if desired.

If we want to keep the kwarg I think that we should name it to `initialize_state` (there are many aspects to initialization)

src/DataWrangling/ERA5/ERA5.jl
```diff
@@ -54,6 +57,7 @@ const ERA5Metadatum = Metadatum{<:ERA5Dataset}
 # ERA5 global coverage: 0-359.75 longitude, -90 to 90 latitude at 0.25 degree resolution
 DataWrangling.longitude_interfaces(::ERA5Metadata) = (-0.125, 359.875)
 DataWrangling.latitude_interfaces(::ERA5Metadata) = (-90, 90)
+DataWrangling.native_resolution(::ERA5Dataset) = 1/4
```
glwagner, line 60: is this used?

src/Atmospheres/Atmospheres.jl
ewquon: I think it gets into the role of the wizard. You could just let it go wild and trust the advective CFL constraint is sufficient, in which case you could drop the `max_Δt` specification altogether. (I'm fine with this, and it's verified moot because max simulated dt was << the estimated max_dt=60 s.)

The other perspective, which is slightly more conservative, is viewing this as setting a constant dt, with the ability to automatically reduce dt as necessary to maintain stability. 

In any case, I think that it's informative that in this case the wizard wants such a small dt (~7 s on a 12km grid).

src/DataWrangling/ERA5/ERA5.jl
ewquon: A line of dead code doesn't bother me if it also serves as documentation and could be useful downstream.

glwagner: I'm more worried about a ill-specified `native_resolution` interface being misused or interpreted, if it doesn't have a clear meaning and isn't tested

ewquon: Actually, I think it might be used to add padding to the requested bounding box. First showed up here: https://github.com/NumericalEarth/NumericalEarth.jl/commit/31fd1e6ee18413b97d88b393642dcdc9e0481c35

glwagner: ah nice. Ok, maybe we can change this to something specific to that, like `default_horizontal_padding`. That will be more generic and make sense for non-uniform grids too.

.github/workflows/ci.yml
giordano: I guess it makes sense if you observed OOMs, but I'm slightly worried about _why_ more memory is being used.  Is this during a Breeze workload?  I have the feeling there was a substantial increase in memory usage after https://github.com/NumericalEarth/Breeze.jl/pull/712, or at least around that time, I was trying to investigate that in https://github.com/NumericalEarth/Breeze.jl/pull/841, but it wasn't 100% conclusive.

glwagner: hm ok. perhaps we can keep it in for now and keep an eye on this.

glwagner:
Amazing work @kaiyuan-cheng and @ewquon

glwagner:
bah merged slightly too soon

PR #392, opened 2026-06-30
Use worksize for folded-grid extents

taimoorsohail:
## Summary
- replace four suspicious `size(grid)` extent calculations with `worksize(grid)`
- align folded-grid setup paths with `RightFaceFolded` launch semantics
- keep the change scoped to the pressure, ERA5, bathymetry, and tracer-diffusion paths we identified

## Why
On `RightFaceFolded` grids, `worksize(grid)` includes the extra folded northern row while `size(grid)` does not. These setup paths could therefore build auxiliary fields over `1:Ny` while runtime kernels later operate over the folded `1:Ny+1` working domain, which is a plausible source of seam-local instability.

## Impact
This keeps extent calculations consistent with folded-grid kernel launches and should reduce the risk of mismatched north-fold state near the seam.

## Validation
- inspected the diff to confirm the change is limited to the four intended sites
- did not run local tests, per repository instructions


src/Atmospheres/hydrostatic_pressure.jl
navidcy: this seems like something for a kernel launch?

src/DataWrangling/ERA5/ERA5_prescribed_atmosphere.jl
```diff
@@ -109,7 +109,7 @@ end
 # (in log space) at arbitrary heights via the grid's per-column geopotential — i.e. faithful ln(p)
 # interpolation of the native ERA5 pressure, no hydrostatic reconstruction.
 function pressure_level_field(grid, dataset, architecture)
-    Nx, Ny, Nz = size(grid)
+    Nx, Ny, Nz = worksize(grid)
```
navidcy: this is not even used?

```suggestion
```

src/Atmospheres/hydrostatic_pressure.jl
simone-silvestri: yeah probably this should be a kernel, I think claude really doesn't like kernels

taimoorsohail: This loop is in main, not introduced by this PR, so I think this is an existing bug (not defending Claude of course ;) )

taimoorsohail: Have a look now @simone-silvestri @navidcy

simone-silvestri: looks good

simone-silvestri (approved):

navidcy:
The CI breaks.

Was wondering, do we want to change the `Base.size` definition?
But then `Base.size` gives the number of cells, right? So perhaps we want to have both `size` and `worksize`?

taimoorsohail:
Tests are passing, can we merge?

PR #394, opened 2026-06-30
Fix Reactant reconcile_state! to initialize state exchanger

### LINKED ISSUE #390: Coupled model gives different (wrong) results under Reactant `@compile` because state-exchange regridder fractional indices are not initialized
Build a coupled `AtmosphereLandModel` on a spatially extended
`(Bounded, Bounded)` `LatitudeLongitudeGrid`, then run a few time
steps. The normal (CPU) run and the Reactant-`@compile`d run produce different
physical results. With Reactant, the surface receives zero downwelling
radiation silently.

Root cause: the state-exchanger regridder fractional indices are never
initialized on the Reactant path, so `interpolate_state!` reads every prescribed
forcing field at fractional index `(0,0)` (the halo) instead of the correct cell.

This only manifests on a spatially extended `(Bounded, Bounded)` exchange grid; a
`(Flat, Flat)` single-column grid makes the regridder indices `nothing` (the cell
is read directly).

Here's the MWE:
```julia
using NumericalEarth, Oceananigans, Oceananigans.Units, Statistics, Printf
using Oceananigans.TimeSteppers: Clock, time_step!, update_state!
using Reactant, CUDA
using Reactant: @trace
Reactant.set_default_backend("cpu")
Reactant.LLVM.clopts("-vectorize-slp=false", "-vectorize-loops=false")

# A spatially extended (Bounded, Bounded) exchange grid — a normal regional setup.
make_grid(arch) = LatitudeLongitudeGrid(arch; latitude = (0, 1), longitude = (0, 1),
                                        size = (4, 4), topology = (Bounded, Bounded, Flat))

# Build a coupled atmosphere–land model the normal way, with constant-in-time
# analytic forcing (so the only thing under test is the coupling).
function build_model(grid)
    times = [0.0, 100 * 86400.0]
    atmos = PrescribedAtmosphere(grid, times)
    set!(atmos; u = 4.0, q = 0.004, p = 101325.0)
    for n in eachindex(times); set!(atmos.tracers.T[n], 290.0); end
    update_state!(atmos)

    rad = PrescribedRadiation(grid, times; land_surface = SurfaceRadiationProperties(0.2, 0.97),
                              ocean_surface = nothing, sea_ice_surface = nothing)
    for n in eachindex(times)
        set!(rad.downwelling_shortwave[n], 400.0)
        set!(rad.downwelling_longwave[n], 320.0)
    end
    update_state!(rad)

    land   = SlabLand(grid)
    fluxes = NumericalEarth.EarthSystemModels.InterfaceComputations.default_atmosphere_land_fluxes(
                 land, eltype(grid); solver_stop_criteria = FixedIterations(10))
    al = atmosphere_land_interface(grid, atmos, land; fluxes)
    return AtmosphereLandModel(atmos, land; radiation = rad, atmosphere_land_interface = al, clock = Clock(grid))
end

Δt = 10minutes
N  = 16
seed!(m) = (parent(m.land.temperature) .= 290.0; parent(m.land.water_storage) .= 50.0; parent(m.land.saturation) .= 0.5; nothing)
Tmean(m) = mean(interior(m.land.temperature, :, :, 1))
SWin(m)  = mean(interior(m.radiation.interface_fluxes.land.downwelling_shortwave, :, :, 1))

# CPU: build + run a few steps eagerly
mc = build_model(make_grid(CPU())); seed!(mc)
for _ in 1:N; time_step!(mc, Δt); end
@info @sprintf("CPU (eager):              ⟨T⟩ = %.4f K   surface SWin = %.2f W/m²", Tmean(mc), SWin(mc))

# Reactant: SAME model, SAME steps, via @compile
function run_steps!(model, Δt, n)
    @trace track_numbers=false for _ in 1:n
        time_step!(model, Δt)
    end
    return Tmean(model)
end
mr = build_model(make_grid(ReactantState())); seed!(mr)
crun = Reactant.@compile raise=true raise_first=true sync=true run_steps!(mr, Δt, N)
crun(mr, Δt, N)
@info @sprintf("Reactant @compile:        ⟨T⟩ = %.4f K   surface SWin = %.2f W/m²", Tmean(mr), SWin(mr))

# Reactant again, but call Oceananigans.initialize!(model) before stepping
mr2 = build_model(make_grid(ReactantState())); seed!(mr2)
Oceananigans.initialize!(mr2)
crun2 = Reactant.@compile raise=true raise_first=true sync=true run_steps!(mr2, Δt, N)
crun2(mr2, Δt, N)
@info @sprintf("Reactant @compile + init: ⟨T⟩ = %.4f K   surface SWin = %.2f W/m²", Tmean(mr2), SWin(mr2))
```

which gives
```julia
[ Info: CPU (eager):              ⟨T⟩ = 287.7431 K   surface SWin = 320.00 W/m²
[ Info: Reactant @compile:        ⟨T⟩ = 282.2632 K   surface SWin = 0.00 W/m²
[ Info: Reactant @compile + init: ⟨T⟩ = 287.7720 K   surface SWin = 320.00 W/m²
```
where the radiaiton is not passed into Reactant `@compile` without `initialize!`. This is because of
https://github.com/NumericalEarth/NumericalEarth.jl/blob/b5be880b5bcee11b3ff3bb62d4fb0fe1890f6dd4/ext/NumericalEarthReactantExt.jl#L21
while usual function dispatch gives
https://github.com/NumericalEarth/NumericalEarth.jl/blob/b5be880b5bcee11b3ff3bb62d4fb0fe1890f6dd4/src/EarthSystemModels/earth_system_model.jl#L86-L90
which does call `initialize!`

This is at least partly intentional as the tests demand it:
https://github.com/NumericalEarth/NumericalEarth.jl/blob/b5be880b5bcee11b3ff3bb62d4fb0fe1890f6dd4/test/test_reactant.jl#L45-L46
but I wonder this workflow is a common contract of typical users, who should largely expect the codes to do the same thing with and without Reactant `@compile`.

If this is a real issue the fix is straightforward as we just need to `initialize!` at some point during the model construction.

@glwagner @simone-silvestri do you have thoughts on this?

xkykai:
## Summary

Closes #390.

The `ReactantOSIM` override of `reconcile_state!` in the Reactant extension was a no-op:

```julia
reconcile_state!(model::ReactantOSIM) = nothing
```

while the CPU dispatch ([`src/EarthSystemModels/earth_system_model.jl`](../blob/main/src/EarthSystemModels/earth_system_model.jl)) does real work:

```julia
function Oceananigans.TimeSteppers.reconcile_state!(model::ESM)
    initialize!(model.interfaces.exchanger, model)
    update_state!(model)
    return nothing
end
```

Because `reconcile_state!` runs at model construction, under Reactant the state-exchanger regridder's fractional indices were never populated. `interpolate_state!` then read forcing fields at halo cell `(0, 0)` instead of the correct cells, so the surface received **zero downwelling radiation** and cooled unphysically (~282 K vs ~288 K on CPU). This only manifested on spatially extended `(Bounded, Bounded)` grids — single-column `(Flat, Flat)` grids were unaffected since their regridder is trivial (`nothing` indices → direct cell access).

## Fix

Mirror the CPU `reconcile_state!`, but trace-and-run each step through Reactant `@jit` so the mutations actually land on the `ReactantState` arrays:

```julia
function reconcile_state!(model::ReactantOSIM)
    @jit Oceananigans.initialize!(model.interfaces.exchanger, model)
    @jit Oceananigans.TimeSteppers.update_state!(model)
    return nothing
end
```

This follows the `@jit` direction suggested by @glwagner and @dkytezab in the issue thread, and makes compiled and eager runs agree without requiring a manual `initialize!()` call. It also lets the differentiable-land examples drop their explicit `Oceananigans.initialize!(model_ad)` workaround.

## Testing

⚠️ Needs validation on the actual Reactant path (Julia 1.12 + Reactant).

### DIFF
--- a/ext/NumericalEarthReactantExt.jl
+++ b/ext/NumericalEarthReactantExt.jl
@@ -19,10 +19,6 @@ const ReactantESM{R, A, L, I, O, F, C} = Union{
     EarthSystemModel{R, A, L, I, O, F, C, <:Distributed{ReactantState}},
 }
 
-# Mirror the CPU `reconcile_state!` (initialize the state-exchanger regridder,
-# then update the state), but trace-and-run each step through Reactant. Eager
-# calls would not mutate the `ReactantState` arrays, leaving the regridder's
-# fractional indices uninitialized so `interpolate_state!` reads halo cell (0, 0).
 function reconcile_state!(model::ReactantESM)
     @jit Oceananigans.initialize!(model.interfaces.exchanger, model)
     @jit Oceananigans.TimeSteppers.update_state!(model)
--- a/ext/NumericalEarthReactantExt.jl
+++ b/ext/NumericalEarthReactantExt.jl
@@ -13,7 +13,8 @@ const OceananigansReactantExt = Base.get_extension(
      Oceananigans, :OceananigansReactantExt
 )
 
-const ReactantOSIM{R, A, L, I, O, F, C} = Union{
+# Any EarthSystemModel (every component combination) carried on `ReactantState`.
+const ReactantESM{R, A, L, I, O, F, C} = Union{
     EarthSystemModel{R, A, L, I, O, F, C, <:ReactantState},
     EarthSystemModel{R, A, L, I, O, F, C, <:Distributed{ReactantState}},
 }
@@ -22,7 +23,7 @@ const ReactantOSIM{R, A, L, I, O, F, C} = Union{
 # then update the state), but trace-and-run each step through Reactant. Eager
 # calls would not mutate the `ReactantState` arrays, leaving the regridder's
 # fractional indices uninitialized so `interpolate_state!` reads halo cell (0, 0).
-function reconcile_state!(model::ReactantOSIM)
+function reconcile_state!(model::ReactantESM)
     @jit Oceananigans.initialize!(model.interfaces.exchanger, model)
     @jit Oceananigans.TimeSteppers.update_state!(model)
     return nothing
--- a/ext/NumericalEarthReactantExt.jl
+++ b/ext/NumericalEarthReactantExt.jl
@@ -18,7 +18,15 @@ const ReactantOSIM{R, A, L, I, O, F, C} = Union{
     EarthSystemModel{R, A, L, I, O, F, C, <:Distributed{ReactantState}},
 }
 
-reconcile_state!(model::ReactantOSIM) = nothing
+# Mirror the CPU `reconcile_state!` (initialize the state-exchanger regridder,
+# then update the state), but trace-and-run each step through Reactant. Eager
+# calls would not mutate the `ReactantState` arrays, leaving the regridder's
+# fractional indices uninitialized so `interpolate_state!` reads halo cell (0, 0).
+function reconcile_state!(model::ReactantOSIM)
+    @jit Oceananigans.initialize!(model.interfaces.exchanger, model)
+    @jit Oceananigans.TimeSteppers.update_state!(model)
+    return nothing
+end
 
 import NumericalEarth.EarthSystemModels: same_time_type
 


ext/NumericalEarthReactantExt.jl
```diff
@@ -18,7 +18,15 @@ const ReactantOSIM{R, A, L, I, O, F, C} = Union{
     EarthSystemModel{R, A, L, I, O, F, C, <:Distributed{ReactantState}},
 }
 
-reconcile_state!(model::ReactantOSIM) = nothing
+# Mirror the CPU `reconcile_state!` (initialize the state-exchanger regridder,
+# then update the state), but trace-and-run each step through Reactant. Eager
+# calls would not mutate the `ReactantState` arrays, leaving the regridder's
+# fractional indices uninitialized so `interpolate_state!` reads halo cell (0, 0).
+function reconcile_state!(model::ReactantOSIM)
```
glwagner: should we define this for all ESM or just OSIM?

glwagner (approved):

ext/NumericalEarthReactantExt.jl
```diff
@@ -18,7 +18,15 @@ const ReactantOSIM{R, A, L, I, O, F, C} = Union{
     EarthSystemModel{R, A, L, I, O, F, C, <:Distributed{ReactantState}},
 }
 
-reconcile_state!(model::ReactantOSIM) = nothing
+# Mirror the CPU `reconcile_state!` (initialize the state-exchanger regridder,
+# then update the state), but trace-and-run each step through Reactant. Eager
+# calls would not mutate the `ReactantState` arrays, leaving the regridder's
+# fractional indices uninitialized so `interpolate_state!` reads halo cell (0, 0).
```
simone-silvestri:
```suggestion
```
I think this is the same as the vanilla one so it does not require this comment

ext/NumericalEarthReactantExt.jl
xkykai: done it on `ReactantESM` now

dkytezab:
tests are failing?

PR #395, opened 2026-06-30
Expose stop_time in ocean and sea-ice sims

hgpeterson:
This PR exposes `stop_time` as a kwarg in to the ocean and sea-ice simulation constructors. It defaults to `Inf` for numeric clocks or `DateTime(9999, 12, 31, 23, 59, 59)` for `DateTime` clocks. To determine which, the `clock` kwarg now defaults to `Clock(grid)`.

### DIFF
--- a/src/EarthSystemModels/EarthSystemModels.jl
+++ b/src/EarthSystemModels/EarthSystemModels.jl
@@ -40,7 +40,7 @@ using Thermodynamics: Thermodynamics as AtmosphericThermodynamics
 
 using Oceananigans: Oceananigans, AbstractModel, initialize!,
                     prognostic_state, restore_prognostic_state!
-using Oceananigans.Architectures: architecture, ReactantState
+using Oceananigans.Architectures: architecture, AbstractArchitecture, ReactantState
 using Oceananigans.Diagnostics: NaNChecker
 using Oceananigans.Fields: ZeroField
 using Oceananigans.Simulations: reset_clock!, Simulation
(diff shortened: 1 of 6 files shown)


hgpeterson:
From what I can tell, CI is mostly failing due to download errors, but I do see that the Reactant test fails with
```
AssertionError: `stop_time` is not supported for ReactantModel
```
I see that this comes from [here](https://github.com/CliMA/Oceananigans.jl/blob/03d4d779c132fa3a2ba2809917f7ffe5467f7945/ext/OceananigansReactantExt/Simulations/simulation.jl#L11) in Oceananigans. I guess the Reactant `Simulation` constructor has to have `stop_time = nothing` whereas the normal default is `stop_time = Inf`. Is it worth changing that in Oceananigans to be more consistent, or is it better to just do something like the following here in NumericalEarth:
```julia
default_stop_time(grid, clock) = default_stop_time(architecture(grid), clock)
default_stop_time(arch, clock) = clock.time isa Number ? Inf : Dates.DateTime(9999, 12, 31, 23, 59, 59)
default_stop_time(::ReactantState, clock) = nothing
```
and then
```julia
function hydrostatic_ocean_simulation(grid;
                                      clock = Clock(grid),
                                      stop_time = default_stop_time(grid, clock),
                                      ...
```
for all of the constructors?

@simone-silvestri

simone-silvestri (approved):

simone-silvestri:
I would use the `default_stop_time` approach

PR #396, opened 2026-07-01
Try slackening Reactant compat

dkytezab:
To accomodate newest version


xkykai:
important because we need https://github.com/EnzymeAD/Reactant.jl/pull/3004 !

xkykai (approved):

PR #400, opened 2026-07-01
[WIP] Add ERA5-Land dataset support to DataWrangling

xkykai:
## Summary

Adds the **ERA5-Land** reanalysis (`reanalysis-era5-land`) as a first-class dataset in `DataWrangling`, so land-surface state (soil temperature, soil moisture, skin temperature, snow) can be loaded as `FieldTimeSeries` / single-snapshot `Field`s on the native grid or regridded onto a target land grid.

The motivating use is **observational ground truth for calibrating land models** (e.g. `SlabLand`) online against reanalysis — the data layer that Phases 2–3 of the land-training effort depend on.

## Approach

ERA5-Land is implemented as **subtypes of the existing `ERA5Dataset`** (`ERA5HourlyLand`, `ERA5MonthlyLand`), reusing the entire CDS download / NetCDF-split / `retrieve_data` / region-crop / regrid pipeline. Only what genuinely differs from ERA5 single-levels is overridden:

- **0.1° grid** — `Base.size` → `(3600, 1801, 1)` and specialized longitude/latitude interfaces (the inherited 0.25° values would be silently wrong).
- **Variable dicts** — skin temperature, soil temperature (4 levels), volumetric soil water (4 layers), 2 m temperature/dewpoint, snow depth + snow-water-equivalent.
- **Instantaneous analysis fields** — `conversion_units = nothing` (no accumulation→flux conversion).
- **Land-only masking** — `default_inpainting = nothing` so masked ocean cells stay `NaN` (to be masked in a loss, not smeared into coastal land).
- **CDS request** — the `reanalysis-era5-land` product rejects `product_type`, so `build_era5_request` now sets it via a dispatched `set_product_type!` hook (ERA5 → `["reanalysis"]`; ERA5-Land → omit). The land `Union` method is strictly more specific than the abstract `ERA5Dataset`, so it resolves unambiguously.

## Files

- `src/DataWrangling/ERA5/ERA5_land.jl` (new)
- `src/DataWrangling/ERA5/ERA5.jl`, `src/DataWrangling/DataWrangling.jl`, `src/NumericalEarth.jl` — includes + exports
- `ext/NumericalEarthCDSAPIExt.jl` — `cds_product`/`cds_varnames`/`nc_varnames`/`coord_vars` dispatch + `set_product_type!` refactor

## Testing

Verified against a **live CDS download** (Borneo bbox), not just static checks:

- ✅ Package loads; both dataset types construct and appear in `supported_datasets()`.
- ✅ `Metadata`/`Metadatum` build; `size(ERA5HourlyLand(), :skin_temperature) == (3600, 1801, 1)`; native grid is 0.1° (N→S).
- ✅ **Critical regression check:** the ERA5-Land CDS request omits `product_type`; the single-level request still includes it.
- ✅ **NetCDF short names confirmed against real files** — `skt, stl1, swvl1, t2m, d2m, sde, sd` all matched the dict exactly (no corrections needed).
- ✅ **Ocean masking end-to-end** — a land/sea bbox loaded with finite land cells (289–304 K) and `NaN` ocean cells, no inpainting.

## Notes / follow-ups

- The monthly product (`reanalysis-era5-land-monthly-means`) shares the code path (only its `cds_product` string differs) and was not exercised with a live download.
- Downloading requires the ERA5-Land licence accepted on the CDS account (separate from ERA5).
- No new `PrescribedLand`-style coupling type is added: ERA5-Land targets are plain `FieldTimeSeries` consumed by a loss.

### DIFF
--- a/ext/NumericalEarthCDSAPIExt.jl
+++ b/ext/NumericalEarthCDSAPIExt.jl
@@ -15,7 +15,9 @@ using NumericalEarth.DataWrangling.ERA5: ERA5Dataset, ERA5Metadata, ERA5Metadatu
                                          ERA5_dataset_variable_names, ERA5_netcdf_variable_names,
                                          ERA5PressureLevelsDataset,
                                          ERA5PressureMetadata, ERA5PressureMetadatum,
-                                         ERA5PL_dataset_variable_names, ERA5PL_netcdf_variable_names
+                                         ERA5PL_dataset_variable_names, ERA5PL_netcdf_variable_names,
+                                         ERA5HourlyLand, ERA5MonthlyLand, ERA5LandDataset,
+                                         ERA5Land_dataset_variable_names, ERA5Land_netcdf_variable_names
 using NumericalEarth.DataWrangling.GloFAS: GloFASDataset, GloFASMetadata, GloFASMetadatum,
                                            GloFAS_netcdf_variable_names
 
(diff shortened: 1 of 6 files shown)


bgroenks96:
Just seeing this now, but I would dispute that ERA5-Land serves as a good "ground truth" for calibrating land surface models. It is itself based on a land surface model that is quite far from state-of-the-art and is evaluated against a relatively incomplete set of observations that are not directly assimilated into the land model state; from the paper (Muñoz-Sabater et al. 2021)

> ERA5-Land does not assimilate observations directly. The observations influence the land surface evolution via the atmospheric forcing.

Thus, you should *not* consider ERA5-Land to be a "best estimate" of historical land surface states as one would do with ERA5, but rather as just any other model product. It's still useful as a reference, of course, and we should have it as a data source.

xkykai:
close since no longer relevant

PR #402, opened 2026-07-01
Replace GridFittedBottom with PartialCellBottom

glwagner:
I also removed `active_cells_map`... hopefully `true` is default, if not we should change that.


simone-silvestri (approved):
I am not sure the default is true

glwagner:
should we merge this?

alesok:
I am afraid the default is false
```julia
function ImmersedBoundaryGrid(grid::AbstractUnderlyingGrid, ib::AbstractImmersedBoundary;
                              active_cells_map::Bool=false,
                              active_z_columns::Bool=active_cells_map)
```
Sorry

glwagner:
I think we should merge this anyway, and fix the problem of a poorly chosen default in Oceananigans later.

PR #404, opened 2026-07-01
Add DerivedDatasetBackend for lazily computed FieldTimeSeries

xkykai:
## Problem

`ERA5PrescribedAtmosphere` keeps five of its six variables behind a partly-in-memory `DatasetBackend` (only `time_indices_in_memory` slices resident), but specific humidity — which has no native ERA5 single-level variable and must be derived from dewpoint and pressure — was materialized **eagerly over the entire date range** into a fully in-memory `FieldTimeSeries` at construction. Two consequences:

- memory for the humidity series scales with forcing length, defeating the laziness of everything else (the ceiling that matters for long calibration runs against disk-backed reanalysis, cf. the Path A / streamed-forcing workstream);
- the construction-time sweep indexes `Tᵈ[n]`, `p[n]` for every `n`, repeatedly reloading the sources' small windows from disk.

## Design

`DerivedDatasetBackend` is a sibling of `DatasetBackend` living next to it in `dataset_backend.jl`. It carries a kernel function, a tuple of source `FieldTimeSeries`, and extra parameters; its `set!` fills each resident slice by evaluating a `KernelFunctionOperation` of

```
func(i, j, k, grid, source_slices..., parameters...)
```

over the corresponding source slices. Indexing a source slice outside its resident window repositions that window lazily, so the derived series is exactly as lazy as its sources. The generic `update_field_time_series!` machinery handles window sliding (forward, backward, and `Cyclical` wrap) unchanged, and the backend `Adapt`s by dropping its host-only closure state, mirroring `DatasetBackend`.

`ERA5PrescribedAtmosphere` then builds `qᵛ` in one line with the same `time_indices_in_memory` as the other variables. **No API change** — the derived series behaves like any partly-in-memory `FieldTimeSeries`, including in-loop `fts[Time(t)]` sampling by the coupled model.

The backend is generic on purpose: any derived forcing variable (wind-speed magnitude, unit conversions with multiple inputs, JRA55-style combinations) can use it.

## Validation

- `test/test_derived_dataset_backend.jl` (new, auto-discovered, no downloads): synthetic two-source product and a parameterized single-source scaling on a small grid; verifies values against direct computation while sliding the window forward, jumping backward, and wrapping past the last index, and asserts the resident-slice count never exceeds the window. Passes locally (21/21).
- Integration check against cached ERA5 (Apr 1–6 2020, 1°×1° box): the constructed atmosphere's humidity backend is `DerivedDatasetBackend(1, 8)` over 133 hourly slices (buffer holds 8 slices, not 133), and every lazily computed slice — including backward window jumps — matches the direct `specific_humidity_from_dewpoint` kernel evaluation bit-for-bit.

## Notes for review

- `min(time_indices_in_memory, length(times))` mirrors the clamp the `Metadata`-based constructor applies to the source series.
- The stale `KernelFunctionOperation` import in `ERA5_prescribed_atmosphere.jl` was removed (the ExplicitImports QA test would flag it); the backend file imports it instead.
- Sources are required to share the derived series' grid and times; this is documented rather than checked, matching the trust level of the surrounding code. Happy to add a constructor assertion if preferred.

### DIFF
--- a/src/DataWrangling/dataset_backend.jl
+++ b/src/DataWrangling/dataset_backend.jl
@@ -1,3 +1,5 @@
+using Oceananigans: location
+using Oceananigans.AbstractOperations: KernelFunctionOperation
 using Oceananigans.OutputReaders: Cyclical, AbstractInMemoryBackend, FlavorOfFTS, time_indices
 
 import Oceananigans.OutputReaders: new_backend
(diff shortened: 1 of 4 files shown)


src/DataWrangling/dataset_backend.jl
```diff
@@ -98,3 +100,66 @@ function set!(fts::DatasetFieldTimeSeries, backend=fts.backend)
 
     return nothing
 end
+
+"""
+    DerivedDatasetBackend{F, S, P} <: AbstractInMemoryBackend{Int}
```
simone-silvestri: maybe `DerivedInMemoryBackend`? this is more similar to `InMemory` than `DatasetBackend`

simone-silvestri:
nice feature. Should it be in Oceananigans though? Maybe we can move it after if you need it immediately?

xkykai:
> nice feature. Should it be in Oceananigans though? Maybe we can move it after if you need it immediately?

Can move to Oceananigans too, but I suppose this is an extension of `DatasetBackend`, which lives on NumericalEarth. Should we keep it here instead? Maybe in the future we would want to migrate also `DatasetBackend` to Oceananigans?

src/DataWrangling/dataset_backend.jl
xkykai: Agreed — renamed to `DerivedInMemoryBackend` (and the test file to match) in 3fee8cdd.

simone-silvestri:
It looks like an extension of `AbstractInMemoryBackend` which is an Oceananigans type no? It seems like it is "data agnostic", it does not know anything of an underlying dataset

xkykai:
> It looks like an extension of `AbstractInMemoryBackend` which is an Oceananigans type no? It seems like it is "data agnostic", it does not know anything of an underlying dataset

agreed, let me move it to oceananigans

xkykai:
migrated to https://github.com/CliMA/Oceananigans.jl/pull/5750

PR #405, opened 2026-07-01
Reconcile coupled-model state on the first compiled step under Reactant

### LINKED ISSUE #403: Coupled model under Reactant `@compile` is wrong when prognostic state is `set!` after construction: the first-step `update_state!` is skipped inside `@trace`
This is the residual of https://github.com/NumericalEarth/NumericalEarth.jl/issues/390. https://github.com/NumericalEarth/NumericalEarth.jl/pull/394 fixed the static half of that issue. The
state-exchanger regridder fractional indices are now initialized at construction, so
`@compile`d runs receive downwelling radiation. 

However, if the prognostic state is `set!` after construction (as any calibration / optimization /
differentiable loop does), a Reactant-`@compile`d `@trace` time-stepping loop is wrong 
because the first compiled step reuses flux/auxiliary state left over from construction instead of 
reconciling it with the newly seeded fields.

### Root cause

`EarthSystemModel`'s `time_step!` begins with

```julia
maybe_prepare_first_time_step!(coupled_model, Δt, callbacks)
```
https://github.com/CliMA/Oceananigans.jl/blob/4c8561061e975549532cced0e420d81fedaf769a/src/TimeSteppers/TimeSteppers.jl#L41-L48

whose generic definition (`Oceananigans/src/TimeSteppers/TimeSteppers.jl`) is

```julia
function maybe_prepare_first_time_step!(model, Δt, callbacks)
    if model.clock.iteration == 0        # data-dependent branch
        reconcile_state!(model)
        update_state!(model, callbacks)  # refresh aux/flux state from prognostic state
    end
    return nothing
end
```

On vanilla runs `clock.iteration::Int`, so on the first step this reconciles the flux/exchanger
state with the fields just written by `set!`. On `ReactantState`, `clock.iteration` is a
`ConcreteRNumber`/`TracedRNumber`, so `if clock.iteration == 0` is data-dependent control
flow on a traced scalar *inside the compiled `@trace` loop* — it does not reproduce the
eager semantics, and the first-step refresh never runs. The first compiled `time_step!`
therefore consumes the flux state computed at construction (from the pre-`set!` state).

### MWE

```julia
using NumericalEarth, Oceananigans, Oceananigans.Units, Statistics, Printf
using Oceananigans.TimeSteppers: time_step!, update_state!
using Reactant, CUDA
using Reactant: @trace
Reactant.set_default_backend("cpu")
Reactant.LLVM.clopts("-vectorize-slp=false", "-vectorize-loops=false")

make_grid(arch) = LatitudeLongitudeGrid(arch; latitude = (0, 1), longitude = (0, 1),
                                        size = (1, 1), topology = (Bounded, Bounded, Flat))

function build_model(arch)
    grid  = make_grid(arch)
    atmos = PrescribedAtmosphere(grid, [0.0, 1.0e8])
    for n in 1:2
        set!(atmos.velocities.u[n], 4.0);        set!(atmos.temperature[n], 290.0)
        set!(atmos.specific_humidity[n], 0.004); set!(atmos.pressure[n], 101325.0)
    end
    update_state!(atmos)
    land   = SlabLand(grid)
    fluxes = NumericalEarth.EarthSystemModels.InterfaceComputations.default_atmosphere_land_fluxes(
                 land, eltype(grid); solver_stop_criteria = FixedIterations(8))
    interface = atmosphere_land_interface(grid, atmos, land; fluxes)
    return AtmosphereLandModel(atmos, land; atmosphere_land_interface = interface)
end

Δt = 10minutes
N  = 16
Tmean(m) = mean(interior(m.land.temperature, :, :, 1))

# Eager reference: set the skin temperature AFTER construction, then step. The first
# `time_step!` refreshes the flux from the seeded T (maybe_prepare_first_time_step!).
mc = build_model(CPU())
parent(mc.land.temperature) .= 320
for _ in 1:N; time_step!(mc, Δt); end
@info @sprintf("CPU eager:               ⟨T⟩ = %.5f K", Tmean(mc))

# Reactant: same, but the first compiled step never runs that refresh (the
# `if clock.iteration == 0` guard does not fire under @trace) — unless we call
# update_state! ourselves before the loop.
function run_norefresh(m, Δt, n)
    parent(m.land.temperature) .= 320
    @trace track_numbers=false for _ in 1:n; time_step!(m, Δt); end
    return Tmean(m)
end
function run_refresh(m, Δt, n)
    parent(m.land.temperature) .= 320
    update_state!(m)
    @trace track_numbers=false for _ in 1:n; time_step!(m, Δt); end
    return Tmean(m)
end

mr = build_model(ReactantState())
c0 = Reactant.@compile raise=true raise_first=true sync=true run_norefresh(mr, Δt, N)
T_norefresh = Reactant.to_number(c0(mr, Δt, N))
@info @sprintf("Reactant, no refresh:    ⟨T⟩ = %.5f K   (Δ vs CPU = %+.3e K)", T_norefresh, T_norefresh - Tmean(mc))

mr2 = build_model(ReactantState())
c1 = Reactant.@compile raise=true raise_first=true sync=true run_refresh(mr2, Δt, N)
T_refresh = Reactant.to_number(c1(mr2, Δt, N))
@info @sprintf("Reactant, update_state!: ⟨T⟩ = %.5f K   (Δ vs CPU = %+.3e K)", T_refresh, T_refresh - Tmean(mc))
```

Output:

```
[ Info: CPU eager:               ⟨T⟩ = 286.88632 K
[ Info: Reactant, no refresh:    ⟨T⟩ = 235.06009 K   (Δ vs CPU = -5.183e+01 K)   # drifts badly
[ Info: Reactant, update_state!: ⟨T⟩ = 286.88632 K   (Δ vs CPU = +5.684e-14 K)   # bit-exact
```

@dkytezab @Pangoraw

xkykai:
Closes https://github.com/NumericalEarth/NumericalEarth.jl/issues/403

## Verification

Minimal coupled MWE shown in https://github.com/NumericalEarth/NumericalEarth.jl/issues/403 (1×1 grid, constant `PrescribedAtmosphere`, default `SlabLand`, no
radiation), skin temperature `set!` to 320 K after construction, 16 compiled steps:

```
CPU eager:               ⟨T⟩ = 286.88632 K
Reactant, no refresh:    ⟨T⟩ = 235.06009 K   (Δ vs CPU = -5.183e+01 K)   # before this fix
Reactant, no refresh:    ⟨T⟩ = 286.88632 K   (Δ vs CPU = +5.684e-14 K)   # with this fix — bit-exact
```


dkytezab:
Should we shelve `first_time_step!` then in Reactant? It's going to be confusing to have two paradigms: where Oceananigans `maybe_prepare_first_time_step!` is a no op on Reactant, the brunt being passed over to `first_time_step!` (call to `update_state!` and `reconcile_state!`) but here we have NumericalEarth `maybe_prepare_first_time_step!` mirroring the latter rather than its namesake.

Pangoraw:
> Should we shelve first_time_step! then in Reactant?

I feel like doing the explicit first time step in Reactant might be better, otherwise we are implicitly adding a lot of code to the loop that will run just for the first iteration. Making compile times longer.

test/test_reactant.jl
```diff
@@ -19,6 +19,11 @@ else
     Reactant.set_default_backend("cpu")
 end
 
+# LLVM's SLP vectorizer miscompiles Reactant-raised kernels (it emits a
+# vector-of-i1 `select` on scalar operands, which fails LLVM verification);
+# disable it, as in the issue-#403 MWE.
+Reactant.LLVM.clopts("-vectorize-slp=false", "-vectorize-loops=false")
```
Pangoraw: is this still needed with latest Reactant?

src/EarthSystemModels/InterfaceComputations/atmosphere_land_fluxes.jl
```diff
@@ -126,7 +126,7 @@ end
 #####
 
 @inline land_saturation(i, j, grid, land_state) =
-    (saturation = convert(eltype(grid), land_field_value(land_state.saturation, i, j)),)
+    (saturation = land_field_value(land_state.saturation, i, j),)
```
glwagner: why is this a NamedTuple?

src/EarthSystemModels/InterfaceComputations/atmosphere_land_fluxes.jl
```diff
@@ -137,7 +137,7 @@ end
 
 # Energy state: only the reservoir (skin-humidity) model needs the bulk temperature.
 @inline interface_energy_state(i, j, grid, ::SkinHumidity, land_state) =
-    (temperature = convert(eltype(grid), land_field_value(land_state.T, i, j)),)
+    (temperature = land_field_value(land_state.T, i, j),)
```
glwagner: same q here

test/test_reactant.jl
```diff
@@ -48,3 +57,76 @@ end
...
+
+    function build_atmosphere_land_model(arch)
+        grid  = make_land_grid(arch)
+        atmos = PrescribedAtmosphere(grid, [0.0, 1.0e8])
+        for n in 1:2
+            set!(atmos.velocities.u[n], 4.0)
+            set!(atmos.temperature[n], 290.0)
+            set!(atmos.specific_humidity[n], 0.004)
+            set!(atmos.pressure[n], 101325.0)
+        end
```
glwagner:
```suggestion
        set!(atmos.velocities.u, 4)
        set!(atmos.temperature, 290)
        set!(atmos.specific_humidity, 0.004)
        set!(atmos.pressure, 101325)
```

test/test_reactant.jl
```diff
@@ -48,3 +57,76 @@ end
...
+        # These are the default atmosphere--land fluxes except for FixedIterations, which
+        # gives the flux solver a fixed trip count so the compiled and eager paths trace
+        # identically — the default convergence criterion is a data-dependent `while` loop
+        # that would not.
+        FT = eltype(grid)
+        fluxes = IC.SimilarityTheoryFluxes(FT;
+                                           stability_functions          = IC.atmosphere_land_stability_functions(FT),
+                                           momentum_roughness_length    = convert(FT, 0.1),
+                                           temperature_roughness_length = convert(FT, 0.01),
+                                           water_vapor_roughness_length = convert(FT, 0.01),
```
glwagner: conversions to `FT` should be unnecessary

test/test_reactant.jl
dkytezab: this should be unecessary now

test/test_reactant.jl
xkykai: there is no method for `set!(::FieldTimeSeries, ::Number)`. should we add one?

src/EarthSystemModels/InterfaceComputations/atmosphere_land_fluxes.jl
xkykai: This precedes me, but I think it's to feed into `AirLandInterfaceState` in the `hydrology` argument, so that other components can access it by writing `Ψ.hydrology.saturation` in a self-explanatory manner.

test/test_reactant.jl
xkykai: I've gone back to check (with https://github.com/CliMA/Oceananigans.jl/pull/5725 already included).

I can drop `-vectorize-loops=false`, but when I drop `-vectorize-slp=false` it gives an error. Here's the finding so far:

LLVM's SLP vectorizer miscompiles the raised kernel: it packs two scalar ifelse/mod lanes into a vector<2xf64> and emits llvm.select %m, %a, %b : vector<2xi1>, f64 — a vector mask with scalar operands, which fails to lower to LLVM IR. Bit-exact vs CPU with the flag, fails without. It's an upstream Reactant/LLVM codegen bug unrelated to Oceananigans #5725.

I'll continue investigating.

dkytezab: Are you using Reactant 0.2.269 or later?

xkykai: no, I'm using 0.2.268 because it is capped in Project.toml:

https://github.com/NumericalEarth/NumericalEarth.jl/blob/b03e24f56b50d65bf67a41d8b94fa44e63fdfcc5/Project.toml#L95

let me free it and retry

xkykai: 0.2.270 works! I'll bump it

src/EarthSystemModels/InterfaceComputations/atmosphere_land_fluxes.jl
glwagner: Ok, maybe we need an issue to clean this up

test/test_reactant.jl
glwagner: yes that'd be great

xkykai: will be implemented in https://github.com/CliMA/Oceananigans.jl/pull/5773

src/EarthSystemModels/InterfaceComputations/atmosphere_land_fluxes.jl
xkykai: what do you suggest? is it a common `InterfaceState` struct for all types of interfaces?

ext/NumericalEarthReactantExt.jl
```diff
@@ -25,6 +26,21 @@ function reconcile_state!(model::ReactantESM)
     return nothing
 end
 
+# Mirror Oceananigans' Reactant paradigm. The generic `maybe_prepare_first_time_step!` runs
+# on every step (called from the top of `time_step!`) and refreshes the auxiliary/flux state
+# only on the first step, guarded by `if model.clock.iteration == 0`. On `ReactantState` the
+# iteration is a traced scalar, so that guard cannot branch inside a compiled loop. Rather
+# than trace a first-step-only conditional into every iteration, we make the hook a no-op and
+# do the refresh explicitly and unconditionally in `first_time_step!` — matching how
+# Oceananigans handles its own Reactant models, and keeping the compiled stepping loop lean.
```
glwagner: make this comment shorter? (claude loves excessively verbose comments...)

src/EarthSystemModels/InterfaceComputations/atmosphere_land_fluxes.jl
glwagner: I think the caller of `land_saturation` can build it into a `NamedTuple`. It doesn't make sense that every implementation of `land_saturation` needs to add the `saturation` key, I think.

glwagner:
```suggestion
    (; saturation = land_field_value(land_state.saturation, i, j))
```

src/EarthSystemModels/InterfaceComputations/atmosphere_land_fluxes.jl
glwagner:
```suggestion
    (; temperature = land_field_value(land_state.T, i, j))
```

xkykai:
@copilot resolve the merge conflicts in this pull request

src/EarthSystemModels/InterfaceComputations/atmosphere_land_fluxes.jl
xkykai: Done — `land_saturation` now returns a scalar and the caller (`interface_hydrology_state`/`interface_energy_state`) builds the NamedTuple, per your suggestion.

src/EarthSystemModels/InterfaceComputations/atmosphere_land_fluxes.jl
xkykai: Applied.

ext/NumericalEarthReactantExt.jl
xkykai: Shortened in af3ac8fc (along with the other verbose comments in the PR).

PR #408, opened 2026-07-02
Bump Reactant compat to 0.2.274 to fix GPU segfault

### LINKED ISSUE #407: Coupled model segfaults on the Reactant GPU backend in reconcile_state!
Constructing (or time-stepping) a coupled model on the **Reactant GPU backend** segfaults during `reconcile_state!`.

### MWE

```julia
using NumericalEarth, Oceananigans, Reactant, CUDA

Reactant.set_default_backend("gpu")

grid = RectilinearGrid(ReactantState(); size = (), topology = (Flat, Flat, Flat))
AtmosphereLandModel(PrescribedAtmosphere(grid), SlabLand(grid))   # <-- segfaults
```

- `set_default_backend("cpu")` → works.
- The crash is at model construction (`reconcile_state!`), before any physics runs.

### Cause

`reconcile_state!` compiles the model kernels with a plain `@jit`. On the GPU backend that lowers the KernelAbstractions launch through EnzymeXLA's kernel-call ABI rewrite (`rewriteKernelCallABI` → `gpu.LaunchFuncOp`), which segfaults on the large coupled `update_state!` kernel (the atmosphere–land turbulent-flux / similarity-solver kernel). A trivial single-kernel `launch!` on the same grid does *not* reproduce it, so it is specific to the complex coupled kernel. Compiling with `raise=true` (lowering to StableHLO instead of a GPU launch) avoids the crashing path.

### Stacktrace (top frames)

```
signal 11 (Segmentation fault)
rewriteKernelCallABI(... gpu::LaunchFuncOp ...)     libReactantExtra.so
LowerJITPass::runOnOperation()                      libReactantExtra.so
mlir::PassManager::run                              libReactantExtra.so
run_pass_pipeline!   Reactant/.../OptimizationPasses.jl:213
compile_mlir!        Reactant/.../Compiler.jl:452
@jit  ->  reconcile_state!  ext/NumericalEarthReactantExt.jl
AtmosphereLandModel(...)
```

### Environment

Reactant 0.2.268, Enzyme 0.13.173, Oceananigans 0.110.6, Julia 1.12.6, NVIDIA Tesla T4 (CC 7.5), CUDA runtime 13.1 (XLA/PJRT).

Fixed by compiling `reconcile_state!` with `raise=true` (PR linked below).

xkykai:
Closes NumericalEarth/NumericalEarth.jl#407.

### New update:

Bumps the `Reactant` compat lower bound `0.2.270 → 0.2.274` (root, `test`, `docs`), which pulls in `Reactant_jll` 0.0.395. That binary fixes the GPU segfault in the coupled `update_state!` kernel launch at model construction (upstream [EnzymeAD/Reactant.jl#3026](<https://github.com/EnzymeAD/Reactant.jl/issues/3026>)), so `reconcile_state!` keeps plain `@jit` — the earlier `raise=true` workaround is no longer needed.

---

On the Reactant GPU backend, reconcile_state!'s plain @jit lowers the coupled model's kernel launch through EnzymeXLA's kernel-call ABI rewrite (rewriteKernelCallABI → gpu.LaunchFuncOp), which segfaults on the large coupled update_state! kernel. This happens at model construction, before any physics runs.

This compiles reconcile_state! with raise=true, so the kernels are raised to StableHLO instead of lowered to a GPU launch, avoiding the crashing path. It's also the lowering the differentiable examples already use for the gradient, so the forward and adjoint passes share the same path.

function reconcile_state!(model::ReactantESM)

* @jit Oceananigans.initialize!(model.interfaces.exchanger, model)
* @jit Oceananigans.TimeSteppers.update_state!(model)
* @jit raise=true Oceananigans.initialize!(model.interfaces.exchanger, model)
* @jit raise=true Oceananigans.TimeSteppers.update_state!(model)<br>return nothing<br>end<br>Verification (NVIDIA Tesla T4, Reactant 0.2.268)<br>The MWE in NumericalEarth/NumericalEarth.jl#407 no longer segfaults; the coupled model builds on GPU.<br>The full differentiable dry-layer slab example runs end-to-end on GPU (build + reverse-mode gradient of the 841-step loop): ∂T/∂𝒫̇ = -1.02e3 K/(kg m⁻² s⁻¹) ≈ -0.047 K/mm, matching the CPU-documented \~0.04 K/mm sign and magnitude.<br>Runtime is unaffected (XLA re-vectorizes the raised StableHLO); the only cost is a somewhat longer compile.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)


glwagner (approved):

dkytezab:
I this is a reasonably fine workaround for the moment, but we should find the underlying cause of the segfault and address it there.

xkykai:
ok, will close this once tests pass as a temporary workaround.
I've added a TODO so if the resolution of https://github.com/EnzymeAD/Reactant.jl/issues/3026 solves this problem we can roll back the fix then.

xkykai:
for the record tests fail right now because of https://github.com/EnzymeAD/Reactant.jl/issues/3028

Pangoraw:
FYI @xkykai there is a recent jll version (Reactant_jll@0.0.395) that fixes the segfault for without `@jit` and also the rounded raise for with `@jit raise=true`

PR #415, opened 2026-07-06
Add ASTER GED v3 surface-emissivity dataset

xkykai:
Adds `ASTERGEDv3` — the Global Emissivity Dataset v3 from ASTER, the Advanced Spaceborne Thermal Emission and Reflection Radiometer (NASA LP DAAC): a static 2000–2008 clear-sky-mean land-surface emissivity climatology on a WGS84 lat/lon grid at 1 km (`:low_1km`, default) or 100 m (`:high_100m`), as a `DataWrangling` submodule.

```julia
using NumericalEarth, ArchGDAL   # ArchGDAL + Earthdata credentials activate the HDF5 read/download

region = BoundingBox(longitude = (-112.8, -111.2), latitude = (35.2, 36.8))
ε = Field(Metadatum(:emissivity; dataset = ASTERGEDv3(), region), grid)
land_surface = SurfaceRadiationProperties(albedo = α, emissivity = ε)
```

## What it does

- Resolves the intersecting 1°×1° HDF5 tiles via NASA CMR, downloads them from Earthdata Cloud (`EARTHDATA_USERNAME` / `EARTHDATA_PASSWORD`, tiles cached for reuse), reads them through GDAL's HDF5 driver, and mosaics + clips to the `BoundingBox`.
- Collapses the five ASTER TIR bands to one broadband ε with [Ogawa & Schmugge (2004)](https://doi.org/10.1175/1087-3562(2004)008%3C0001:MSBEOT%3E2.0.CO;2) coefficients (they sum to 1, so it is a convex combination), decoding in `Float32` (mean ×0.001, SDev ×0.0001). The decoded broadband field is cached as a regional NetCDF.
- Returns a reduced `(Center, Center, Nothing)` `Field`, safely `stateindex`-able at any `k`, so it plugs directly into `SurfaceRadiationProperties`. `:emissivity_uncertainty` propagates the per-band SDev to a broadband σ(ε).
- `ASTERGEDv3` is exported from `NumericalEarth`; `:high_100m` / `:low_1km` map internally to NASA's `AG100` / `AG1KM` short names.

## Gap filling

ASTER GED retrieves an emissivity almost everywhere, **including over water** (open-ocean broadband ε ≈ 0.98, against 0.94–0.96 over land, so a real value rather than a fill). Genuine gaps appear only where a clear-sky retrieval was never obtained (persistent cloud in the humid tropics, snow screened as cloud) and decode to `NaN`.

Those gaps are filled from the same surface: `retrieve_data` blanks the water before inpainting, so a coastal land gap draws on land donors alone instead of inheriting the ocean value, and water keeps its own retrievals. Water gaps are left to the ungated pass that `Field` already runs, so the returned field is finite everywhere.

Generic inpainting is otherwise untouched, apart from `propagating` gaining a stall check so gaps with no reachable donor stop the sweeps rather than spinning at `maxiter = Inf`.

## Shared Earthdata access

CMR granule discovery and authenticated download move out of the ArchGDAL extension into `DataWrangling`, as product-agnostic `cmr_granules` / `earthdata_download` that any Earthdata-hosted dataset can use. ASTER GED keeps only its short name and version.

## Example — 100 m broadband emissivity

Clear-sky desert (Grand Canyon, gap-free, four-tile mosaic) against the humid tropics (Congo, cloud gaps inpainted). Plots the `Field`s directly, no NetCDF round-trip:

```julia
using NumericalEarth
using NumericalEarth.DataWrangling: BoundingBox
using Oceananigans
using ArchGDAL      # activates the ASTER GED HDF5 read/download extension
using CairoMakie

dataset = ASTERGEDv3(resolution = :high_100m)

function emissivity_field(dataset, name, longitude, latitude; size = (512, 512))
    grid = LatitudeLongitudeGrid(CPU(); size, longitude, latitude,
    region = BoundingBox(; longitude, latitude)
    return Field(Metadatum(name; dataset, region), grid)
end

gc_emissivity  = emissivity_field(dataset, :emissivity,             (-112.8, -111.2), (35.2, 36.8))
gc_uncertainty = emissivity_field(dataset, :emissivity_uncertainty, (-112.8, -111.2), (35.2, 36.8))
congo_emissivity = emissivity_field(dataset, :emissivity, (18, 20), (-1, 1))

fig = Figure(size = (1650, 520), backgroundcolor = :white)
Label(fig[0, 1:6], "ASTER GED v3 broadband emissivity (100 m)"; fontsize = 19, font = :bold)

ax1 = Axis(fig[1, 1]; title = "Emissivity — Grand Canyon (4-tile mosaic)",
           xlabel = "longitude (°)", ylabel = "latitude (°)", aspect = 1)
hm1 = heatmap!(ax1, gc_emissivity; colormap = :viridis, colorrange = (0.90, 0.98))
Colorbar(fig[1, 2], hm1; label = "broadband ε")

ax2 = Axis(fig[1, 3]; title = "Uncertainty σ(ε) (SDev path)",
           xlabel = "longitude (°)", ylabel = "latitude (°)", aspect = 1)
hm2 = heatmap!(ax2, gc_uncertainty; colormap = :viridis, colorrange = (0, 0.02))
Colorbar(fig[1, 4], hm2; label = "broadband σ(ε)")

ax3 = Axis(fig[1, 5]; title = "Emissivity — Congo basin (cloud gaps inpainted)",
           xlabel = "longitude (°)", ylabel = "latitude (°)", aspect = 1)
hm3 = heatmap!(ax3, congo_emissivity; colormap = :viridis, colorrange = (0.90, 0.98))
Colorbar(fig[1, 6], hm3; label = "broadband ε")

save("asterged_emissivity_map_100m.png", fig)
```

## Caveats

The broadband coefficients are a Sahara-desert regression applied globally. Being a static clear-sky mean, the product also misses snow (ε 0.03–0.08 higher) and soil-moisture effects, and its sampling biases it toward dry states. A docs `@example` page is a follow-up.


xkykai:
Running the script above gives
asterged_emissivity_map_100m

glwagner:
Why did you "save for plotting" to NetCDF? is there some issue plotting the `Field`?

ext/NumericalEarthArchGDALExt.jl
```diff
@@ -45,4 +50,184 @@ function NumericalEarth.DataWrangling.IBCAO.reproject_ibcao_to_netcdf(tiff_path,
...
+            longitude = Float64.(read_asterged_subdataset(h5, "//Geolocation/Longitude")[:, 1, 1])
+            latitude  = Float64.(read_asterged_subdataset(h5, "//Geolocation/Latitude")[1, :, 1])
+
+            ii = findall(λ -> (west - margin) <= λ <= (east + margin), longitude)
+            jj = findall(φ -> (south - margin) <= φ <= (north + margin), latitude)
+            (isempty(ii) || isempty(jj)) && continue
+
+            mean_dn = read_asterged_subdataset(h5, "//Emissivity/Mean")[ii, jj, :]
+            sdev_dn = read_asterged_subdataset(h5, "//Emissivity/SDev")[ii, jj, :]
+            lwmap   = read_asterged_subdataset(h5, "//Land_Water_Map/LWmap")[ii, jj, 1]
```
glwagner: Does this mean that you cannot download these datasets independently?

src/DataWrangling/ASTERGED/ASTERGED.jl
```diff
@@ -0,0 +1,404 @@
...
+
+"""
+    ASTERGEDv3 <: AbstractStaticDataset
+
+ASTER Global Emissivity Dataset (GED) v3: a static (2000–2008 clear-sky mean)
+climatology of land-surface emissivity on a plain geographic (WGS84 lat/lon)
+grid, distributed as HDF5 in 1°×1° tiles. Two resolutions are supported:
+
+- `:AG100` — 100 m (3 arcsec, 1000×1000 px/tile). Primary, highest resolution.
+- `:AG1km` — 1 km (30 arcsec, 100×100 px/tile). Coarser sibling.
```
glwagner: why the "AG" prefix?

src/DataWrangling/ASTERGED/ASTERGED.jl
```diff
@@ -0,0 +1,404 @@
...
+struct ASTERGEDv3 <: AbstractStaticDataset
+    resolution :: Symbol                        # :AG100 (100 m) or :AG1km (1 km)
+    broadband_coefficients :: Vector{Float64}   # 5-vector for the ε_broadband synthesis
+    water_emissivity :: Float64                 # constant substituted where the tile land-water map says water
+end
+
+"""
+    ASTERGEDv3(; resolution = :AG100,
+                 broadband_coefficients = OGAWA_SCHMUGGE_2004_BROADBAND_COEFFICIENTS,
+                 water_emissivity = 0.97)
```
glwagner: interesting... it could motivate a `NumericalEarth.defaults` struct, if this needs to be in sync with other defaults (which I am not sure where those are)

glwagner: I also wonder if we would want to support values other than constants, but maybe not (and could be for another PR)

src/DataWrangling/ASTERGED/ASTERGED.jl
```diff
@@ -0,0 +1,404 @@
...
+"""
+    broadband_emissivity_map(decoded_bands, coefficients)
+
+Collapse a decoded emissivity array of shape `(5, Nx, Ny)` (band index first, as
+in the HDF5 `/Emissivity/Mean` layout) to a broadband `(Nx, Ny)` array via
+[`broadband_emissivity`](@ref) along the band dimension.
+"""
+function broadband_emissivity_map(decoded_bands, coefficients)
+    _, Nx, Ny = size(decoded_bands)
+    result = Array{Float32}(undef, Nx, Ny)
```
glwagner: should this be done on GPU? Or at least permitted?

glwagner:
It might make sense to add a section in the docs tutorial for land data + computations that are done here. It can be a future PR though.

glwagner:
Another note to self is that I am wondering how this will all work in a distributed context, eg if we run 100m resolution globally, we want to do the calculations in here independently on every node. It's ok to put optimizations in a future PR but we should think about it a bit.

src/DataWrangling/ASTERGED/ASTERGED.jl
xkykai: that's the name in the raw data, but I agree it's not needed. I'll make the symbols `:high_100m` and `low_1km` instead. if you have better ideas!

ext/NumericalEarthArchGDALExt.jl
xkykai: yeah the raw `.h5` data comes with all 3 variables bundled. but if you think we should have the option to build our netcdf files for selected variables only I can make that change

xkykai:
> Why did you "save for plotting" to NetCDF? is there some issue plotting the `Field`?

no there's nothing wrong with it. I was just stitching together two vibe-coded scripts because it'll take a while to rerun. but a script like
```julia
using NumericalEarth
using NumericalEarth.DataWrangling: BoundingBox, Metadatum
using Oceananigans
using ArchGDAL  # activates the HDF5 tile read/download extension
using CairoMakie

dataset = ASTERGEDv3()

#####
##### 1. Multi-tile static emissivity map — US Southwest (Grand Canyon / Colorado Plateau)
#####

southwest = BoundingBox(longitude = (-112.8, -111.2), latitude = (35.2, 36.8))
southwest_grid = LatitudeLongitudeGrid(CPU(), Float32;
                                       size = (320, 320),
                                       longitude = (-112.8, -111.2),
                                       latitude = (35.2, 36.8),

emissivity  = Field(Metadatum(:emissivity; dataset, region = southwest), southwest_grid)
uncertainty = Field(Metadatum(:emissivity_uncertainty; dataset, region = southwest), southwest_grid)

#####
##### 2. Coverage-hole probe — Congo basin (persistent cloud)
#####

congo = BoundingBox(longitude = (18, 20), latitude = (-1, 1))
congo_grid = LatitudeLongitudeGrid(CPU(), Float32;
                                   size = (200, 200),
                                   longitude = (18, 20),
                                   latitude = (-1, 1),

congo_emissivity = Field(Metadatum(:emissivity; dataset, region = congo), congo_grid)

#####
##### Plot the Fields directly — Oceananigans' Makie extension supplies the coordinates
#####

fig = Figure(size = (1650, 520))
Label(fig[0, 1:6], "ASTER GED v3 AG100 static broadband emissivity"; fontsize = 19, font = :bold)

ax1 = Axis(fig[1, 1]; title = "Emissivity — US Southwest (4-tile mosaic)",
           xlabel = "longitude (°)", ylabel = "latitude (°)", aspect = 1)
hm1 = heatmap!(ax1, emissivity; colormap = :viridis, colorrange = (0.90, 0.98))
Colorbar(fig[1, 2], hm1; label = "broadband ε")

ax2 = Axis(fig[1, 3]; title = "Uncertainty (SDev path)",
           xlabel = "longitude (°)", ylabel = "latitude (°)", aspect = 1)
hm2 = heatmap!(ax2, uncertainty; colormap = :viridis, colorrange = (0, 0.02))
Colorbar(fig[1, 4], hm2; label = "broadband σ(ε)")

ax3 = Axis(fig[1, 5]; title = "Congo basin — coverage holes inpainted, water filled",
           xlabel = "longitude (°)", ylabel = "latitude (°)", aspect = 1)
hm3 = heatmap!(ax3, congo_emissivity; colormap = :viridis, colorrange = (0.90, 0.98))
Colorbar(fig[1, 6], hm3; label = "broadband ε")

save("asterged_emissivity_static_map.png", fig)
```

would work fine.

xkykai:
> Another note to self is that I am wondering how this will all work in a distributed context, eg if we run 100m resolution globally, we want to do the calculations in here independently on every node. It's ok to put optimizations in a future PR but we should think about it a bit.

Suggestions by Fable 5:

1. Each rank should build `Metadatum` with its own `BoundingBox`
2. Derive each rank's `BoundingBox` automatically from its local grid (plus an interpolation margin).
3. Each rank must fetch its own dataset.
4. A persistent granule-keyed tile cache instead so neighboring ranks sharing a boundary tile don't re-download it.
5. Gap inpainting is window-local: a rank whose window lies entirely inside a retrieval gap (humid-tropics cloud holes) has no valid source — either exchange halos between inpainting iterations or pre-fill gaps from the "1km" product.

src/DataWrangling/ASTERGED/ASTERGED.jl
xkykai: refactored (together with other components that use `water_emissivity`) to `default_water_emissivity`

ext/NumericalEarthArchGDALExt.jl
```diff
@@ -45,4 +50,186 @@ function NumericalEarth.DataWrangling.IBCAO.reproject_ibcao_to_netcdf(tiff_path,
...
+
+# Earthdata-authenticated download of a single tile, retried on transient
+# failures (Earthdata Cloud drops connections occasionally). Credentials come
+# from the EARTHDATA_USERNAME / EARTHDATA_PASSWORD environment variables.
+function earthdata_download(url, path; attempts = 3)
+    haskey(ENV, "EARTHDATA_USERNAME") && haskey(ENV, "EARTHDATA_PASSWORD") ||
+        error("NASA Earthdata credentials not found. Set EARTHDATA_USERNAME and " *
+              "EARTHDATA_PASSWORD (register free at https://urs.earthdata.nasa.gov).")
+    username = ENV["EARTHDATA_USERNAME"]
+    password = ENV["EARTHDATA_PASSWORD"]
```
bgroenks96: I guess this isn't exactly a long-term solution, right? Storing passwords as plaintext in environment variables isn't exactly good security practice. But I guess fine for now.

bgroenks96 (approved):

ext/NumericalEarthArchGDALExt.jl
xkykai: Yeah I suppose. I was just following ECCO's pattern:
https://github.com/NumericalEarth/NumericalEarth.jl/blob/2266ea820b7a7cd85b34b986f889c99331e09274/src/DataWrangling/ECCO/ECCO.jl#L277-L278

xkykai: I've updated this part to follow ECCO's treament exactly

ext/NumericalEarthArchGDALExt.jl
```diff
@@ -45,4 +56,185 @@ function NumericalEarth.DataWrangling.IBCAO.reproject_ibcao_to_netcdf(tiff_path,
     return nothing
 end
 
+#####
+##### ASTER GED emissivity ingest (HDF5 tiles on a plain lat/lon grid)
+#####
+##### Resolves the 1°×1° HDF5 tiles intersecting the region via NASA CMR, downloads
```
glwagner: say what "ASTER GED" means

glwagner: consider adding a link to description

ext/NumericalEarthArchGDALExt.jl
```diff
@@ -45,4 +56,185 @@ function NumericalEarth.DataWrangling.IBCAO.reproject_ibcao_to_netcdf(tiff_path,
...
+            @warn "CMR granule query failed (attempt $attempt of $attempts); retrying..." url error
+            sleep(2attempt)
+        end
+    end
+end
+
+# Query CMR for the ASTER GED tile `.h5` download URLs intersecting `bbox`, paging
+# until a short page signals the last one, and keeping one download URL per granule
+# (preferring a protected `data`-host endpoint).
+function NumericalEarth.DataWrangling.ASTERGED.earthdata_cmr_granules(short_name, version, bbox::BoundingBox; page_size = 2000)
```
glwagner: is this a more general process of finding the intersection between tiles and bounding boxes, not specific to ASTERGED?

src/DataWrangling/ASTERGED/ASTERGED.jl
```diff
@@ -0,0 +1,409 @@
...
+const OGAWA_SCHMUGGE_2004_BROADBAND_COEFFICIENTS = [0.088, 0.053, 0.174, 0.380, 0.305]
+
+#####
+##### Dataset type
+#####
+
+"""
+    ASTERGEDv3 <: AbstractStaticDataset
+
+ASTER Global Emissivity Dataset (GED) v3: a static (2000–2008 clear-sky mean)
```
glwagner: define "ASTER"

src/DataWrangling/ASTERGED/ASTERGED.jl
```diff
@@ -0,0 +1,409 @@
...
+
+Data source: https://www.earthdata.nasa.gov/data/catalog/lpcloud-ag100-003
+Reference: Hulley et al. (2015), GRL, doi:10.1002/2015GL065564.
+"""
+struct ASTERGEDv3 <: AbstractStaticDataset
+    resolution :: Symbol   # :low_1km (1 km) or :high_100m (100 m)
+end
+
+"""
+    ASTERGEDv3(; resolution = :low_1km)
```
glwagner:
```suggestion
    ASTERGEDv3(resolution = :low_1km)
```

could consider positional arg (but not necessary)

src/DataWrangling/ASTERGED/ASTERGED.jl
```diff
@@ -0,0 +1,409 @@
...
+Base.show(io::IO, dataset::ASTERGEDv3) = print(io, summary(dataset))
+
+const ASTERGEDMetadatum = Metadatum{<:ASTERGEDv3}
+
+# Variable name in the regional NetCDF written by the download step (which stores
+# the already-decoded broadband floats — see the ArchGDAL extension).
+const ASTERGED_variable_names = Dict(
+    :emissivity             => "emissivity",
+    :emissivity_uncertainty => "emissivity_uncertainty",
+)
```
glwagner: apparently, the names are already good!

src/DataWrangling/inpainting.jl
```diff
@@ -13,71 +13,107 @@ end
...
+        fill_halo_regions!(field)
+        propagate_step!(inpainting, field, mask, nothing, substituting_field)
+    end
 
-    # Need temporary field to avoid a race condition
-    parent(substituting_field) .= parent(field)
+    launch!(arch, grid, size(field), _fill_nans!, field)
+    fill_halo_regions!(field)
 
-    while propagating(field, mask, iter, inpainting)
```
glwagner, line 42: what happens to the `propagating` criteria?

src/DataWrangling/inpainting.jl
```diff
@@ -13,71 +13,107 @@ end
...
+    iter = 0
+    previous_nans = -1
 
+    # Temporary field to avoid a race condition between propagate and substitute.
+    parent(substituting_field) .= parent(field)
+
+    while true
+        nans = sum(isnan, field; condition = interior(mask))
+        (nans == 0 || nans == previous_nans || iter ≥ inpainting.maxiter) && break
+        previous_nans = nans
```
glwagner: this removes the dispatchable function `propagating` which has the effect of hard-coding nearest neighbor inpainting

src/DataWrangling/metadata_field.jl
```diff
@@ -231,14 +235,23 @@ function Oceananigans.Fields.Field(metadata::Metadatum, arch=CPU();
     field = Field{LX, LY, LZ}(grid)
 
     if !isnothing(inpainting)
+        # Resolve the optional region partition before the cache lookup so the cache
+        # key can distinguish a surface-aware result from a surface-agnostic one.
+        if isnothing(regions)
+            regions = inpainting_regions(metadata, field)
+        end
+        regions_applied = !isnothing(regions)
```
glwagner: maybe clarify whats happening

glwagner (approved):
looks good but consider reverting to the original inpainting design

src/DataWrangling/inpainting.jl
xkykai: I added a `previous_gaps` option so that once gaps no longer decrease we stop

xkykai:
@copilot resolve the merge conflicts in this pull request

src/DataWrangling/inpainting.jl
xkykai: solved, also added a stall check where loop exists when gaps can no longer be filled

ext/NumericalEarthArchGDALExt.jl
xkykai: is slightly more general, abstracted out for earthdata-type datasets

PR #416, opened 2026-07-06
ETH canopy height → aerodynamic roughness, and as the DSM bare-earth object height

xkykai:
## Summary

Adds the ETH global 10 m canopy-height dataset and a Raupach drag-partition roughness closure. One measured canopy field feeds two consumers: the aerodynamic roughness of the surface, and the object height `bare_earth_elevation` subtracts from a Digital Surface Model.

Targets `xk/dsm` for `bare_earth_elevation` (#465); rebase to `main` once that lands.

## API

- `ETHSentinel2CanopyHeight()` — ETH Sentinel-2 10 m canopy height (Lang et al. 2023). Tiles are Cloud-Optimized GeoTIFFs read through `/vsicurl/` range requests using the public share token, so a windowed read never pulls a whole 415 MB tile. Requires `using ArchGDAL`. Non-forest `0` is a valid height; only the no-data byte is masked to `NaN`.
- `canopy_height_field(grid, dataset)` — area-averages the 10 m pixels onto a model grid (`gdalwarp -r average`, not point sampling). Regional windows require a `BoundingBox`.
- `tall_canopy_fraction_field(grid, dataset; threshold = 2)` — fraction of each cell standing under canopy at least `threshold` m tall. A threshold cannot be recovered from a cell mean, so the tiles are aggregated ten cells per grid-cell side (never finer than the product's 10 m pixels) and thresholded there.
- `DragPartitionRoughness(; vegetation_type)` — Raupach (1994), parameterized for land cover by Jasinski et al. (2005) and recalibrated by Borak et al. (2025). Leaf area index, canopy height, and IGBP class → momentum roughness length `ℓᵐ` and zero-plane displacement `d`. Allocation-free and kernel-safe.
- `canopy_roughness(closure, leaf_area_index, canopy_height)` — `Number`, `Field`, or `FieldTimeSeries` in, matching output out; `compute_aerodynamic_roughness!` for the in-place `:xy` kernel.
- `canopy_drag_parameters`, `representative_canopy_height`, `nonvegetated_roughness` — per-IGBP-class tables (Borak Tables 2–4).
- `bare_earth_elevation(surface, object_heights...)` — DSM minus the per-cell maximum object height, clamped at zero.

```julia
using NumericalEarth, ArchGDAL

canopy_height   = canopy_height_field(grid, ETHSentinel2CanopyHeight())
z_bare          = bare_earth_elevation(z_dsm, canopy_height)
leaf_area_index = Field{Center, Center, Nothing}(grid); set!(leaf_area_index, 5)
ℓᵐ, d = canopy_roughness(DragPartitionRoughness(), leaf_area_index, canopy_height)
```

## Notation

`ℓᵐ` momentum roughness length, `d` zero-plane displacement, `𝒜` leaf area index, `𝒜ᶜ` its critical (skimming) value — all added to `docs/src/appendix/notation.md`. `𝒜ᶜ` caps only the wind ratio, so `ℓᵐ` keeps skimming down while `d` keeps rising toward the canopy top as the canopy densifies.

## Example — `examples/eth_canopy_bare_earth.jl`

Central Amazon (Rio Negro–Solimões) at ~110 m. DSM elevation, the ETH canopy, the bare-earth DTM, and the `ℓᵐ` that same canopy induces:

fig1_canopy_dsm_roughness

Along 3°S the DSM rides a canopy height above bare earth over forest and drops onto it over the river floodplains — the gap is the removed canopy:

fig2_transect

The same closure driven with idealized inputs, checked against Borak et al. (2025) Table 5:

fig3_closure_verification

## Not in scope

- `d` is produced but not yet consumed by the flux solver (MOST places the surface at z = 0).
- Live COG reads stay out of CI: the read path is covered by a synthetic GeoTIFF fixture, and `test_ethsentinel2canopy_downloading.jl` exercises the real COGs but is excluded from the default suite.
- The example's leaf area index is uniform, and the DSM is token-free ETOPO 2022 rather than the commercial-use GLO-30.


examples/eth_canopy_bare_earth.jl
```diff
@@ -0,0 +1,179 @@
...
+
+# ## (1) Bare-earth terrain — DSM minus the measured canopy
+z_bare  = bare_earth_elevation(z_dsm, canopy_height)
+removed = compute!(Field(z_dsm - z_bare))      # the canopy lift removed from the terrain
+
+# ## (2) Roughness — the same canopy through the Raupach closure
+# The basin is evergreen broadleaf forest, a uniform dense canopy (LAI 5) here.
+# `DragPartitionRoughness()` defaults to the `:evergreen_broadleaf_forest` class.
+lai = Field{Center, Center, Nothing}(grid); set!(lai, 5)
+z0, d0 = Field{Center, Center, Nothing}(grid), Field{Center, Center, Nothing}(grid)
```
glwagner: is `z0` roughnless length? we have another notation for that

examples/eth_canopy_bare_earth.jl
```diff
@@ -0,0 +1,179 @@
...
+# ## (3) Is the roughness closure right? Idealized verification
+#
+# The `z₀` map above is only as trustworthy as the closure behind it. Driving that same
+# closure with idealized inputs recovers its mathematical signatures and its physical
+# magnitudes — the behavior expected from Raupach (1994) and the class values tabulated by
+# Borak et al. (2025). `canopy_roughness(closure, Λ, h)` is the scalar form of the operator
+# that filled the `z₀` field above; it also accepts a `Field` or a `FieldTimeSeries`.
+
+closure = DragPartitionRoughness()                  # evergreen broadleaf forest defaults
+Λmax    = closure.parameters.maximum_area_index
```
glwagner: not sure I'm ready to use `Λ` for "area" 😂

examples/eth_canopy_bare_earth.jl
```diff
@@ -0,0 +1,179 @@
...
+
+# A west–east transect at 3.0°S: the DSM rides a canopy height above the bare-earth line.
+jrow = size(grid, 2) ÷ 2
+x = 1:size(grid, 1)
+fig = Figure(size = (1200, 440))
+ax = Axis(fig[1, 1]; xlabel = "west → east (grid cells)", ylabel = "elevation (m)",
+          title = "DSM vs bare-earth along 3°S: the gap is the ETH canopy")
+lines!(ax, x, Array(interior(z_dsm,  :, jrow, 1)); linewidth = 2, label = "DSM")
+lines!(ax, x, Array(interior(z_bare, :, jrow, 1)); linewidth = 2, label = "bare-earth")
+lines!(ax, x, Array(interior(removed, :, jrow, 1)); linewidth = 2, color = :seagreen, label = "removed canopy")
```
glwagner: The Array interior disease it is known as

ext/NumericalEarthArchGDALExt.jl
```diff
@@ -45,4 +104,161 @@ function NumericalEarth.DataWrangling.IBCAO.reproject_ibcao_to_netcdf(tiff_path,
...
+    try
+        return ArchGDAL.gdalwarp(datasets, options) do warped
+            gt = ArchGDAL.getgeotransform(warped)   # [x₀, Δλ, 0, y₀, 0, -Δφ]
+            band = Float32.(ArchGDAL.read(warped, 1))
+            Nx, Ny = size(band)
+            # GDAL rows run north→south; flip to south→north. Cell centers come from the
+            # geotransform, so they stay exact even when `-tr` snaps the extent to whole pixels.
+            data = reverse(band, dims = 2)
+            longitude = [gt[1] + (i - 0.5) * gt[2] for i in 1:Nx]
+            latitude  = [gt[4] + (Ny - j + 0.5) * gt[6] for j in 1:Ny]
```
glwagner: hmm

src/Lands/roughness/canopy_classes.jl
```diff
@@ -0,0 +1,93 @@
...
+representative_canopy_height(FT, ::Val{:savanna})                     = FT(9.02)
+representative_canopy_height(FT, ::Val{:grassland})                   = FT(1.39)
+representative_canopy_height(FT, ::Val{:permanent_wetland})           = FT(6.69)
+representative_canopy_height(FT, ::Val{:cropland})                    = FT(1.32)
+representative_canopy_height(FT, ::Val{:cropland_vegetation_mosaic})  = FT(1.40)
+
+"""
+$(TYPEDSIGNATURES)
+
+Prescribed momentum roughness length and zero-plane displacement `(z0, d0)` (meters) for a
```
glwagner: perhaps \ell^r for roughness and \ell^d for displacement. Open to others, but z0 and d0 are truly horrible

src/Lands/roughness/canopy_roughness_field.jl
```diff
@@ -0,0 +1,132 @@
...
+                              closure.sublayer_influence, closure.iterations)
+    gap = convert(FT, NaN)
+    return ifelse(valid, z0, gap), ifelse(valid, d0, gap)
+end
+
+@inline (closure::DragPartitionRoughness)(cell) = aerodynamic_parameters(closure, cell)
+
+@kernel function _compute_aerodynamic_roughness!(z0m, d0, closure, lai, canopy_height)
+    i, j = @index(Global, NTuple)
+    cell = (lai           = property_value(lai, i, j),
```
glwagner: Is `Λ` being proposed for lai? should find a mathematical notation and stick to it.

src/Lands/roughness/canopy_classes.jl
xkykai: what about `ℓᵐ` and `𝒹` hahah, don't really want \ell^d for displacement since it is not a "length" per se

glwagner: that script d is a little horrifying? plain "d" is ok with me or \delta? Not sure what is taken.

glwagner: d^c, c for "canopy"

xkykai: going with `d` because superscript has been used to imply "canopy". also renamed `SkinHumidity`'s thickness symbol `d` → `δˢ`, in line with `δᵛ` (dry_layer_depth) anyway.

src/Lands/roughness/canopy_roughness_field.jl
xkykai: changed to `leaf_area_index`

examples/eth_canopy_bare_earth.jl
xkykai: changed to `𝒜` hahah

examples/eth_canopy_bare_earth.jl
```diff
@@ -0,0 +1,183 @@
...
+#
+# DSM stand-in: ETOPO 2022 (token-free; `GLO30()` is the commercial-use 30 m surface model).
+# The ETH product is 10 m; `canopy_height_field` area-averages it onto the land grid straight
+# from the windowed COGs (anonymous `/vsicurl/`). Needs `using ArchGDAL` and `using CairoMakie`.
+
+using NumericalEarth
+using NumericalEarth.DataWrangling.ETHSentinel2Canopy: ETHSentinel2CanopyHeight, canopy_height_field
+using NumericalEarth.Lands: DragPartitionRoughness, compute_aerodynamic_roughness!,
+                            canopy_roughness, canopy_wind_ratio,
+                            canopy_drag_parameters, representative_canopy_height
```
glwagner: should these be exported by `using NumericalEarth.Lands`

src/Lands/roughness/igbp_canopy_classes.jl
glwagner, line 1: maybe igbp_canopy_heights.jl ?

src/Lands/roughness/canopy_roughness_closure.jl
```diff
@@ -0,0 +1,211 @@
...
+    form_drag_coefficient :: FT
+    "substrate (ground) drag coefficient `Cˢ`"
+    substrate_drag_coefficient :: FT
+    "maximum friction-to-wind ratio `(u★/Uh)ₘₐₓ`"
+    maximum_friction_ratio :: FT
+    "roughness-sublayer wind-profile decay coefficient `c`"
+    sublayer_decay_coefficient :: FT
+    "displacement coefficient `α`"
+    displacement_coefficient :: FT
+    "critical (skimming) leaf area index `𝒜ᶜ`"
```
glwagner: I think we have agreed not to put docstrings in the struct

src/Lands/roughness/canopy_roughness_closure.jl
```diff
@@ -0,0 +1,211 @@
...
+their `α` the `displacement_coefficient`).
+Monotonic and near-linear in the leaf area index, so it carries a clean seasonal signal.
+Uses the actual index (not capped at `𝒜ᶜ`). `𝒜ᶜ` limits only the wind ratio `γ`
+(the skimming cap on `u★/Uh`), while displacement keeps rising toward `h` as the canopy
+densifies — so `d` reaches the class-averaged values of Borak et al. (2025, Table 5).
+"""
+@inline function zero_plane_displacement(leaf_area_index, γ, h, p::DragPartitionParameters)
+    Cᴿ = p.form_drag_coefficient
+    Cˢ = p.substrate_drag_coefficient
+    𝒜  = max(leaf_area_index, 0)
```
glwagner, line 98: epsilon^la for "leaf area fraction" or delta, lambda...

src/Lands/roughness/canopy_roughness_closure.jl
```diff
@@ -0,0 +1,211 @@
...
+    parameters :: DragPartitionParameters{FT}
+    "representative canopy height (m), the fallback where no measured height is supplied"
+    representative_height :: FT
+    "von Kármán constant `ϰ`"
+    von_karman_constant :: FT
+    "roughness-sublayer influence (constant 0.193; [Raupach 1995](@cite raupach1995corrigenda))"
+    sublayer_influence :: FT
+    "data-quality ceiling on the leaf area index; a larger index is treated as fill/artifact and gapped"
+    maximum_valid_leaf_area_index :: FT
+    "fixed-point iterations for the wind ratio `γ`"
```
glwagner: same comment regarding the docstring

glwagner (approved):
still brainstorming a symbol for "leaf area index"

src/Lands/roughness/canopy_roughness_closure.jl
bgroenks96: Why?

src/Lands/roughness/canopy_roughness_closure.jl
xkykai: sticking with `𝒜`

glwagner (approved):

src/Lands/roughness/canopy_roughness_closure.jl
glwagner: Because it makes the code harder to read. We decided that it was nicer to have comments in-line, eg

```julia
struct MyStruct
    a # what a is
    b # what b is
end
```

this matches the convention in Oceananigans as well, so we have a uniform convention throughout. At the time long ago, both conventions were prevalent in Oceananigans. We went through and changed every struct to follow the comment convention.

glwagner: from a legibility perspective, it's hard to read / glance at which lines are code (distinguished only by the `::` mark) and which lines are docstring (inside `"`).

ext/NumericalEarthArchGDALExt/ethcanopy.jl
```diff
@@ -135,7 +135,7 @@ end
 # array matches the grid's cell count.
 function canopy_field(grid, data)
     h = Field{Center, Center, Nothing}(grid)
-    interior(h, :, :, 1) .= data
+    interior(h) .= on_architecture(architecture(grid), reshape(data, size(data, 1), size(data, 2), 1))
```
glwagner, line 138: does `set!(h, data)` not work here? may be simipler just to fix `set!` first.

PR #419, opened 2026-07-07
Per-column, GPU-safe surface- and boundary-layer heights for Breeze coupling

### LINKED ISSUE #379: `surface_layer_height(::BreezeAtmosphere)` scalar-indexes `Δz` → GPU "scalar indexing is disallowed" on stretched / terrain-following vertical grids
## Summary

`surface_layer_height(::BreezeAtmosphere)` reads a single vertical spacing on the **host** via
`zspacing(1, 1, 1, grid, …)`. On a uniform vertical grid that returns a `Number` (fine), but on a
**stretched or terrain-following** grid `Δz` is a device array, so the read is a host-side scalar index
of a `CuArray` → Oceananigans' `disallow_scalar` guard throws. This makes any GPU `AtmosphereLandModel`
(or atmosphere↔sea-ice) run with a realistic vertical grid crash on the first coupled `update_state!`.

## Where

`ext/NumericalEarthBreezeExt/breeze_atmosphere_interface.jl:30`:

```julia
# Note: for stretched grids on GPU, this may require allowscalar.
function NumericalEarth.EarthSystemModels.surface_layer_height(atmosphere::BreezeAtmosphere)
    grid = ...
    return Oceananigans.zspacing(1, 1, 1, grid, Center(), Center(), Center()) / 2
end
```

It is called host-side, once per coupled step, in `compute_atmosphere_land_fluxes!`
(`src/EarthSystemModels/InterfaceComputations/atmosphere_land_fluxes.jl:64`, consumed as the scalar
`zᵃᵗ` at `:174`). **The atmosphere↔sea-ice path does the same** (`atmosphere_sea_ice_fluxes.jl:10,109`),
so this is not land-specific — any Breeze-atmosphere surface coupling on a GPU stretched grid hits it.

## Reproduce

Build a Breeze `AtmosphereModel` on a `GPU()` grid with a stretched (or `TerrainFollowingVerticalDiscretization`)
`z`, attach a `SlabLand` via `AtmosphereLandModel`, and step it:

```
ERROR: Scalar indexing is disallowed.
... getindex(::CuArray, ::Int) ...
[in] surface_layer_height(::BreezeAtmosphere)  @ breeze_atmosphere_interface.jl:32
```

Never caught to date because the only coupled example/test (`examples/breeze_over_slab_land.jl`, the
coarse-LES surface-layer demo) uses **uniform Δz** on GPU, and `examples/era5_breeze.jl` (stretched +
terrain) has no land interface.

## Impact

- Blocks atmosphere↔land/sea-ice surface fluxes on GPU for **any vertically-stretched grid** — i.e.
  essentially every realistic atmosphere (boundary-layer resolution requires near-surface stretching).
- **Secondary correctness bug:** even where it runs, it returns the `(1, 1, 1)` corner column's
  half-`Δz` and applies it **domain-wide**. On a terrain-following grid the first-cell thickness varies
  per column, so the surface-layer reference height is wrong away from that corner.

The `allowscalar` workaround in the code comment removes the crash but is both slow (host read every
step) and keeps the domain-wide-from-one-column error — so it is not a real fix.

## Proposed fix

Make the surface-layer reference height **per-column and GPU-safe**: drop the host-side scalar
`surface_layer_height` property and instead compute `zᵃᵗ = ½ · Δz(i, j, 1)` per surface cell **inside**
the `_compute_atmosphere_*_interface_state!` kernels (each thread reads its own column — GPU-safe and
terrain-correct). The flux kernels already receive `grid`, so this removes the host scalar read
entirely. (If a scalar API is still wanted for uniform grids, restrict it to that case.)

## References

- Surfaced while wiring surface drag into the ERA5 dynamical-downscaling demo (#273); the SlabLand
  coupling lives in #220 (this is not in its "Known limitations" list).

ewquon:
Closes NumericalEarth/NumericalEarth.jl#379.

`surface_layer_height(::BreezeAtmosphere)` read `zspacing(1,1,1,grid,…)` on the host. On a<br>stretched or terrain-following grid `Δz` is a device array, so this scalar-indexes a `CuArray`<br>→ "scalar indexing is disallowed" on the first coupled `update_state!`; and even where it ran it<br>applied one corner column's half-`Δz` domain-wide, so the MOST reference height was wrong per<br>column on terrain (both failure modes are described in NumericalEarth/NumericalEarth.jl#379).

This makes the surface-layer reference height per-column and GPU-safe, and does the same for the<br>boundary-layer height:

* `surface_layer_height` is now a per-column 2-D `Field` for Breeze atmospheres, materialized<br>on-device by an `:xy` fill kernel (`½·Δz(i,j,1)`) and cached once in `interfaces.properties`<br>(the grid geometry is static — no per-step host read). A new 2-arg<br>`surface_layer_height(atmosphere, exchange_grid)` has a generic fallback returning the existing<br>scalar, so prescribed/Speedy atmospheres are unchanged.
* The three flux kernels (land/ocean/sea-ice) read `zᵃᵗ` and `h_bℓ` through a shared<br>`field_value(x,i,j)` accessor that transparently handles a scalar or a `Field`, so the scalar<br>path is bit-identical and the field path is per-column. (`field_value` is promoted from the<br>existing land-local `land_field_value`.)
* `boundary_layer_height(::BreezeAtmosphere)` returns the closure's live per-column `zi` field<br>when it diagnoses one (duck-typed on the closure fields), falling back to the 600 m constant.<br>Refreshed each step (not cached, since it evolves).
* Widen Breeze compat to include 0.7.

Why cache a field rather than compute `½·Δz(i,j,1)` inline in the kernel (as NumericalEarth/NumericalEarth.jl#379 suggests): the<br>flux kernels are shared with prescribed atmospheres, whose reference height is a fixed *measurement*<br>height unrelated to the exchange-grid `Δz`. The scalar-vs-`Field` polymorphism keeps both correct.

### Verification

* **CPU:** `field_value` units; stretched-grid `AtmosphereLandModel` field path (per-column `Field`,<br>nonzero surface fluxes); terrain-following reference height varies per column and matches the<br>Gal-Chen Jacobian `½·Δr₁·(1−h/H)` within 2 %; scalar fallback preserved.
* **GPU (Tesla T4):** construction + `update_state!` complete under `CUDA.allowscalar(false)` on<br>both stretched and terrain-following grids (the crash case in NumericalEarth/NumericalEarth.jl#379); reference-height field is<br>per-column; fluxes finite.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)


ext/NumericalEarthBreezeExt/breeze_atmosphere_interface.jl
```diff
@@ -25,18 +26,55 @@ NumericalEarth.EarthSystemModels.thermodynamics_parameters(atmos::BreezeAtmosphe
...
-function NumericalEarth.EarthSystemModels.surface_layer_height(atmosphere::BreezeAtmosphere)
-    grid = atmosphere.grid
-    return Oceananigans.zspacing(1, 1, 1, grid, Center(), Center(), Center()) / 2
+# The MOST reference height is the lowest cell-center elevation, Δz(i,j,1)/2. It is filled
+# per column on-device (co-indexed with the atmosphere grid via the direct index-copy in
+# `interpolate_state!`), which sidesteps the host-side scalar read into a device Δz array
+# (the GPU crash in #379). Built once and cached in `interfaces.properties` (grid geometry
+# is static). Only a terrain-following grid makes the first-cell height vary per column;
+# on any other grid (uniform or vertically stretched) it is horizontally uniform, so we
+# collapse it to a scalar (the consumer `field_value` handles either transparently).
```
ewquon:
```suggestion
# The MOST reference height is the lowest cell-center elevation, Δz(i,j,1)/2,
# filled per column on-device.
```

ext/NumericalEarthBreezeExt/breeze_atmosphere_interface.jl
```diff
@@ -25,18 +26,50 @@ NumericalEarth.EarthSystemModels.thermodynamics_parameters(atmos::BreezeAtmosphe
...
+# terrain-following grid has no well-defined domain-wide surface-layer height, so
+# callers must use the grid-aware `(atmosphere, exchange_grid)` form above. (The
+# generic scalar fallback in EarthSystemModels serves prescribed/Speedy atmospheres,
+# whose reference height is a fixed measurement height.)
+
+# The boundary-layer height is diagnosed per column by the turbulence closure when it
+# provides one (e.g. Breeze's ScaleAdaptiveTKE writes z_i into its `zi` closure field);
+# return that live field so the flux kernels read it per column via `field_value`. Fall
+# back to a 600 m constant otherwise. Duck-typed on the closure fields so this works
+# across closures and Breeze versions. Not cached: z_i evolves each step.
```
ewquon:
```suggestion
# The boundary-layer height is diagnosed per column by the turbulence closure
# at each step when it provides one. Fall back to a 600 m constant otherwise.
```

src/EarthSystemModels/InterfaceComputations/component_interfaces.jl
```diff
@@ -449,7 +449,13 @@ function ComponentInterfaces(atmosphere, ocean, sea_ice=nothing;
     exchanger = StateExchanger(exchange_grid, radiation, atmosphere, land, ocean, sea_ice;
                                atmosphere_correction = exchanger_correction)
 
-    properties = (; gravitational_acceleration)
+    # The surface-layer (MOST reference) height is fixed by the atmosphere grid, so
+    # build it once here rather than per coupled step. Scalar for prescribed
+    # atmospheres; a per-column 2-D field for grid-aware atmospheres (Breeze). The
```
ewquon, line 474: Any issue with this, @glwagner?

glwagner: I think a fixed height is a good idea!

One thing we may want to support in the future is to use a different prescribed height (via interpolation) than simply `dz/2`. I am also not sure if anything additional (or less?) would be needed for https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2018MS001534

src/EarthSystemModels/InterfaceComputations/atmosphere_sea_ice_fluxes.jl
```diff
@@ -106,15 +106,15 @@ end
 
     # Build thermodynamic and dynamic states in the atmosphere and interface.
     ℂᵃᵗ = atmosphere_properties.thermodynamics_parameters
-    zᵃᵗ = atmosphere_properties.surface_layer_height
+    zᵃᵗ = field_value(atmosphere_properties.surface_layer_height, i, j)
```
glwagner: there is another function `stateindex` which looks fairly similar to this...

I suggest calling this function `*index` to be reminiscent of `getindex`, which has the form `getindex(a, i, j)` as an example. could be `maybeindex(f, i, j)`?

src/EarthSystemModels/InterfaceComputations/InterfaceComputations.jl
```diff
@@ -102,6 +102,14 @@ function interface_kernel_parameters(grid)
     return kernel_parameters
 end
 
+# Per-cell value from a scalar constant or a 2-D `Field`. A `Number` (e.g. a
+# prescribed measurement height or the 600 m BL-height fallback) broadcasts to
+# every column; a `Field` (e.g. Breeze's per-column surface- or boundary-layer
+# height) is read at column `(i, j)`. Used by the atmosphere–surface flux kernels
+# to consume `surface_layer_height` / `h_bℓ` uniformly.
+@inline field_value(x::Number, i, j) = x
+@inline field_value(x, i, j) = @inbounds x[i, j, 1]
```
glwagner: as mentioned above, another possible name for this is `maybeindex(f, i, j)`, which is reminiscent of `getindex`.

glwagner: may want to additionally indicate that this is 2D specific

src/EarthSystemModels/InterfaceComputations/component_interfaces.jl
ewquon: I've opened #420 :)

src/EarthSystemModels/InterfaceComputations/InterfaceComputations.jl
ewquon: Following your `*index` convention, I implemented `state2dindex` that wraps `stateindex` for a 2d field

glwagner: it isn't my convention, it's julia's! :-D

ewquon:
Heads-up on the `main` merge here: #323 added an `@allowscalar` `surface_layer_height(::BreezeAtmosphere)` to get the GPU coupling running. This merge **drops** it in favor of this PR's field-based version — a deliberate resolution, not an accident.

Why: #323's fix clears the GPU scalar-indexing **crash** (mode 1 of #379), but not the second half — it still reads column `(1,1)` (`zspacing(1,1,1,…)`) and applies that half-`Δz` **domain-wide**, so the MOST reference height is wrong per column on stretched/terrain-following grids (#379's "secondary correctness bug"). It's also the per-step host read that #379 explicitly called "not a real fix." This PR replaces it with a per-column, on-device field (terrain-correct, no `@allowscalar`), so keeping both would just be a redundant, less-correct path.

No regression for `main`: the GPU crash stays fixed, and terrain columns now get the correct reference height.

cc @glwagner

ext/NumericalEarthBreezeExt/breeze_atmosphere_interface.jl
```diff
@@ -26,20 +26,41 @@ NumericalEarth.EarthSystemModels.thermodynamics_parameters(atmos::BreezeAtmosphe
...
-# the terrain metrics), so `zspacing` scalar-indexes a device array — wrap the single
-# host read in `@allowscalar`. This runs once per `update_state!`, not per cell.
-function NumericalEarth.EarthSystemModels.surface_layer_height(atmosphere::BreezeAtmosphere)
-    grid = atmosphere.grid
-    return @allowscalar Oceananigans.zspacing(1, 1, 1, grid, Center(), Center(), Center()) / 2
+# The MOST reference height is the lowest cell-center elevation, Δz(i,j,1)/2,
+# filled per column on-device.
+@kernel function _fill_surface_layer_height!(zref, atmos_grid)
+    i, j = @index(Global, NTuple)
+    @inbounds zref[i, j, 1] = Oceananigans.zspacing(i, j, 1, atmos_grid, Center(), Center(), Center()) / 2
```
ewquon: Use `znode`

ewquon: And use `z_1` instead of `zref`

ext/NumericalEarthBreezeExt/breeze_atmosphere_interface.jl
```diff
@@ -26,20 +26,41 @@ NumericalEarth.EarthSystemModels.thermodynamics_parameters(atmos::BreezeAtmosphe
...
+    zref = Oceananigans.Field{Center, Center, Nothing}(exchange_grid)
+    arch = architecture(exchange_grid)
+    launch!(arch, exchange_grid, interface_kernel_parameters(exchange_grid),
+            _fill_surface_layer_height!, zref, atmosphere.grid)
+    # Per-column field only where terrain makes it vary; otherwise one uniform value,
+    # read off the device via a bulk host copy (no scalar GPU indexing).
+    if atmosphere.grid isa TerrainFollowingGrid
+        return zref
+    else
+        return first(Array(Oceananigans.interior(zref)))
```
glwagner:
```suggestion
        return @allowscalar znode(1, 1, 1, grid, Center(), Center(), Center())
```

ext/NumericalEarthBreezeExt/breeze_atmosphere_interface.jl
```diff
@@ -26,20 +26,41 @@ NumericalEarth.EarthSystemModels.thermodynamics_parameters(atmos::BreezeAtmosphe
...
+end
+
+function NumericalEarth.EarthSystemModels.surface_layer_height(atmosphere::BreezeAtmosphere, exchange_grid)
+    zref = Oceananigans.Field{Center, Center, Nothing}(exchange_grid)
+    arch = architecture(exchange_grid)
+    launch!(arch, exchange_grid, interface_kernel_parameters(exchange_grid),
+            _fill_surface_layer_height!, zref, atmosphere.grid)
+    # Per-column field only where terrain makes it vary; otherwise one uniform value,
+    # read off the device via a bulk host copy (no scalar GPU indexing).
+    if atmosphere.grid isa TerrainFollowingGrid
```
ewquon: Fix this code flow for efficiency

ext/NumericalEarthBreezeExt/breeze_atmosphere_interface.jl
```diff
@@ -26,20 +26,41 @@ NumericalEarth.EarthSystemModels.thermodynamics_parameters(atmos::BreezeAtmosphe
...
 end
 
-NumericalEarth.EarthSystemModels.surface_layer_height(atmos::BreezeAtmosphereSim) =
-    NumericalEarth.EarthSystemModels.surface_layer_height(component_model(atmos))
+NumericalEarth.EarthSystemModels.surface_layer_height(atmos::BreezeAtmosphereSim, exchange_grid) =
+    NumericalEarth.EarthSystemModels.surface_layer_height(component_model(atmos), exchange_grid)
+
+# The boundary-layer height is diagnosed per column by the turbulence closure
+# at each step when it provides one. Fall back to a 600 m constant otherwise.
+diagnosed_boundary_layer_height(closure_fields) = 600
```
ewquon: wtf

glwagner (approved):

ext/NumericalEarthBreezeExt/breeze_atmosphere_interface.jl
ewquon: Note: `znode` gives an absolute height rather than height above ground level — sticking with `zspacing/2`

PR #421, opened 2026-07-07
Opt-in Jiménez (2012) land surface-layer stability functions

ewquon:
Adds the Jiménez et al. (2012) revised-MM5 stability functions as opt-in, pluggable types for `SimilarityTheoryFluxes`. No defaults change: Large–Yeager remains the land default, and nothing existing is touched — the diff is purely additive.

## What

- `FairallMomentumStabilityFunction` / `FairallScalarStabilityFunction` — unstable side: the Fairall et al. (1996) blend of the Kansas-type Paulson (1970) form with the convective form of Grachev et al. (2000) (`aᶜ` = 10 momentum / 34 scalar), using the `f = ζ²/(1+ζ²)` blend already familiar from the Edson functions.
- `ChengBrutsaertMomentumStabilityFunction` / `ChengBrutsaertScalarStabilityFunction` — stable side: Cheng & Brutsaert (2005), `ψ = -a ln[ζ + (1+ζᵇ)^(1/b)]`, milder than the linear Large–Yeager form at strong stability.
- Combined by the exported `jimenez_stability_functions(FT)` factory; each function clamps ζ to its regime and returns exactly zero outside it, so `SplitStabilityFunction` can evaluate both branches NaN-free.

Usage, composing with the velocity corrections from #439 for a full Jiménez-style coarse-grid configuration:

```julia
fluxes = SimilarityTheoryFluxes(FT;
    stability_functions = jimenez_stability_functions(FT),
    subgrid_velocities = SubgridVelocityCorrection(FT; mesoscale = mahrt_sun_subgrid_velocity(Δx)))
```

## Testing

- New `test/test_jimenez_stability_functions.jl`: point values at ζ ∈ {−5, −1, −0.1, 0, 0.1, 1, 10} against independently written reference formulas, regime-sign sanity, both-branch NaN safety across ζ = 0, CB05-milder-than-linear at ζ = 10, and Float32 type preservation.
- End-to-end coupled testset in `test/test_slab_land.jl`: finite, sign-correct fluxes and monotonic u★ across warm/neutral/cold forcing with the opt-in functions (CPU+GPU; verified on a T4).
- Formulas verified to machine precision against the WRF v4.3 reference implementation (`phys/module_sf_sfclayrev.F`).

## History

An earlier revision of this PR also carried a standalone Mahrt–Sun subgrid-velocity field; that piece was redesigned and landed via #439 (`SubgridVelocityCorrection`), and this branch has been rebased onto it — only the stability functions remain here.


simone-silvestri:
I think the `subgrid_velocity` concept is intertwined to the gustiness.
What do you think about generalizing the `gustiness` field as to not add another field to the SimilarityFluxes type?

ewquon:
> I think the `subgrid_velocity` concept is intertwined to the gustiness. What do you think about generalizing the `gustiness` field as to not add another field to the SimilarityFluxes type?

This is a good suggestion @simone-silvestri. I'm not sure `gustiness` is the right umbrella term but I think it would be a good idea to have a `SubgridVelocityCorrection(;convective, mesoscale)` that allows for `ConvectiveGustiness` (from Beljaars, based on w* with the two existing parameters) and a simpler mesoscale correction (from Mahrt–Sun, based on dx). I can open a separate issue to provide more detail if you think this makes sense.

PR #422, opened 2026-07-07
Bump Reactant compat to 0.2.270

xkykai:
Bumps the `Reactant` compat bound to `0.2.235 - 0.2.270` and moves the test-environment to `0.2.270`.

0.2.270 fixes an LLVM SLP-vectorizer regression (present in the 0.2.256–0.2.268 window) that miscompiled Reactant-raised kernels into an invalid `select` with a `vector<2xi1>` mask and scalar `f64` operands, which failed MLIR → LLVM IR translation (`could not convert to LLVM IR`). With 0.2.270 the coupled-model Reactant path compiles cleanly again with no vectorizer workaround, and the coupled `AtmosphereLandModel` first-step result is bit-exact vs the eager CPU reference. See https://github.com/NumericalEarth/NumericalEarth.jl/pull/405#discussion_r3539300537 for more details


dkytezab:
Hmm, can we bump the compat in another fashion (not just updating the upper bound)? Then we don't have to repeatedly change this for new releases. We could probably just set 0.2.269 (or 270, not sure which has the required changes) as the min version, as otherwise if Reactant is restricted to an earlier version by a different package a user will hit the same issues we've seen here

dkytezab:
> can we bumping

hmm

xkykai:
> Hmm, can we bump the compat in another fashion (not just updating the upper bound)? Then we don't have to repeatedly change this for new releases. We could probably just set 0.2.269 (or 270, not sure which has the required changes) as the min version, as otherwise if Reactant is restricted to an earlier version by a different package a user will hit the same issues we've seen here

sounds good, I'll wait to see if tests pass (which is defined for 0.2.270 and above). if all tests pass safe to say we don't need the lower version compat anymore

xkykai:
@dkytezab it appears that most tests (except data downloading) pass with 0.2.270. I've updated Project.toml and it is ready to merge

test/Project.toml
```diff
@@ -50,7 +50,7 @@ CopernicusMarine = "0.2"
 Dates = "<0.0.1, 1"
 Downloads = "<0.0.1, 1"
 ExplicitImports = "1.15"
-GPUCompiler = "~1.9"
+GPUCompiler = "~1.9, ~1.22"
```
dkytezab: Was there a reason for the upper bound here (and above with CUDA)? Otherwise looks good

dkytezab (approved):

test/Project.toml
xkykai: let me bump to 
```julia
CUDA = "5.9.5, 6"
GPUCompiler = "1"
```
and see what happens to tests

PR #426, opened 2026-07-09
Add OpenLandMap-soilDB 30 m soil texture & bulk density dataset

xkykai:
This PR is based on top of https://github.com/NumericalEarth/NumericalEarth.jl/pull/323/ so is intended to merge only after https://github.com/NumericalEarth/NumericalEarth.jl/pull/323/.

Adds `OpenLandMapSoilDB`: a DataWrangling dataset for OpenLandMap-soilDB 30 m soil texture (sand/silt/clay) and bulk density, over the three native depth intervals (0–30, 30–60, 60–100 cm). These are the inputs a pedotransfer function needs to derive van Genuchten parameters for `VariablySaturatedHydrology`.

Built on the same pattern as `SoilGrids2`/`GLO30`. The one new piece is a windowed `/vsicurl` cloud-optimized-GeoTIFF reader in the ArchGDAL extension: it masks nodata → NaN, applies the band scale/offset, and stacks the three depths into a regional NetCDF. A `BoundingBox` is required (the global 30 m grid is ~1.44M × 528k cells).

```julia
using NumericalEarth, ArchGDAL
region = BoundingBox(longitude = (-112.3, -111.9), latitude = (36.0, 36.4))
clay = Field(Metadatum(:clay_fraction; dataset = OpenLandMapSoilDB(), region), CPU())
```

Not included yet: the pedotransfer function (texture → van Genuchten parameters).


xkykai:
running 
```julia
using NumericalEarth   # OpenLandMapSoilDB, BoundingBox, Metadatum
using Oceananigans     # Field, CPU, interior, λnodes, φnodes, Center
using ArchGDAL         # activates the windowed cloud-optimized-GeoTIFF reader
using CairoMakie

region = BoundingBox(longitude = (-112.3, -111.9), latitude = (36.0, 36.4))

# Native 30 m horizontal window × three depth intervals (0–30, 30–60, 60–100 cm),
# read straight from the cloud-optimized GeoTIFFs. No credentials needed.
panels = [(:sand_fraction, "sand fraction", "kg/kg", :YlOrBr),
          (:silt_fraction, "silt fraction", "kg/kg", :YlGnBu),
          (:clay_fraction, "clay fraction", "kg/kg", :OrRd),
          (:bulk_density,  "bulk density",  "kg/m³", :dense)]

fields = map(p -> Field(Metadatum(p[1]; dataset = OpenLandMapSoilDB(), region), CPU()), panels)

grid = first(fields).grid
lon = Array(λnodes(grid, Center()))
lat = Array(φnodes(grid, Center()))
ksurf = 3   # surface 0–30 cm layer (depths stored deepest-first)

fig = Figure(size = (1100, 980), fontsize = 15)
Label(fig[0, 1:2], "OpenLandMap-soilDB 30 m — 0–30 cm — Grand Canyon window"; fontsize = 18, font = :bold)

for (n, ((_, title, unit, cmap), field)) in enumerate(zip(panels, fields))
    A = Array(interior(field))[:, :, ksurf]
    finite_pct = round(100 * count(isfinite, A) / length(A); digits = 1)
    row, col = fldmod1(n, 2)
    ax = Axis(fig[row, col]; title = "$title  ($finite_pct% finite)",
              xlabel = "longitude (°)", ylabel = "latitude (°)", aspect = DataAspect())
    lo, hi = extrema(filter(isfinite, A))
    hm = heatmap!(ax, lon, lat, A; colormap = cmap, colorrange = (lo, hi),
                  nan_color = RGBAf(0.85, 0.85, 0.85, 1))
    Colorbar(fig[row, col][1, 2], hm; label = unit)
end

save("openlandmap_soildb_texture_map.png", fig)
```
gives
openlandmap_soildb_texture_map

ext/NumericalEarthArchGDALExt.jl
```diff
@@ -45,4 +48,126 @@ function NumericalEarth.DataWrangling.IBCAO.reproject_ibcao_to_netcdf(tiff_path,
...
+# GDAL's bundled libcurl may not find a CA bundle (breaking https /vsicurl reads);
+# point it at a CA file, honoring the usual override env vars first, then Julia's
+# bundled roots, then the common system locations.
+function vsicurl_ca_bundle()
+    for var in ("SSL_CERT_FILE", "CURL_CA_BUNDLE", "GDAL_HTTP_CAINFO")
+        path = get(ENV, var, "")
+        !isempty(path) && isfile(path) && return path
+    end
+    candidates = [joinpath(Sys.BINDIR, "..", "share", "julia", "cert.pem"),
+                  "/etc/ssl/certs/ca-certificates.crt",
```
bgroenks96: This is OS specific. Is the assumption that non-linux systems are covered by the first candidate?

ext/NumericalEarthArchGDALExt.jl
```diff
@@ -45,4 +48,126 @@ function NumericalEarth.DataWrangling.IBCAO.reproject_ibcao_to_netcdf(tiff_path,
...
+            x0, dx, _, y0, _, dy = ArchGDAL.getgeotransform(ds)
+            width  = ArchGDAL.width(ds)
+            height = ArchGDAL.height(ds)
+
+            xoff  = clamp(floor(Int, (W - x0) / dx), 0, width - 1)
+            yoff  = clamp(floor(Int, (N - y0) / dy), 0, height - 1)
+            xsize = clamp(ceil(Int, (E - x0) / dx) - xoff, 1, width - xoff)
+            ysize = clamp(ceil(Int, (S - y0) / dy) - yoff, 1, height - yoff)
+
+            band   = ArchGDAL.getband(ds, 1)
```
bgroenks96: As I understand it, this is supposed to be a general extension providing ArchGDAL support, so probably this should not be hardcoded.

ext/NumericalEarthArchGDALExt.jl
```diff
@@ -45,4 +48,126 @@ function NumericalEarth.DataWrangling.IBCAO.reproject_ibcao_to_netcdf(tiff_path,
...
+    W, E = bbox.longitude
+    S, N = bbox.latitude
+
+    layers = Matrix{Float32}[]
+    longitude = Float64[]
+    latitude  = Float64[]
+
+    for source in sources
+        ArchGDAL.read(source) do ds
+            # EPSG:4326 geotransform: [x₀, Δλ, 0, y₀, 0, Δφ] with Δφ < 0 (north→south).
```
bgroenks96: I thought the standard in NumericalEarth (also by CF conventions) is south-to-north? Looks like this is also the case for [WGS84](https://spatialreference.org/ref/epsg/4326/).

And what exactly is this comment referring to? Are we *assuming* that `getgeotransform` will return the EPSG:4326 transform? If so, that should probably be more clearly documented and maybe we should see if we can somehow fail loudly if it doesn't match.

ext/NumericalEarthArchGDALExt.jl
```diff
@@ -45,4 +48,126 @@ function NumericalEarth.DataWrangling.IBCAO.reproject_ibcao_to_netcdf(tiff_path,
...
+            band   = ArchGDAL.getband(ds, 1)
+            scale  = ArchGDAL.getscale(band)
+            offset = ArchGDAL.getoffset(band)
+            nodata = ArchGDAL.getnodatavalue(band)
+
+            raw = ArchGDAL.read(ds, 1, xoff, yoff, xsize, ysize)  # (lon, lat), north-first
+            decoded = reverse(decode_cog_window(raw, scale, offset, nodata), dims = 2)
+            push!(layers, decoded)
+
+            if isempty(longitude)
```
bgroenks96: Can you explain what this is doing? Why the `-0.5` and why do we only set `latitude` and `longitude` when `longtidue` is empty? Isn't it trivially empty at this point? It hasn't been set yet AFAICT.

bgroenks96 (approved):

bgroenks96 (commented):
Oops, accidentally approved before when I just wanted to leave comments. But I will approve once those comments are resolved.

ext/NumericalEarthArchGDALExt.jl
xkykai: Good point! I've changed to universal
https://github.com/NumericalEarth/NumericalEarth.jl/blob/fee6fbda10cbd1011515e869cae14f45d0505ce9/ext/NumericalEarthArchGDALExt.jl#L66

ext/NumericalEarthArchGDALExt.jl
xkykai: should be addressed now with `validate_geographic_northup` function

ext/NumericalEarthArchGDALExt.jl
xkykai: moved `OpenLandMap`-specific functionality to `OpenLandMap.jl`
https://github.com/NumericalEarth/NumericalEarth.jl/blob/fd2985b3ec4aeb0f8e04375971473faee856ea07/src/DataWrangling/OpenLandMap/OpenLandMap.jl#L221

ext/NumericalEarthArchGDALExt.jl
xkykai: sorry for the magic number, it's because we are moving from 0-indexed cell corner to 1-indexed cell center. code comments have been made clearer:
https://github.com/NumericalEarth/NumericalEarth.jl/blob/df91f0c8c2e31d1e00a6cf33df01257b4a055092/ext/NumericalEarthArchGDALExt.jl#L136-L138

glwagner (approved):

PR #435, opened 2026-07-13
Fix `PrescribedAtmosphere` import in ERA5 module

simone-silvestri:
At the moment we have the following warning:
```julia
┌ NumericalEarth
│  WARNING: Constructor for type "PrescribedAtmosphere" was extended in `ERA5` without explicit qualification or import.
│    NOTE: Assumed "PrescribedAtmosphere" refers to `Atmospheres.PrescribedAtmosphere`. This behavior is deprecated and may differ in future versions.`
│    NOTE: This behavior may have differed in Julia versions prior to 1.12.
│    Hint: If you intended to create a new generic function of the same name, use `function PrescribedAtmosphere end`.
│    Hint: To silence the warning, qualify `PrescribedAtmosphere` as `Atmospheres.PrescribedAtmosphere` in the method signature or explicitly `import Atmospheres: PrescribedAtmosphere`.
```

### DIFF
--- a/src/DataWrangling/ERA5/ERA5_prescribed_atmosphere.jl
+++ b/src/DataWrangling/ERA5/ERA5_prescribed_atmosphere.jl
@@ -6,17 +6,13 @@ using Oceananigans.Architectures: on_architecture
 using Oceananigans.Fields: CenterField, interior
 using Thermodynamics: Liquid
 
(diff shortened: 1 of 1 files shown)


taimoorsohail (approved):
LGTM

PR #346, opened 2026-06-12
Tracer budget closure test for `TripolarGrid` and fix a `RightFaceFolded` bug

taimoorsohail:
This PR adds a tracer conservation test for a `TripolarGrid` setup. Note that at the moment the test is quite heavy. I would appreciate your advice in making it a bit less heavy, but I wasn't sure what components are important (or not) in ensuring conservation. We know tracers are conserved in Oceananigans.jl, so if there is an issue it is likely in the coupling with other datasets. Hence I took a maximalist approach. Let's work on making this PR more lightweight but still revealing if there is a budget non-closure.

PR #186, opened 2026-04-23
Add an energy conservation example and test

simone-silvestri:
This PR introduces an energy conservation example and a test for energy conservation.
It depends on the fixes scattered through here and ClimaSeaIce (#164 and https://github.com/CliMA/ClimaSeaIce.jl/pulls).
 
The freshwater flux is also calculated but not balanced, so when the issue of energy conservation is resolved and this is merged I will focus on tracking down mass conservation

PR #436, opened 2026-07-13
Ensure Freshwater conservation

simone-silvestri:
With #186 and  #346 we have ensured that the coupled system conserves energy down to machine precision.
This PR adds freshwater conservation to the mix.

It depends on a couple of PRs: 
- https://github.com/CliMA/Oceananigans.jl/pull/5788
- https://github.com/NumericalEarth/NumericalEarth.jl/pull/244
- https://github.com/CliMA/ClimaSeaIce.jl/pull/166


taimoorsohail (approved):
I am not across the code 100% but I think what you have done is not applicable to fixed vertical grids, correct? Otherwise it looks good.

simone-silvestri:
> I am not across the code 100% but I think what you have done is not applicable to fixed vertical grids, correct? Otherwise it looks good.

We probably need to fix a couple of things to make sure the budget closes for static grids.

PR #440, opened 2026-07-14
Clean up testing infrastructure

simone-silvestri:
our testing suite has ballooned over 1hr of testing time. This PR tries to reduce the testing time. 
I will add here the numbered fixes as they come in then we can see what effect this produced


test/test_tracer_budget.jl
```diff
@@ -90,25 +90,27 @@ end
             radiation  = JRA55PrescribedRadiation(arch; time_indices_in_memory)
             atmosphere = JRA55PrescribedAtmosphere(arch; time_indices_in_memory)
 
-            ecco_set = MetadataSet(:temperature, :salinity,
-                                   dataset = ECCO4Monthly(),
-                                   date = DateTime(1993, 1, 1))
+            # An idealized, stably stratified initial state: a warm, slightly fresh surface over
```
taimoorsohail: Should we apply this to the sea ice test below too?

simone-silvestri: yes that would be great

taimoorsohail: OK I'll work on that

xkykai:
perhaps related to this PR (or not) is that @glwagner and I were thinking if data downloading tests should be placed onto a weekly/nightly CI with their statuses reported on a website so as to not burden the data servers too much

glwagner:
yes, that would be a good addition to this PR

simone-silvestri:
Ok, I have created a nightly downloading build that will publish a "status" page in the NumericalEarth gh-pages branch with the status of the download sources.

It is possible to enable the build on a PR by adding the "run downloading tests" label. That will publish the status on previews/PR#

.github/scripts/build_dataset_status.jl
```diff
@@ -0,0 +1,210 @@
...
+
+struct DatasetRecord
+    platform :: String
+    dataset :: String
+    variable :: String
+    succeeded :: Bool
+    message :: String
+    seconds :: Float64
+    bytes :: Int
+end
```
glwagner: this is superceded by a weekly chron (which is probably better than nightly?)

giordano:
@simone-silvestri may I suggest to look at #647 + #648 + #651 + #653 before pushing forward with this?  They take a different approach of avoiding network access as much as possible (after all of these PRs are merged vast majority of tests, and all of those that run in CI for PRs in particular, won't need Internet access at all, improving reliability of our testing infrastructure)

.github/workflows/ci.yml
```diff
@@ -73,6 +73,15 @@ jobs:
               using Pkg;
               Pkg.instantiate()
           '
+      # ParallelTestRunner schedules longest-first from a duration history it keeps in a
+      # scratchspace. Without it every duration defaults to Inf, the sort is a no-op and the
+      # order falls back to random -- leaving the longest test to start last and run alone.
+      # A few KB, keyed to ParallelTestRunner's UUID; the blanket scratchspace cache stays off.
+      - uses: actions/cache@v4
```
giordano:
```suggestion
      - uses: actions/cache@v6
```

simone-silvestri:
Ups, by mistake I thought it was accepted but actually #577 was the accepted one
@giordano @glwagner let me know if you can approve reotroactively otherwise I can revert this PR

PR #444, opened 2026-07-15
No-flux BC for penetrative radiation

simone-silvestri:
Penetrative radiation, on main, is lost upon touching the bottom. This PR implements a no-flux boundary condition for penetrative radiation whereas all the remaining energy is dumped onto the last cell above the immersed boundary.
This was also a bug in the tracer budget tests for a mutable z-star where, changing the surface level in time we were losing different amounts of radiation (bottom is at 100m). 
With the new implementation we can significantly reduce the tolerances for the tracer budget tests.


src/Oceans/radiative_forcing.jl
```diff
@@ -46,11 +47,13 @@ const c = Center()
...
 
 # In zstar we might have positive z, so  `exp(κ * z)` is not correct
+# Radiation that reaches the bottom is dumped on the last cell
 @inline function beers_law_radiation(i, j, k, grid, J₀ , κ)
     Nz = size(grid, 3)
     z  = Oceananigans.Grids.znode(i, j, k,    grid, c, c, f)
     η  = Oceananigans.Grids.znode(i, j, Nz+1, grid, c, c, f)
-    return J₀ * exp(κ * (z - η))
+    J  = J₀ * exp(κ * (z - η))
+    return ifelse(inactive_cell(i, j, k - 1, grid), zero(J), J)
```
taimoorsohail, line 56: why k-1? Because we want to skip the surface right? what about the case where k = 1?

taimoorsohail (approved):
Left a comment, otherwise LGTM.

src/Oceans/radiative_forcing.jl
simone-silvestri: k-1 is here because we are looking at interfaces while `inactive_cell` checks centers. Basically this says, if the cell below you (k-1) is immersed, zero out the flux at this interface (which is equivalent of saying dump all the energy in this cell)

PR #445, opened 2026-07-15
Correct timestamp for JRA55 rivers and icebergs

simone-silvestri:
The multi-year JRA55 rivers and icebergs start at 12pm not at midnight. The data filling in set! was skipped without warning because the dates were not matching the native dates in the file.
This PR fixes it and introduces an assert ensuring all the requested dates are filled


test/test_jra55.jl
```diff
@@ -217,5 +217,22 @@ using NumericalEarth.JRA55: download_JRA55_cache
                 end
             end
         end
+
+        @info "Testing MultiYearJRA55 daily variables on $A..."
```
taimoorsohail: It looks to me like this test (1) checks the output time (12PM) (2) Tests downloading and (3) ensures the value is nonzero. 

I think (1) is covered by https://github.com/NumericalEarth/NumericalEarth.jl/pull/445/changes#diff-5796de728734089a0853fac52381f3490f2ad246e80121558533b6cbd1b249b4R114

I think (2) should be put into the downloading tests (right?)

simone-silvestri: I am actually planning to remove the MultiYearJRA55 :temperature tests in favor of the rivers because they are much shorter to download, this is how I found out this bug. Maybe I can do it now directly.

taimoorsohail: Ah cool!

simone-silvestri:
@taimoorsohail this should be ready now

taimoorsohail (approved):

PR #454, opened 2026-07-18
Land-coupled nested atmospheres: SlabLand + RRTMGP for the ERA5 downscaling example

glwagner:
## Summary

Makes a `NestedModel` a first-class `EarthSystemModel` atmosphere component, so a parent-driven LAM couples to `SlabLand` and RRTMGP radiation through the standard machinery:

```julia
nest       = nested_atmosphere_model(grid, dataset; dates, ...)
radiation  = RadiativeTransferModel(grid, AllSkyOptics(), constants;
                                    solar_position = ApparentSolarPosition(epoch = start_date),
                                    surface_albedo = CopernicusAlbedo(),
                                    schedule = TimeInterval(1hour))
model      = AtmosphereLandModel(Simulation(nest; Δt), land; radiation)
```

## What's here

- **`NestedModel` as an ESM component** (`src/NestedModels/nested_component.jl`): the component interface forwards to the child, for both the bare nest and its `Simulation` wrapper; the CFL wizard reaches the child.
- **Child rain reaches the surface**: the Breeze exchanger carries `surface_precipitation_flux` as its `Jʳⁿ` state, feeding the land bucket and ocean freshwater budget (zero-flux fallback for non-precipitating schemes; snow TODO).
- **Double-counting guard**: `net_fluxes` throws if the atmosphere's momentum bottoms are `BulkDrag` rather than coupling flux fields.
- **RRTMGP for nests**: `materialize_earth_system_radiation!` rebuilds child/nest/simulation around the materialized proxy; the ESM constructor binds the atmosphere–land *interface* skin temperature (not `land.temperature` — they differ for non-bulk formulations) into an RTM constructed without one. A Breeze RTM is now given no `ComponentExchanger`, so its RRTMGP solver internals never enter the atmosphere–land flux kernel (that path was GPU-fatal, silently harmless on CPU).
- **`Float32` coupled clock**: the `EarthSystemModel`'s default clock now adopts the atmosphere's own time type, so a `Float32` nest (as in the example) no longer needs an explicitly matching clock.
- **Dataset surface albedo** (`src/DataWrangling/CopernicusLandAlbedo/`): `surface_albedo = CopernicusAlbedo()` materializes the CGLS 1 km blue-sky albedo at the dekad nearest the solar epoch, with water pixels set to open water. It downloads through `CopernicusClimateDataStore` — the single CDS backend the example now uses for both ERA5 and the albedo (the CDSAPI albedo path is retired).
- **Example**: `breeze_downscaling_era5.jl` is on the MC3E squall-line domain (all land) with the full coupling, plus a land-response figure; it also carries the earlier terrain-smoothing work and the origin/main (0.6.0) merge. It now runs from a **dedicated `examples/Project.toml`** (rather than borrowing the docs environment) and type-erases the section-animation observables so per-frame type drift can't break the record loop.

## Data wrangling changes (moves, not reverts)

Review flagged the ERA5/CDS diffs as possible merge-resolution accidents — they are deliberate, and no functionality from `main` was lost (each removed symbol was checked against its new home):

- **Backend-agnostic batching/splitting moved to `src`**: the calendar-month batching, CDS cost-cap logic, and multi-step NetCDF splitting that lived in `NumericalEarthCDSAPIExt` (`cds_utils.jl`, `era5.jl`) now live in `src/DataWrangling/ERA5/ERA5_batched_downloads.jl`, shared by both CDS backends (CDSAPI.jl and era5cli via CopernicusClimateDataStore.jl). `split_nc_multistep` was renamed `split_era5_nc_by_datetime` in the move.
- **Albedo request/extraction/repacking moved to `src/DataWrangling/CopernicusLandAlbedo/`**: backends supply only `retrieve`.
- **ERA5 single levels**: valid dates extended through 2025 and IVT variables added — needed to drive nests from recent dates.

## Upstream dependencies

All Breeze prerequisites have landed: [#855](https://github.com/NumericalEarth/Breeze.jl/pull/855), [#856](https://github.com/NumericalEarth/Breeze.jl/pull/856), [#857](https://github.com/NumericalEarth/Breeze.jl/pull/857), [#901](https://github.com/NumericalEarth/Breeze.jl/pull/901) (field-valued surface emissivity/albedo), and [#791](https://github.com/NumericalEarth/Breeze.jl/pull/791) ("RRTMGP on terrain-following compressible grids", merged Aug 11) are all in **Breeze v0.9.0, registered 2026-08-19**. The git-revision pins are retired: the `docs`, `examples`, and `test` environments are bound to registered Breeze 0.9 (so unit-test CI exercises the full RTM paths, and the resolver can't assemble a valid-but-broken 0.8 environment), and the root compat adds 0.9 beside its existing bounds since Breeze is a weakdep there.

## Testing

`test_breeze_coupling.jl` gains nested construction / flux-routing / clock-sync tests and a Kessler rain-routing test (CPU-only, per Breeze.jl#858). A 1-hour A100 smoke of the full stack (terrain-following RRTMGP, observed CGLS albedo, child rain into the bucket, all writers, animations, land-response figure) ran green end-to-end ([details](https://github.com/NumericalEarth/NumericalEarth.jl/pull/454#issuecomment-5012436787)).

A later GPU re-run went nonfinite (~t=26 min, surface thermodynamics); this traced to a stale local `examples` manifest (gitignored) that had pinned an interim Breeze commit (`0a0d8d69`) predating Breeze's own immediate fix (`5d3f3425`) — not to any NumericalEarth change. The environments now resolve to registered Breeze 0.9.0, which contains that fix.


glwagner:
1-hour GPU smoke of the example is green end-to-end on an A100: nest from cached ERA5, SlabLand + interface-bound RRTMGP (terrain-following, observed CGLS albedo, hourly solves), child rain into the bucket, all writers, both animations, land-response figure. The simulated hour cost 7.2 minutes wall — the full 18-hour hindcast projects to ~2.2 hours. Peak updrafts ~34 m/s with qʳ up to ~6 g/kg as the squall line develops.

ewquon:
Since you guys are touching the breeze-atmosphere interface, I wanted to make you aware of #472. I think this is a good time to quickly address it.

kaiyuan-cheng:
@glwagner @ewquon We may want to narrow the scope of this PR to land-coupling, as the radiation part still needs some work on Breeze's end.

glwagner:
> @glwagner @ewquon We may want to narrow the scope of this PR to land-coupling, as the radiation part still needs some work on Breeze's end.

Is there any harm in what's been done so far?

kaiyuan-cheng:
> > @glwagner @ewquon We may want to narrow the scope of this PR to land-coupling, as the radiation part still needs some work on Breeze's end.
> 
> Is there any harm in what's been done so far?

The RRTMGP coupling does not seem to work with the main branch of Breeze.

glwagner:
> > > @glwagner @ewquon We may want to narrow the scope of this PR to land-coupling, as the radiation part still needs some work on Breeze's end.
> > 
> > 
> > Is there any harm in what's been done so far?
> 
> The RRTMGP coupling does not seem to work with the main branch of Breeze.

So we should expect this case to blow up, is that right?

kaiyuan-cheng:
> > > > @glwagner @ewquon We may want to narrow the scope of this PR to land-coupling, as the radiation part still needs some work on Breeze's end.
> > > 
> > > 
> > > Is there any harm in what's been done so far?
> > 
> > 
> > The RRTMGP coupling does not seem to work with the main branch of Breeze.
> 
> So we should expect this case to blow up, is that right?

Yes, this case will crash because Breeze's RRTMGP path feeds the radiation scheme the wrong atmospheric state.

glwagner:
Re the RRTMGP-on-Breeze-main question above ([comment](https://github.com/NumericalEarth/NumericalEarth.jl/pull/454#issuecomment-5246704658)): the 72 h December-2025 AR hindcast ran all-sky RRTMGP on Breeze main (compressible, terrain-following) without crashing — on main the wrong-state feed shows up as a quiet heating-rate bias (cloud state built on the static reference density, `all_sky_radiative_transfer_model.jl:318`), not a blow-up. The fix is NumericalEarth/Breeze.jl#791, which I've just updated with a merge of current Breeze main. Once that merges, the radiation scope of this PR shouldn't need narrowing — results produced meanwhile carry the bias and deserve a re-run.

docs/Project.toml
```diff
@@ -26,12 +27,14 @@ Oceananigans = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
...
 RRTMGP = "a01a1ee8-cea4-48fc-987c-fc7878d79da1"
 Reactant = "3c362404-f566-11ee-1572-e11a4b42c853"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
 SeawaterPolynomials = "d496a93d-167e-4197-9f49-d3af4ff8fe40"
 SpeedyWeather = "9e226e20-d153-4fed-8a5b-493def4f21a9"
 Suppressor = "fd094767-a336-5f1f-9728-57cf17d0bbfb"
 
 [sources]
 NumericalEarth = {path = ".."}
+Breeze = {url = "https://github.com/NumericalEarth/Breeze.jl.git", rev = "main"}
```
xkykai: should we use released versions to prevent breakage in the future?

examples/breeze_downscaling_era5.jl
```diff
@@ -1,8 +1,10 @@
-# # ERA5 → 12 km convection-permitting hindcast (Breeze + NestedModel)
+# # ERA5 → 12 km land-coupled hindcast (Breeze + NestedModel + SlabLand)
```
xkykai:
```suggestion
# # ERA5 → 12 km land-coupled, convection-permitting hindcast (Breeze + NestedModel + SlabLand)
```

examples/breeze_downscaling_era5.jl
```diff
@@ -109,22 +124,25 @@ grid = LatitudeLongitudeGrid(arch;
...
-                                dir = era5_datadir,
-                                terrain = ETOPO2022(),
-                                terrain_blend_width = relax_width,
-                                relaxation_rate = 1/300,
-                                relaxation_width = relax_width,
-                                momentum_advection = WENO(order = 5))
+nest = nested_atmosphere_model(grid, dataset;
+                               dates,
+                               dir = era5_datadir,
+                               terrain = ETOPO2022(),
```
xkykai, line 134: doesn't have to address this but I wonder if `GLO30()` would work, would be nice if it does

examples/Project.toml
```diff
@@ -0,0 +1,24 @@
...
+GeoInterface = "cf35fbd7-0cd7-5166-be24-54bfbe79505f"
+NaturalEarth = "436b0209-26ab-4e65-94a9-6526d86fea76"
+NumericalEarth = "904d977b-046a-4731-8b86-9235c0d1ef02"
+Oceananigans = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
+Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
+RRTMGP = "a01a1ee8-cea4-48fc-987c-fc7878d79da1"
+
+[sources]
+NumericalEarth = {path = ".."}
+Breeze = {url = "https://github.com/NumericalEarth/Breeze.jl.git", rev = "main"}
```
xkykai: same here

ext/NumericalEarthBreezeExt/breeze_air_land_radiation.jl
```diff
@@ -1,23 +1,54 @@
...
+##### The RTM's downwelling longwave is atmospheric transfer and remains fixed between
+##### scheduled solves. Its upwelling longwave also contains the surface boundary condition
+##### from the last solve, so it cannot be used directly while the surface temperature evolves.
+##### Reconstruct the boundary every coupled step from the live surface temperature and
+##### emissivity: `ℐˡʷꜛ = ε σ Tₛ⁴ - (1-ε) ℐˡʷꜜ`, where downwelling flux is negative.
 #####
-##### This runs in `update_state!` after the turbulent (sensible + latent)
-##### flux has been written to `surface_energy_flux`, so the kernel adds the
-##### radiative term on top.
+##### Shortwave is NOT: RRTMGP reflects the surface albedo internally, but Breeze stores
```
xkykai: NOT?

ext/NumericalEarthBreezeExt/breeze_atmosphere_interface.jl
```diff
@@ -72,14 +74,24 @@ NumericalEarth.EarthSystemModels.boundary_layer_height(atmos::BreezeAtmosphereSi
...
                                                                                    correction = nothing)
+    # Breeze's surface rain-flux diagnostic (positive down, kg m⁻² s⁻¹): built once, computed
+    # each `interpolate_state!`. Schemes with no precipitating species define no method —
+    # fall back to an inert zero field.
+    # TODO: move the fallback into Breeze; add a snow analog (Jˢⁿ stays zero below).
+    Jʳⁿ = if applicable(surface_precipitation_flux, atmosphere, atmosphere.microphysics)
+        surface_precipitation_flux(atmosphere)
+    else
+        Oceananigans.CenterField(exchange_grid)
+    end
```
xkykai, line 84: is this how we usually write if/else statements?

ext/NumericalEarthBreezeExt/breeze_nested_atmosphere.jl
```diff
@@ -275,7 +343,21 @@ function mean_surface_pressure(dataset, child_grid, date, dir)
     p₀ = Field{Center, Center, Nothing}(child_grid)
     set!(p₀, Metadatum(:surface_pressure; dataset = single_level_dataset, date,
                        region = BoundingBox(child_grid), dir))
-    return sum(interior(p₀)) / length(interior(p₀))
+    # Reduce across ranks so every rank anchors the same hydrostatic reference
+    # (`all_reduce` is the identity on serial architectures).
+    arch = architecture(child_grid)
+    return all_reduce(+, sum(interior(p₀)), arch) / all_reduce(+, length(interior(p₀)), arch)
```
xkykai, line 350: should we make a `sum` operation in Oceananigans and make it `all_reduce`? same for `Integral` and related functions as well

ext/NumericalEarthBreezeExt/breeze_surface_albedo.jl
```diff
@@ -0,0 +1,28 @@
...
+# `surface_albedo = CopernicusAlbedo()` in a RadiativeTransferModel: CGLS blue-sky albedo
+# on the model grid at the dekad nearest the solar epoch; water/missing pixels (NaN in the
+# land product) become open-water albedo.
+
+albedo_epoch(solar_position::Breeze.ApparentSolarPosition) = solar_position.epoch
+albedo_epoch(solar_position) = nothing
+
+# The hook lands with NumericalEarth/Breeze.jl#857. TODO: unconditionalize after release.
+if isdefined(Breeze.AtmosphereModels, :materialize_surface_property)
+    @eval function Breeze.AtmosphereModels.materialize_surface_property(dataset::CopernicusAlbedo, grid, solar_position)
```
xkykai, line 14: nice seeing my PR put into good use!

ext/NumericalEarthCDSAPIExt/cds_utils.jl
xkykai, line 1: I am a bit confused by this and all of the data ingestion edits in this PR in general? Are these intentional? Why are they removed/edited? Is it something bad, or is it an error when resolving merge conflicts? Because a lot of these seemed to be the previous version when compared against `main`

ext/NumericalEarthCDSAPIExt/era5.jl
xkykai, line 1: same as above, a bit confused by what's going on here. why is there ERA5 data wrangling edits in this PR?

src/Bathymetry/smooth_topography.jl
xkykai, line 1: would be nice to show a before/after topography smoothing as a figure perhaps, even just a draft figure will do. right now I'm not exactly sure what it does and if it looks correct

src/DataWrangling/CopernicusLandAlbedo/CopernicusLandAlbedo.jl
xkykai, line 1: I see now, the non CDS-related methods are being moved here. fine for now, perhaps indicate that in the PR description cause this might seem like silent scope creep. also some of these are my fault so sorry about that

src/DataWrangling/ERA5/ERA5_batched_downloads.jl
xkykai, line 1: potential scope creep?

src/DataWrangling/ERA5/ERA5_single_levels.jl
xkykai, line 1: seems like ERA5-related extensions are extensive enough to call for an updated PR description/a separate PR

src/Radiations/air_sea_interface_radiation_state.jl
```diff
@@ -21,19 +24,21 @@ end
     return (; σ, α, ϵ, ℐꜜˢʷ, ℐꜜˡʷ)
 end
 
+@inline _surface_radiation_state(::Nothing, rk, exchanger_state, i, j, k, grid, time) =
+    _zero_radiation_state(grid)
+
 @inline EarthSystemModels.InterfaceComputations.air_sea_interface_radiation_state(
         rk, exchanger_state, i, j, k, grid, time) =
-    _surface_radiation_state(rk.surface_properties.ocean,
+    _surface_radiation_state(get(rk.surface_properties, :ocean, nothing),
```
xkykai, line 32: is this a new way of calling? if it is better should we do something to standardize?

xkykai (commented):
also consider making some of the code comments/docstrings more concise. I think I will reread this again another time for another review pass since it is bulky!

docs/Project.toml
glwagner: yeah for sure, we can't merge with this

ext/NumericalEarthBreezeExt/breeze_air_land_radiation.jl
glwagner: yeah what the hey does this mean

ext/NumericalEarthBreezeExt/breeze_atmosphere_interface.jl
glwagner: this is how you write it if you want the assigned variable (`Jʳⁿ` in this case) to be in the outer scope

ext/NumericalEarthBreezeExt/breeze_nested_atmosphere.jl
glwagner: i actually think we do have global reductions already. so the fix here may be simply to use that. i will look into it

ext/NumericalEarthCDSAPIExt/cds_utils.jl
glwagner: should we revert data ingestion edits? indeed, they may be stale

docs/Project.toml
kaiyuan-cheng: We're using the main branch for now as we continue merging prerequisite PRs into Breeze to get everything working with this PR.

ext/NumericalEarthBreezeExt/breeze_nested_atmosphere.jl
```diff
@@ -298,18 +299,25 @@ function NumericalEarth.NestedModels.nested_atmosphere_model(parent_atmosphere::
...
+    # calls `mask(x, y, z)`, so wrap a scalar mask in a callable. Momentum and energy relax toward the
+    # parent's SPECIFIC state, which `SpecificForcing` weights by the child's own `ρᵈ` at kernel time;
+    # relaxing toward the parent's `ρθ`/`ρu`/`ρv` instead equilibrates at `θ = θₚ ρᵈₚ / ρᵈ`, an absolute
+    # error `θ Δρᵈ / ρᵈ` (≈3 K per 1% density mismatch: the lateral-boundary cold rim). `ρᵈ` itself is
+    # absent because Breeze's compressible continuity kernels overwrite `Gⁿ.ρᵈ` with `-∇·m` and never
+    # read `forcing.ρᵈ`, so a mass-nudging entry is silently discarded; were that to change, the
+    # specific form would gain a `θ Δρᵈ / ρᵈ` cross-term and the density-weighted form would be unbiased.
+    # The wrap is explicit and keyed by the density-weighted prognostic rather than left to Breeze's
+    # specific-key dispatch, so a caller's own `θ`/`u`/`v` forcing combines with the relaxation instead
+    # of replacing it in the `merge` below.
```
glwagner, line 303: isn't this addressed by @ewquon 's PR? could be duplicating work, not sure. we also discussed the importance of avoiding SpecificForcing and Relaxation.

kaiyuan-cheng: This is indeed @ewquon 's PR that has been merged recently. There is no duplication.

src/Radiations/air_sea_interface_radiation_state.jl
glwagner: no, I think this is actually a bad idea. but `get` is a normal way to extract properties from `Dict`.

src/DataWrangling/ERA5/ERA5_single_levels.jl
glwagner: ill look at that

ext/NumericalEarthBreezeExt/breeze_nested_atmosphere.jl
glwagner: oh got it.

docs/Project.toml
glwagner: Breeze#791 merged and Breeze 0.9.0 — which contains it and #855–#857 — was registered today, so the branch pins are retired in db96e792 (thanks @kaiyuan-cheng): the docs, examples, and test environments require registered Breeze 0.9 (so unit CI exercises the full RTM paths) and the root compat adds it.

examples/Project.toml
glwagner: Retired along with the docs pin in db96e792 — the examples environment now requires registered Breeze 0.9.

examples/breeze_downscaling_era5.jl
glwagner: It does work today through the region-windowed form (GLO30 is tile-based, so it refuses a global regrid): `metadatum = Metadatum(:bottom_height; dataset = GLO30(), region = BoundingBox(grid))`, then `regrid_topography(grid, metadatum)` — and `nested_atmosphere_model` accepts the resulting elevation `Field` via `terrain`. Making `terrain = GLO30()` auto-window to the child grid's bounding box would be a nice follow-up.

ext/NumericalEarthBreezeExt/breeze_air_land_radiation.jl
glwagner: That sentence was mangled — it meant "shortwave is *not* reconstructed the way longwave is". The header is rewritten (and cut from 30 lines to 7) in 181accc6.

ext/NumericalEarthBreezeExt/breeze_nested_atmosphere.jl
glwagner: Oceananigans does ship the primitive — `Oceananigans.DistributedComputations.all_reduce` — which is what this call site composes with a rank-local `sum`. Making `sum`/`mean`/`Integral` on distributed `Field`s globally-reducing by default belongs upstream; 181accc6 leaves a TODO at the call site to that effect.

ext/NumericalEarthCDSAPIExt/cds_utils.jl
glwagner: Checked symbol by symbol: nothing here reverts main — these are moves. The backend-agnostic batching/splitting machinery (`batch_datetimes_for_cds`, `group_by_calendar_month`, `foreach_nc`, `is_zip`, `ncvar_copy_tslice!`, `split_nc_multistep` → renamed `split_era5_nc_by_datetime`, …) moved to `src/DataWrangling/ERA5/ERA5_batched_downloads.jl`, shared by the CDSAPI and era5cli backends. The PR description now has a "Data wrangling changes (moves, not reverts)" section covering all of it.

ext/NumericalEarthCDSAPIExt/era5.jl
glwagner: Same story as `cds_utils.jl` — the deleted functions moved to `src/DataWrangling/ERA5/ERA5_batched_downloads.jl` so both CDS backends share them; see the new "Data wrangling changes" section in the description.

src/DataWrangling/CopernicusLandAlbedo/CopernicusLandAlbedo.jl
glwagner: Done — the description now has a "Data wrangling changes (moves, not reverts)" section spelling out the ext→src moves.

src/DataWrangling/ERA5/ERA5_batched_downloads.jl
glwagner: This file is the destination of the ext→src move rather than new scope — it is the shared home for the batching/splitting machinery both CDS backends use. Covered in the description's new "Data wrangling changes" section.

src/DataWrangling/ERA5/ERA5_single_levels.jl
glwagner: Looked into it: the single-levels edit itself is small — the practical date range extends through 2025 and three integrated-vapor variables (`tcwv`, `viwve`, `viwvn`) are added, needed to drive nests from recent dates. The bulk of the ERA5 diff is the machinery move described in the new "Data wrangling changes" section of the description, so I'd keep it here rather than split it.

src/Bathymetry/smooth_topography.jl
glwagner: Here's before/after on two configurations: the example's own (MC3E domain, 12 km, the default 2 passes) and steep terrain where the filter actually matters (Cascades at 3 km, 4 passes — the settings from the atmospheric-river runs):

smooth_topography before/after

What to look for:

- **MC3E, 2 passes**: the difference is confined to the Front Range at the western edge (max |Δ| ≈ 845 m on the steepest peaks); the Great Plains — where the squall line lives — are essentially untouched.
- **Cascades, 4 passes**: the difference map is exactly the grid-scale ridge/valley ripple the filter exists to remove; the large-scale ranges survive.
- **Transect (~Mt Rainier)**: broad terrain shape is preserved while two-grid-length oscillations vanish; the isolated volcano summit is damped (near-grid-scale by construction — each pass multiplies a mode's amplitude by `(1+cos kΔx)(1+cos kΔy)/4`, annihilating the 2Δx mode). Bounded edges replicate the nearest interior value, so the terrain isn't pulled toward zero at the boundary.

The motivation: point-sampled regridding leaves grid-scale orographic roughness that excites standing 2Δx noise in the near-surface flow on the terrain-following coordinate — on a 3 km Cascadia run this filter cut near-surface σ(U) noise roughly in half.

glwagner:
Review pass addressed — summary of what changed:

- **Breeze pins retired** (the merge blocker): Breeze#791 merged and **Breeze 0.9.0 was registered today** with everything this PR needs (#791, #855–#857), so `docs`/`examples` no longer pin a git revision and root/test compat accept 0.9 (db96e792, thanks @kaiyuan-cheng). Unit CI now exercises the full RTM paths against a registered release.
- **Comments tightened** (181accc6): the long narrative comment blocks flagged in review are cut to 1–4 lines across the Breeze coupling and data wrangling (the `breeze_air_land_radiation.jl` header went from 30 lines to 7, and its garbled "Shortwave is NOT:" sentence is fixed). Net −70 comment lines.
- **PR description updated** with a "Data wrangling changes (moves, not reverts)" section: the ERA5/CDS diffs that looked like stale merge artifacts are deliberate moves of backend-agnostic batching/splitting machinery into `src/DataWrangling/ERA5/ERA5_batched_downloads.jl` (shared by the CDSAPI and era5cli backends) — verified symbol by symbol that nothing from `main` was lost.
- **Topography smoothing figure** posted on the `smooth_topography.jl` thread, before/after on the example's own MC3E 12 km configuration and on 3 km Cascades terrain.
- **Not in scope here**: #472 (coupled momentum stress applied at centers while `ρu`/`ρv` live at faces) predates this PR and affects the ocean coupling too — keeping it as its own focused fix rather than folding it into an already-bulky PR.

glwagner:
Heads-up: the Reactant-extension CI job started failing on a fresh resolve — Reactant 0.2.280 can't lower `enzymexla.math.fmuladd` from CUDACore's `muladd` device override (reached through Oceananigans' WENO interpolants). It's an upstream regression, already fixed but not yet in a registered release (EnzymeAD/Reactant.jl#3196: 0.2.279 passes, 0.2.280 fails, same CUDACore 6.2.1 / Oceananigans 0.110.16 in both runs). 29099c81 caps the test environment at `0.2.278 - 0.2.279` until the fix registers.

Note **`main` is red with the same Reactant failure** (its latest CI run, Aug 18) — the same one-line cap there would green it up for every open PR.

kaiyuan-cheng (approved):

PR #462, opened 2026-07-21
Fix AVISO downloads for daily and monthly variables

taimoorsohail:
## Summary

- use the monthly AVISO variable table when building monthly filenames
- test every variable that each daily and monthly AVISO dataset says is available
- test monthly sea-level anomaly loading with FieldTimeSeries

## Why

The monthly AVISO product only contains sea-level anomaly. The download test was asking it for free surface and velocity fields that are only available in the daily product, which caused a KeyError.

The field-loading test also used DatasetRestoring. Sea-level anomaly is data that can be loaded into a field, but it is not one of the model fields supported by DatasetRestoring. The test now checks FieldTimeSeries directly instead.

## Impact

AVISO download tests now follow the variables provided by each product. Unsupported monthly variables are not hidden by a fallback to the daily variable table.

## Validation

- git diff --check
- local tests were not run, following the repository testing instructions; GitHub CI will run them

### DIFF
--- a/test/test_aviso_downloading.jl
+++ b/test/test_aviso_downloading.jl
@@ -3,7 +3,7 @@ include("download_utils.jl")
 
 using CopernicusMarine
 using NumericalEarth
-using NumericalEarth.DataWrangling: BoundingBox, metadata_path
+using NumericalEarth.DataWrangling: BoundingBox, available_variables, metadata_path
 using Oceananigans.Fields: interior
 using Oceananigans.Grids: Bounded, Flat, LatitudeLongitudeGrid, topology
 using Oceananigans.OutputReaders: time_indices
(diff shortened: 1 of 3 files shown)


test/test_aviso_downloading.jl
```diff
@@ -28,25 +28,32 @@ using Oceananigans.OutputReaders: time_indices
...
 @testset "Downloading AVISO data" begin
-    variables = (:free_surface, :sea_level_anomaly, :zonal_geostrophic_velocity, :meridional_geostrophic_velocity)
     region = BoundingBox(longitude=(200, 202), latitude=(35, 37))
-    dataset = AVISOMonthly()
     date = DateTime(2020, 1, 1)
+    datasets_and_variables = (
+        (AVISODaily(), (:free_surface,
+                        :sea_level_anomaly,
+                        :zonal_geostrophic_velocity,
+                        :meridional_geostrophic_velocity)),
```
simone-silvestri:
```suggestion
    datasets_and_variables = (
        (AVISODaily(), (:sea_level_anomaly),
```
I guess just one product is enough? (to reduce burden on the testing infrastructure)

simone-silvestri (approved):

PR #401, opened 2026-07-01
Slab canopy: big-leaf conductance, canopy air space, interception, and tiled land

xkykai:
## Overview

This PR adds a **vegetation surface** for `SlabLand` — from a single-source ("big-leaf") canopy up to a **two-source canopy air space** (optionally with the canopy air's own heat and moisture storage, and with a prognostic ground skin), plus a **prognostic canopy interception store** and a **surface water store** for rain the soil cannot take — exposed as composable interface formulations that plug into the existing Monin–Obukhov flux solver. It builds on #323 (`glw/systematic-land-model`, merged), which supplied the variably-saturated hydrology, the water-coupled energy budget, and the `DryLayerHumidity` bare-soil evaporation model this work builds on.

The canopy carries its own physiology and coupling physics rather than borrowing the bare-soil ones: transpiration stress is keyed to plant-available water between wilting point and field capacity (§12), and the ground↔canopy-air exchange can respond to canopy density and wind (§13).

The design target is **atmosphere–land coupled simulation at weather / tropical-cyclone scales** (days-to-weeks, ~100 m–1 km), not a seasonal carbon-cycle land model. At that scale a cloud-resolving atmosphere consumes four things from the land: the surface humidity (→ latent heat, moisture supply, cloud base), the radiometric skin temperature (→ sensible heat, upwelling longwave, boundary-layer growth), roughness, and albedo. Everything here is aimed at getting the **turbulent-flux lower boundary condition** — above all the **Bowen ratio** (the sensible/latent partition) — right over a storm-length window, reusing the ground `SlabLand` for soil water and energy.

Defaults and equations follow ClimaLand (Deck et al. 2026, *JAMES*, App. C–E for the turbulent/physiology closures, App. D for the canopy radiative transfer) unless noted, checked against the primary sources and the CTSM / Noah-MP / ClimaLand source code; where this implementation departs from them the docstring says so. Every new path is allocation-free and type-stable on CPU; GPU and Enzyme/Reactant coverage is limited to what the test suite exercises.

The coupling contract (§22): land fluxes are computed once per coupled step with the step's `Δt`, the slab is driven by the skin→bulk conduction wherever a skin closes its own energy balance, and the land water store is debited by the source fluxes (soil evaporation + transpiration), the canopy store by the wet-canopy evaporation.

[...]

PR #464, opened 2026-07-21
Add ESA WorldCover land-cover dataset (fractional cover)

xkykai:
Adds **ESA WorldCover** as a `DataWrangling` dataset, turning the 10 m global land-cover raster into the continuous fields a fractional-cover land surface consumes: a vegetation fraction `f_veg`, a per-class area fraction for each of the 11 legend classes, and a majority land-cover class.

Class codes are categorical, so they are aggregated by counting and never averaged. A windowed anonymous-S3 COG read at 10 m is reduced by an integer factor onto a ~110 m native grid; onto a model grid the fractions are regridded conservatively and the class is the argmax of those fractions.

## Variables

- `:vegetation_fraction` — area fraction of vegetated classes (tree, shrub, grass, crop, herbaceous wetland, mangrove); the mosaic weight `f_veg`, a drop-in for a `TiledLandInterface`'s data-free default cover fraction.
- `:<class>_fraction` — one per legend class (`:cropland_fraction`, `:built_up_fraction`, …), each in `[0, 1]`.
- `:landcover_class` — the majority class code, always an exact legend code.

`ESAWorldCover(; version = :v200, aggregation_factor = 12)`. `aggregation_factor` sets the native resolution (`12` → ~110 m, `120` → ~1 km) and is encoded in the cache filename.

## Regridding

`interpolate_physical!` gains a metadatum-dispatched method so a dataset can choose how its variables land on a model grid; the default stays bilinear. WorldCover regrids the fractions with Oceananigans' conservative, area-weighted `regrid!` (needs 0.110.13 for its multi-dimensional support, hence the compat bump) and takes `:landcover_class` as their argmax, the class covering the most area of each target cell. Bilinear interpolation of class codes would produce intermediate values outside the legend. Cells whose fractions sum to ~0, outside the product's coverage, are masked.

## Data access

- Anonymous public bucket `s3://esa-worldcover/`, no credentials; requires `using ArchGDAL`.
- Regional windows only — the metadatum must carry a `BoundingBox`.
- `version = :v200` (2021, default) or `:v100` (2020); ESA warns the two are not comparable.
- CC-BY 4.0, DOI v200 `10.5281/zenodo.7254221` (Zanaga et al. 2022).

## Usage

Veluwe forest and the Randmeren lakes, central Netherlands:

```julia
using ArchGDAL   # activates the anonymous-COG read path

region  = BoundingBox(longitude = (5.45, 5.95), latitude = (52.05, 52.45))
dataset = ESAWorldCover()
grid = LatitudeLongitudeGrid(CPU(); size = (150, 120), longitude = region.longitude,
                             latitude = region.latitude, topology = (Bounded, Bounded, Flat))

f_veg    = Field(Metadatum(:vegetation_fraction; dataset, region), grid)   # mosaic weight
cropland = Field(Metadatum(:cropland_fraction;   dataset, region), grid)   # area fraction
class    = Field(Metadatum(:landcover_class;     dataset, region), grid)   # legend codes
```

Pass `CPU()` instead of `grid` to get a field on the dataset's own ~110 m grid.
`examples/esa_worldcover_fractions.jl` produces the figures below.

## Testing

91 unit tests (no network): legend, mode aggregation, per-class fractions summing to 1 over valid pixels, no-data masking, integer-factor alignment, dataset interface, bounded-region requirement, window/native-grid alignment. `test/test_esaworldcover_downloading.jl` covers the live read and is excluded from CI.

Verified against the live bucket over the window above: model-grid class codes stay within the legend, the eleven fractions sum to `1 ± 2e-8`, and `f_veg` matches the summed vegetated fractions to 2e-8.


xkykai:
## Demonstration

```julia
using NumericalEarth, Oceananigans, ArchGDAL, CairoMakie, Statistics

region  = BoundingBox(longitude = (5.45, 5.95), latitude = (52.05, 52.45))
dataset = ESAWorldCover()
grid = LatitudeLongitudeGrid(CPU(); size = (150, 120), longitude = region.longitude,
                             latitude = region.latitude, topology = (Bounded, Bounded, Flat))

legend = (tree_cover = (10, "#006400"), shrubland = (20, "#ffbb22"),
          grassland = (30, "#ffff4c"), cropland = (40, "#f096ff"),
          built_up = (50, "#fa0000"), bare_sparse_vegetation = (60, "#b4b4b4"),
          snow_and_ice = (70, "#f0f0f0"), permanent_water_bodies = (80, "#0064c8"),
          herbaceous_wetland = (90, "#0096a0"), mangroves = (95, "#00cf75"),
          moss_and_lichen = (100, "#fae6a0"))
class_names  = keys(legend)
class_codes  = map(first, values(legend))
class_colors = map(last, values(legend))

fveg_native  = Field(Metadatum(:vegetation_fraction; dataset, region), CPU())
fveg_model   = Field(Metadatum(:vegetation_fraction; dataset, region), grid)
class_native = Field(Metadatum(:landcover_class; dataset, region), CPU())
class_model  = Field(Metadatum(:landcover_class; dataset, region), grid)
fractions    = NamedTuple(name => Field(Metadatum(Symbol(name, :_fraction); dataset, region), grid)
                          for name in class_names)

array(field) = Array(interior(field))[:, :, 1]
λ_native, φ_native = λnodes(class_native.grid, Center()), φnodes(class_native.grid, Center())
λ_model,  φ_model  = λnodes(grid, Center()), φnodes(grid, Center())

to_index(field) = map(array(field)) do code
    i = isnan(code) ? nothing : findfirst(==(round(Int, code)), class_codes)
    isnothing(i) ? NaN : Float64(i)
end
index_native, index_model = to_index(class_native), to_index(class_model)

fig = Figure(size = (1180, 640), fontsize = 15)
for (col, (name, λ, φ, index)) in enumerate((("native ~110 m", λ_native, φ_native, index_native),
                                             ("model grid (area majority)", λ_model, φ_model, index_model)))
    ax = Axis(fig[1, col]; title = "Majority class ($name)", xlabel = "longitude", ylabel = "latitude")
    heatmap!(ax, λ, φ, index; colormap = cgrad(collect(class_colors), categorical = true),
             colorrange = (0.5, length(class_codes) + 0.5))
end
present = sort(unique(filter(!isnan, vcat(vec(index_native), vec(index_model)))))
Legend(fig[1, 3], [PolyElement(color = class_colors[Int(i)]) for i in present],
       [replace(string(class_names[Int(i)]), "_" => " ") for i in present], "class"; framevisible = false)
save("esa_worldcover_landcover_class.png", fig)

shown = [name for name in class_names if mean(array(fractions[name])) > 0.01]
ncols = 3
fig = Figure(size = (360ncols + 90, 300 * cld(length(shown), ncols)), fontsize = 14)
for (k, name) in enumerate(shown)
    i, j = fldmod1(k, ncols)
    ax = Axis(fig[i, j]; title = replace(string(name), "_" => " "), xlabel = "longitude", ylabel = "latitude")
    heatmap!(ax, λ_model, φ_model, array(fractions[name]); colormap = :viridis, colorrange = (0, 1))
end
Colorbar(fig[:, ncols + 1]; colorrange = (0, 1), colormap = :viridis, label = "area fraction")
save("esa_worldcover_class_fractions.png", fig)

fig = Figure(size = (1120, 460), fontsize = 15)
ax = Axis(fig[1, 1]; title = "Vegetation fraction f_veg (model grid)", xlabel = "longitude", ylabel = "latitude")
Colorbar(fig[1, 2], heatmap!(ax, λ_model, φ_model, array(fveg_model); colormap = :YlGn, colorrange = (0, 1)); label = "f_veg")
hist!(Axis(fig[1, 3]; title = "distribution of f_veg", xlabel = "f_veg", ylabel = "cells"),
      vec(array(fveg_model)); bins = 30, color = (:seagreen, 0.8))
save("esa_worldcover_f_veg.png", fig)

fraction_sum = sum(array(fractions[name]) for name in class_names)
fig = Figure(size = (1500, 460), fontsize = 15)
ax = Axis(fig[1, 1]; title = "Σ class fractions (≈ 1)", xlabel = "longitude", ylabel = "latitude")
Colorbar(fig[1, 2], heatmap!(ax, λ_model, φ_model, fraction_sum; colormap = :balance, colorrange = (0.95, 1.05)))
ax = Axis(fig[1, 3]; title = "f_veg (native ~110 m)", xlabel = "longitude", ylabel = "latitude")
heatmap!(ax, λ_native, φ_native, array(fveg_native); colormap = :YlGn, colorrange = (0, 1))
ax = Axis(fig[1, 4]; title = "f_veg (model grid)", xlabel = "longitude", ylabel = "latitude")
Colorbar(fig[1, 5], heatmap!(ax, λ_model, φ_model, array(fveg_model); colormap = :YlGn, colorrange = (0, 1)); label = "f_veg")
save("esa_worldcover_fveg_check.png", fig)

```

### Figures

**Majority land-cover class (native ~110 m)**

esa_worldcover_landcover_class

**Per-class area fractions (model grid)**

esa_worldcover_class_fractions

**Vegetation fraction `f_veg` and its distribution**

esa_worldcover_f_veg

**Σ fractions ≈ 1, and native-vs-model `f_veg`**

esa_worldcover_fveg_check

ext/NumericalEarthArchGDALExt/worldcover.jl
```diff
@@ -45,4 +47,161 @@ function NumericalEarth.DataWrangling.IBCAO.reproject_ibcao_to_netcdf(tiff_path,
     return nothing
 end
 
+#####
+##### ESA WorldCover: anonymous COG tiles → regional NetCDF
```
bgroenks96, line 2: Maybe put the WorldCover specific code in a separate file? I guess that goes for all of the other dataset specific utilities as well.

src/DataWrangling/WorldCover/WorldCover.jl
```diff
@@ -0,0 +1,519 @@
...
+                                    bare_sparse_vegetation  = 60,
+                                    snow_and_ice            = 70,
+                                    permanent_water_bodies  = 80,
+                                    herbaceous_wetland      = 90,
+                                    mangroves               = 95,
+                                    moss_and_lichen         = 100)
+
+const ESA_WORLDCOVER_CLASS_CODES = values(ESA_WORLDCOVER_CLASS_NAMES)
+
+# Classes counted as vegetated when forming vegetation_fraction. Excluding
```
bgroenks96: Is this documented somewhere other than a comment?

src/DataWrangling/WorldCover/WorldCover.jl
```diff
@@ -0,0 +1,519 @@
...
+##### Pure aggregation helpers (the main, unit-testable deliverable).
+##### These operate on plain arrays of `UInt8` (or `Integer`) class codes with
+##### no IO. `codes` is a block of fine 10 m pixels covering one coarse cell.
+##### No-data (0) pixels are excluded from every count.
+#####
+
+"""
+    mode_aggregate(codes)
+
+Return the majority (mode) class code over `codes`, ignoring no-data (`0`)
```
bgroenks96: Why not just use `StatsBase.mode` with `skipmissing`...?

src/DataWrangling/WorldCover/WorldCover.jl
```diff
@@ -0,0 +1,519 @@
...
+
+# output
+ESAWorldCover(version = :v200, aggregation_factor = 12)
+```
+
+Data source: https://esa-worldcover.org/en/data-access ;
+DOI v200 `10.5281/zenodo.7254221` (Zanaga et al. 2022), license CC-BY 4.0.
+"""
+struct ESAWorldCover <: AbstractStaticDataset
+    version :: Symbol
```
bgroenks96: I think an `@enum` would be better for this if there are a finite number of valid options.

src/DataWrangling/WorldCover/WorldCover.jl
```diff
@@ -0,0 +1,519 @@
...
+"""
+function class_fractions(codes; class_names = ESA_WORLDCOVER_CLASS_NAMES)
+    fractions = map(c -> class_fraction(codes, c), values(class_names))
+    return NamedTuple{keys(class_names)}(fractions)
+end
+
+"""
+    vegetation_fraction(codes; vegetated_classes = ESA_WORLDCOVER_VEGETATED_CLASSES)
+
+Return the fraction of valid (non-`0`) pixels in `codes` belonging to the
```
bgroenks96: I don't really understand what this is for? Why do you need to aggregate over the set of classes?

xkykai: this is intended to be used as the "vegetation fraction" of each land tile which is to be fed into the vegetation model in https://github.com/NumericalEarth/NumericalEarth.jl/pull/401 (a `f_veg`-like variable in other land models. do you think that to be sensible?

bgroenks96: In principle, yes, but I think I am confused. Wouldn't that require accessing the actual data within a grid cell? I thought `codes` here was just the vector of different class labels. Is it a `Field`?

xkykai: `codes` is not a `Field`, it is all the class labels of every 10 m pixel inside one coarsened cell. We count those labels to get the fraction of each class within the coarsened (N·10 m)² cell, which is what makes fractional quantities well defined there. we can also then use this much more easily since fractions are a continuous field.

src/DataWrangling/WorldCover/WorldCover.jl
xkykai: fixed, documented at constructor of `ESAWorldCover()`

src/DataWrangling/WorldCover/WorldCover.jl
xkykai: the no-data is not `missing`, it's zero instead. 

we could use `mode`, but we need to:
1. skip zero values
2. perform tie-breaking by choosing the lower number. 

with `StatsBase.mode` it'd look like
```julia
# (a) StatsBase does the counting
function majority_class(pixels)
    valid = Iterators.filter(c -> c in ESA_WORLDCOVER_CLASS_CODES, pixels)
    iterate(valid) === nothing && return 0
    return minimum(modes(valid))          # `minimum` gives the lowest-code tie-break
end

# (b) StatsBase does the argmax over the histogram we already have
function majority_class(pixels)
    counts = class_counts(pixels)
    sum(counts) == 0 && return 0
    return mode(collect(ESA_WORLDCOVER_CLASS_CODES), Weights(collect(values(counts))))
end

```

at the same time we already need `class_counts` in order to get class fractions so we might as well do both ourselves (see updated code in https://github.com/NumericalEarth/NumericalEarth.jl/blob/c9bcccf460429a5008b3769fd5d5d511b03975b8/src/DataWrangling/WorldCover/WorldCover.jl#L213-L215) since `mode` doesn't output the class counts and we can then use `class_counts` to get at the mode.
https://github.com/NumericalEarth/NumericalEarth.jl/blob/c9bcccf460429a5008b3769fd5d5d511b03975b8/src/DataWrangling/WorldCover/WorldCover.jl#L226-L238

src/DataWrangling/WorldCover/WorldCover.jl
bgroenks96: Ok, so it would be called from a kernel? Or where is it actually used currently? I only see tests.

src/DataWrangling/WorldCover/WorldCover.jl
bgroenks96: Ok

src/DataWrangling/WorldCover/WorldCover.jl
xkykai: it's used here during the preprocessing of the dataset:
https://github.com/NumericalEarth/NumericalEarth.jl/blob/7bd91931a8b0804a5e70e797b46300b330e71e00/ext/NumericalEarthArchGDALExt/worldcover.jl#L103

it currently only runs on the CPU though, because it is a pre-processing step. it could be written in a kernel though

glwagner (approved):

PR #465, opened 2026-07-21
Fix and automate regional reads of global rasters; add `bare_earth_elevation`

### LINKED ISSUE #498: Regional reads are off by one cell when the on-disk file spans the region rather than the padded native grid
`native_grid` pads a `BoundingBox` region outward by up to one native cell per side (`restrict` brackets the bbox with cell *centers*). Writers are expected to return a file covering that padded grid: `era5_request_area` over-fetches two cells for exactly this reason, and `CopernicusDEM.zarr_to_netcdf` windows on `native_grid(metadatum)`. Nothing checks it, and a file spanning only the region is placed one cell off without a warning.

`set_region_data!` is handed the file's own coordinate vectors — `read_file_coords` reads them from the NetCDF — and `region_info` does resolve the offset from them, then clamps it non-negative:

```julia
di = clamp(i₁ - 1, 0, max(length(λc) - Nx, 0))
```

With a file smaller than the grid the upper bound collapses to `0`, discarding the resolved offset, and the data is written from grid index 1.

MWE — a 4-cell file over `(0, 4)`, read onto the 6-cell grid `restrict` builds for that region:

```julia
using NumericalEarth.DataWrangling: BoundingBox, set_region_data!

# The file: cell centers and data, as `read_file_coords` and `retrieve_data` return them.
λc = φc = collect(0.5:1.0:3.5)
data = reshape(Float64[10i + j for i in 1:4, j in 1:4], 4, 4, 1)

grid = LatitudeLongitudeGrid(CPU(); size = (6, 6), longitude = (-1, 5), latitude = (-1, 5),
λnodes(grid, Center())   # -0.5  0.5  1.5  2.5  3.5  4.5 → the file's first cell is grid cell 2

field = Field{Center, Center, Nothing}(grid)
metadatum = Metadatum(:bottom_height; dataset = ETOPO2022(),   # traits only; nothing read from disk
                      region = BoundingBox(longitude = (0, 4), latitude = (0, 4)))
set_region_data!(field, data, λc, φc, metadatum)
interior(field, :, :, 1)
```

`data` carries centers 0.5 … 3.5, which are grid cells 2 … 5, so it belongs at `[2:5, 2:5]`. It lands at `[1:4, 1:4]`, and the last two cells repeat the file's edge:

```
 11.0  12.0  13.0  14.0  14.0  14.0
 21.0  22.0  23.0  24.0  24.0  24.0
 31.0  32.0  33.0  34.0  34.0  34.0
 41.0  42.0  43.0  44.0  44.0  44.0
 41.0  42.0  43.0  44.0  44.0  44.0
 41.0  42.0  43.0  44.0  44.0  44.0
```

The grid's outer cells have no file coverage, so no lookup can fill them correctly — the file is the thing that is wrong, not the placement. Rerun with a file that does cover the grid (`λc = φc = collect(-0.5:1.0:4.5)`, 6×6 data) and placement is exact.

So `region_info` should compare the file's span against the target grid and error, naming the writer's obligation, instead of clamping the offset and misplacing the data.

xkykai:
## Summary

Reading a window of a global raster (ETOPO, GEBCO, IBCAO/IBCSO, Copernicus DEM) required a
hand-built `BoundingBox` and misplaced the window in three cases. This derives the window from
the target grid and fixes the placement. `bare_earth_elevation` is added on top:
`max(z_DSM − maxₖ object_heightₖ, 0)`, so a DSM's canopy and roof lift stays in the roughness
closure's displacement height instead of being counted twice as terrain.

## Fixes

Each returned wrong data without erroring.

- A Float32 grid node landing a few ULPs below a global file's first longitude label was
  wrapped by +360°, so a grid at the west edge read from the antipode. On ETOPO the offset
  came out as `di = 21239` (longitude +174°) instead of `0` (−180°).
- A window overhanging a global file's east edge was clamped, repeating edge columns, rather
  than continuing across the seam.
- A file materialized on the requested box rather than the padded native grid — which `restrict`
  widens by up to one native cell per side — had its resolved offset clamped away, writing every
  value one cell south-west with the trailing cells repeating the file's edge. `region_info` now
  takes the offset from the file's coordinate labels and rejects a window that falls outside the
  file, naming both spans. Fixes #498.

## API

- `default_region(dataset, grid)` — datasets that must be windowed (Copernicus DEM, ASTER GED)
  derive a padded, coverage-clamped `BoundingBox`; whole-file datasets return `nothing`. A grid
  outside a dataset's coverage, or crossing its longitude seam, now errors instead of
  extrapolating silently.
- `dataset_bounding_box(dataset, grid; padding)`, `BoundingBox(grid; padding)`, and
  `default_horizontal_padding(dataset)` for per-dataset padding.
- `read_windowed_variable` reads only the block covering the grid, so a global file is no
  longer materialized whole.
- `bare_earth_elevation(surface_elevation, object_heights)` and
  `bare_earth_elevation(grid, object_heights; dataset = GLO30(), kw...)`. Object heights combine
  by per-cell maximum with `NaN` counting as zero, so canopy-only and building-only fields
  compose; the difference is clamped at sea level.

The bathymetry cache key includes the region, so a windowed and a global read of the same
dataset and grid cannot collide on disk.

## Example

```julia
grid = LatitudeLongitudeGrid(CPU(); size = (150, 110), longitude = (-60.5, -59.0),
                             latitude = (-3.5, -2.4), topology = (Bounded, Bounded, Flat))

dsm = regrid_topography(grid; dataset = GLO30())   # read window derived from the grid
dtm = bare_earth_elevation(dsm, canopy_height)
```

`examples/bare_earth_terrain.jl` runs that over the central Amazon against a synthetic canopy.
`examples/bare_earth_san_francisco.jl` uses measured 3D-GloBFP building heights and checks
whether GLO-30 contains the lift at all: it carries the low-rise fabric but not the towers,
because its downtown pixels are radar-void infill from building-free DEMs. Both need
`using Zarr` and a `DESTINE_ACCESS_TOKEN`.

## Caveat

`regrid_topography` samples rather than cell-averages when coarsening, so a large coarsening ratio
(37× in the Amazon example) leaves aliased sub-grid texture in the elevation. Unchanged by this PR;
conservative regridding is tracked in CliMA/ClimaOcean.jl#253.

## Tests

`test/test_bare_earth_terrain.jl` covers the subtraction (clamp, per-cell max, `NaN`-as-zero),
region derivation per dataset, and the windowed-read indexing including the three cases above.
`test/test_bare_earth_terrain_downloading.jl` is excluded from CI and exercises the live
download path.


xkykai:
running
```julia
using NumericalEarth, Oceananigans, Zarr   # Zarr enables the GLO-30 read

# Central-Amazon box on a ~55 m grid (near GLO-30's native 30 m)
region = BoundingBox(longitude = (-60.25, -59.55), latitude = (-3.25, -2.55))
grid   = LatitudeLongitudeGrid(CPU();
                               longitude = region.longitude, latitude = region.latitude,
                               size = (1400, 1400), topology = (Bounded, Bounded, Flat))

# A canopy-height field on the grid.
# NOTE: synthetic stand-in here — swap in a measured canopy dataset for a real product.
canopy = set!(Field{Center, Center, Nothing}(grid), 35)

# GLO-30 DSM minus the object heights → bare-earth terrain, in one call:
z_bare = bare_earth_elevation(grid, canopy; dataset = GLO30(), region)
```
gives
bare_earth_terrain_glo30

examples/bare_earth_terrain.jl
```diff
@@ -0,0 +1,182 @@
...
+
+latitude  = -3.5, -2.4
+longitude = -60.5, -59.0
+
+grid = LatitudeLongitudeGrid(CPU(); latitude, longitude, size = (150, 110),
+                            topology = (Bounded, Bounded, Flat))
+
+dsm_dataset = GLO30()
+
+dsm_elevation = regrid_topography(grid; dataset = dsm_dataset)
```
glwagner:
```suggestion
dsm_elevation = regrid_topography(grid; dataset = GLO30())
```

examples/bare_earth_terrain.jl
```diff
@@ -0,0 +1,182 @@
...
+# a canopy-height dataset.
+
+λ, φ, _ = nodes(grid, Center(), Center(), Center())
+elevation = interior(dsm_elevation, :, :, 1)
+
+canopy_height          = 35.0   # m, tall tropical canopy
+terra_firme_elevation  = 30.0   # m, above this is upland forest; below is river floodplain
+
+canopy = Field{Center, Center, Nothing}(grid)
+set!(canopy, ifelse.(elevation .≥ terra_firme_elevation, canopy_height, 0.0))
```
glwagner: do we have to use `interior`? I thought that inequality were supported with binary ops. maybe not.

glwagner: eg something like

```julia
canopy = canopy_height * (elevation ≥ terra_firme_elevation) |> Field
```

is easier to read.

I'm a little confused by the distinction between "canopy" and "canopy height" here.

examples/bare_earth_terrain.jl
```diff
@@ -0,0 +1,182 @@
...
+
+# ## Bare-earth terrain
+#
+# `bare_earth_elevation` subtracts the object heights from the DSM and clamps at sea level.
+# The DSM-minus-bare-earth difference *is* the removed canopy — the signal that belongs to
+# roughness, not terrain.
+
+bare_elevation = bare_earth_elevation(dsm_elevation, canopy)
+
+removed_object_height = dsm_elevation - bare_elevation
```
glwagner:
```suggestion
bare_elevation = bare_earth_elevation(dsm_elevation, canopy)
removed_object_height = dsm_elevation - bare_elevation
```

examples/bare_earth_terrain.jl
```diff
@@ -0,0 +1,182 @@
...
+
+# ## Feeding the atmosphere elevation correction
+#
+# `SlabLand` uses terrain through [`ElevationCorrection`](@ref): the near-surface atmosphere
+# is lapse-corrected over `Δz = z_surface − z_atmosphere`, the gap between the model's
+# surface elevation and the coarse elevation the atmosphere data assumes.
+
+elevation_difference = bare_elevation - reference_elevation
+
+correction = ElevationCorrection(bare_elevation, reference_elevation)
```
glwagner, line 103:
```suggestion
elevation_difference = bare_elevation - reference_elevation
correction = ElevationCorrection(bare_elevation, reference_elevation)
```

examples/bare_earth_terrain.jl
```diff
@@ -0,0 +1,182 @@
...
+band!(transect_axis, λ, vec(interior(bare_transect)), vec(interior(dsm_transect));
+      color = (:seagreen, 0.35), label = "removed canopy (synthetic)")
+lines!(transect_axis, dsm_transect;  color = :black,     label = "DSM")
+lines!(transect_axis, bare_transect; color = :firebrick, label = "bare-earth DTM")
+axislegend(transect_axis; position = :lt)
+
+Label(fig[0, 1:6], rich("Bare-earth terrain over the central Amazon — DSM minus canopy height\n",
+                        rich("canopy height is SYNTHETIC (an elevation-gated stand-in, not measured); terrain is real",
+                             fontsize = 14, color = :firebrick)),
+      fontsize = 20)
```
glwagner, line 150: slightly more compact variable names would probably improve readability here

examples/bare_earth_terrain.jl
glwagner, line 1: this example is not added to `make.jl`

examples/bare_earth_terrain.jl
xkykai: there's not really much of a difference, one is just a sample toy value

examples/bare_earth_terrain.jl
xkykai: I don't intend for this to be a fully built example, right now I'm putting it in `examples/` as a stop gap, but in https://github.com/NumericalEarth/NumericalEarth.jl/issues/479 I will move it to its rightful place

PR #467, opened 2026-07-22
Extend tracer budget closure tests to known non-closure cases

taimoorsohail:
This PR extends the `tracer_budget_tests.jl` file to include two known non-closure cases:

- Fixed Vertical Grids
- `RightFaceFolded` topology with `sea_ice_simulation(dynamics!=nothing)`

This PR will serve to test closure in these two cases, and will assist in fixing any potential non-closure issues.


simone-silvestri:
to have a conservative code with sea ice dynamics we need https://github.com/CliMA/ClimaSeaIce.jl/pull/138 which should be almost ready (I am finishing it)

PR #470, opened 2026-07-23
Generalize the freshwater exchange flux to all advected tracers

jagoosw:
Previously only `T` and `S` received top boundary conditions and the freshwater dilution flux was hard-coded for those two names. Now every advected tracer (all of `tracers` except the CATKE `:e` variable) gets a top boundary condition carrying the dilution flux `(cᴺ - cᶠ) Jʷ`.

- `surface_tracer_value` dispatches generically on the tracer name.
- New `FreshwaterConcentration` wraps a scalar or `Field` concentration `cᶠ`, evaluated live as `cᶠ · Jʷ` by `content_flux_value`; temperature keeps its precomputed source-resolved enthalpy field and salinity its `ZeroField`.
- `freshwater_exchange_flux` now applies the carried flux for a general tracer on every grid, with temperature remaining the exception that only cancels volume movement on mutable grids.
- New `hydrostatic_ocean_simulation` keyword arguments: `freshwater_tracer_concentration` (per-tracer `cᶠ`, default `0`) and `tracer_flux_fields` (per-tracer externally-written surface flux field, defaulting to `Jᵀ`/`Jˢ` for `T`/`S` and `ZeroField` otherwise).

@simone-silvestri

### DIFF
--- a/src/Oceans/Oceans.jl
+++ b/src/Oceans/Oceans.jl
@@ -11,7 +11,7 @@ using Oceananigans.BoundaryConditions: DefaultBoundaryCondition, DiscreteBoundar
                                        FieldBoundaryConditions, FluxBoundaryCondition, getbc
 using Oceananigans.BuoyancyFormulations: SeawaterBuoyancy
 using Oceananigans.Coriolis: HydrostaticSphericalCoriolis
-using Oceananigans.Fields: Field, CenterField, set!, interior
+using Oceananigans.Fields: Field, CenterField, AbstractField, set!, interior
 using Oceananigans.Forcings: MultipleForcings, DiscreteForcing
 using Oceananigans.Grids: Grids, inactive_node, Face, Center, xspacings, yspacings, RectilinearGrid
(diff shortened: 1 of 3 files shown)


simone-silvestri:
The `FreshwaterConcentration` you have added is another way to do what 
https://github.com/NumericalEarth/NumericalEarth.jl/blob/f85931cfe194d182491b5c8d180deb9a58883828/src/Oceans/assemble_net_ocean_fluxes.jl#L151
already does for temperature.

Is there a way to unify the two? Where do you assemble the tracer fluxes?

jagoosw:
Oh yeah I'll work out a way to make them both go together. Is it correct that the virtual heat and salt flux don't include the input of fresh water from sea ice here?
https://github.com/NumericalEarth/NumericalEarth.jl/blob/f85931cfe194d182491b5c8d180deb9a58883828/src/Oceans/assemble_net_ocean_fluxes.jl#L129-L151
And why why temperature doesn't change on immutable grids here?
https://github.com/NumericalEarth/NumericalEarth.jl/blob/f85931cfe194d182491b5c8d180deb9a58883828/src/Oceans/ocean_simulation.jl#L89-L91

PR #474, opened 2026-07-27
Extend budget diagnostics to salinity and mass

taimoorsohail:
## Summary

- extend `BudgetComputation` from temperature to salinity-content and mass-content budgets
- expose quantity-specific public term names for heat, salinity, and mass
- include explicit freshwater volume fluxes in freshwater-content diagnostics
- use salt-content fluxes for salinity budgets and admitted freshwater mass for mutable-grid mass budgets
- preserve the fixed-grid virtual-salinity-flux formulation and component consistency
- add focused fixed-grid and mutable-grid diagnostic coverage

## Why

The freshwater-volume coupling was moved into `interfaces.net_fluxes.ocean.η`, while the freshwater-content diagnostic still inferred freshwater solely from the stored salinity flux. Because the stored salinity flux now contains only salt exchange, the explicit volumetric contribution was omitted. This also made the new salinity and mass budget surface terms use the wrong physical quantities.

## User impact

`net_ocean_freshwater_flux` now reports an outward-positive freshwater-content flux that includes surface volume changes. Salinity budgets use the applied salt-content surface flux, while mass budgets include only mass actually admitted through a mutable surface. Fixed-volume grids retain their virtual-salinity-flux treatment. The quantity-specific `BudgetComputation` property names remain in place.

## Validation

- `git diff --check`
- focused fixed-grid and mutable-grid assertions added to `test/test_diagnostics_2.jl`
- local Julia tests were not run, following the repository instruction to use GitHub CI for test execution


src/Diagnostics/budget_diagnostic.jl
```diff
@@ -0,0 +1,303 @@
...
+struct BudgetComputation{H, P, T, S, R, A, B}
+    tracer_name :: Symbol
+    heat_content :: H
+    previous_heat_content :: P
+    tendency :: T
+    surface_flux :: S
+    radiative_heat_flux :: R
+    applied_radiative_heat_flux :: A
+    residual :: B
+end
```
simone-silvestri, line 41: maybe we can avoid the `getfield` overload by defining

```suggestion
struct BudgetComputation{H, P, T, S, R, A, B}
    tracer_name :: Symbol
    content :: H
    previous_content :: P
    tendency :: T
    fluxes :: S
    residual :: B
end
```
where for :temperature we allocate
```
fluxes = (; surface_flux, radiative_flux, applied_radiative_heat_flux)
```
while for the others
```
fluxes = (; surface_flux)
```

src/Diagnostics/budget_diagnostic.jl
```diff
@@ -0,0 +1,303 @@
...
+
+budget_inventory(::Val{:salinity}, esm) =
+    Field(Integral(esm.ocean.model.tracers.S, dims=3))
+
+function budget_inventory(::Val{:mass}, esm)
+    ocean = esm.ocean
+    model = ocean.model
+    ρᵒᶜ = reference_density(ocean)
+    return Field(Integral(0 * model.tracers.T + ρᵒᶜ, dims=3))
```
simone-silvestri, line 138:
```suggestion
    grid = esm.ocean.grid
    ρᵒᶜ = set!(CenterField(grid), reference_density(esm.ocean))
    return Field(Integral(ρᵒᶜ, dims=3))
```

PR #424, opened 2026-07-09
MODIS leaf area index and land cover, with a class-aware seasonal gap fill

xkykai:
**Merge only after CliMA/Oceananigans.jl#5925 lands**, then replace the local `time_average` with a `DateTime` wrapper over Oceananigans' one.

Ingests MODIS MCD15A2H leaf area index and MCD12Q1 land cover, builds a multi-year seasonal climatology from the first, and fills what compositing leaves behind with a class-aware gap fill: time interpolation, then a same-class donor curve scaled to the cell's own level, then the local class mean. Closes #650.

Shared read-path changes: a read onto the native grid copies instead of interpolating (the bilinear stencil dilates every `NaN`), a one-cell region offset on fine Float32 grids is fixed on both axes, `averaging_window` is defined for the MODIS composites and the ten-day CGLS albedo files, and a whole seasonal climatology no longer warns that `Cyclical()` wraps.

Public API:
- Exported: `MCD15A2H`, `MCD12Q1`, `MODISLAIClimatology`, `build_lai_climatology!`. Both datasets need `ArchGDAL` and a NASA Earthdata login.
- `DataWrangling`: `fill_gaps!` (gains `cyclic` and a per-column `max_gap`), `fill_seasonal_gaps!`, `gap_fill_denial`, `time_average`, `window_bounds`, `class_fractions`.
- `DataWrangling.MODISLand`: the class tables and helpers (`igbp_class_names`, `class_maximum_gap`, `zero_non_vegetated!`, `period_index`, `retained_retrieval_metadatum`, `lai_screening_mask`).

```julia
using NumericalEarth, ArchGDAL, Oceananigans, Dates
using NumericalEarth.DataWrangling: fill_seasonal_gaps!
using NumericalEarth.DataWrangling.MODISLand: class_maximum_gap, igbp_non_vegetated_classes

region = BoundingBox(longitude = (-92.5, -91.5), latitude = (36.5, 37.5))
classes = Field(Metadatum(:landcover_class; dataset = MCD12Q1(), region, date = DateTime(2019)))

climatology = MODISLAIClimatology(years = 2015:2019)
leaf_area_index = FieldTimeSeries(Metadata(:leaf_area_index; dataset = climatology, region);
                                  time_indices_in_memory = 46)

fill_seasonal_gaps!(leaf_area_index, classes; max_gap = class_maximum_gap(classes),
                    valid_range = (0, 10), unfilled_classes = igbp_non_vegetated_classes)
```

PR #478, opened 2026-07-29
Place window-averaged samples at their window centers

### LINKED ISSUE #477: ECCO4Monthly and EN4Monthly dates sit 15 days before the files' own time coordinate
`all_dates` gives `ECCO4Monthly` and `EN4Monthly` a node time at the start of each month, but both products state in their own `time` and `time_bnds` that each value is the mean over that month, valid at the month's center. A `FieldTimeSeries` pins a value at its node time and interpolates linearly between nodes, so every sample is read about 15 days before the time it describes.

EN4 surface temperature at 40°N, 60°W, 2010:

| value in the file | what the file says it is | node time assigned |
|---|---|---|
| 293.82 K | mean over `[Jun 1, Jul 1)`, `time` = 2010-06-16T00:00 | 1 June |
| 299.85 K | mean over `[Jul 1, Aug 1)`, `time` = 2010-07-16T12:00 | 1 July |

Ask that series for 1 July and it returns the node value, 299.85 K. With the nodes at the times the files give, it returns 296.79 K, essentially the midpoint of the two monthly means, because 1 July falls halfway between 16 June and 16 July. So the model gets mid-July's ocean on 1 July: two weeks early and 3 K too warm. Across 2010 the two axes differ by up to 3.07 K at that cell, against a 12.79 K seasonal range.

```julia
using NumericalEarth, Dates, NCDatasets
using NumericalEarth.DataWrangling: metadata_path
using Downloads: download

metadatum = Metadatum(:temperature; dataset = EN4Monthly(), date = DateTime(2010, 6, 1))
download(metadatum)

NCDataset(metadata_path(metadatum)) do ds
    ds["time"][1], ds["time_bnds"][:, 1]   # 2010-06-16T00:00, [2010-06-01, 2010-07-01]
end
```

ECCO is the same: `NVEL_1993_01.nc` carries `time = 1993-01-16T12:00` with bounds `[1993-01-01, 1993-02-01]`, and PO.DAAC documents the V4r4 monthly `time` variable as "center time of averaging period" ([ECCO_L4_OCEAN_VEL_05DEG_MONTHLY_V4R4](https://podaac.jpl.nasa.gov/dataset/ECCO_L4_OCEAN_VEL_05DEG_MONTHLY_V4R4)).

- Declared at [`ECCO.jl#L66`](https://github.com/NumericalEarth/NumericalEarth.jl/blob/aaaef71f173af41582f312bb70ad80e49bbc8fba/src/DataWrangling/ECCO/ECCO.jl#L66), [`#L69`](https://github.com/NumericalEarth/NumericalEarth.jl/blob/aaaef71f173af41582f312bb70ad80e49bbc8fba/src/DataWrangling/ECCO/ECCO.jl#L69) and [`EN4.jl#L28`](https://github.com/NumericalEarth/NumericalEarth.jl/blob/aaaef71f173af41582f312bb70ad80e49bbc8fba/src/DataWrangling/EN4/EN4.jl#L28).

`all_dates` should keep returning month starts, since they are the file key and the centers are not a `Month(1)` range (16 June 00:00 to 16 July 12:00 is 30.5 days). The node belongs at the window center instead, applied where the date becomes a time. `set!` at a single date is unaffected.

xkykai:
Fixes #477.

`ECCO4Monthly` and `EN4Monthly` files hold a calendar-month mean and carry that window's center in their own `time` variable, while `all_dates` names them by the first of the month. `native_times` used the naming date as the `FieldTimeSeries` node time, so an EN4 series asked for 1 July returned July's mean, 3 K too warm at 40°N 60°W where the files' own dates give 296.79 K.

- `sample_window(metadatum)` — the `(start, stop)` dates of the window a value represents, and the only thing a dataset declares. Zero-width by default, so instantaneous products are unchanged. `window_center`, `time_window_offset`, `sample_window_span` and `uncovered_time_gaps` derive from it, and `native_times` places each node at its window center.
- Declared for every monthly-mean product (ECCO2/ECCO4, both ECCO-Darwin, EN4, AVISO, GLORYS, WOA, ERA5 monthly single-level and pressure-level), and for ERA5's hourly accumulations and mean rates, which cover the hour *ending* at the stamp.
- A cyclical series repeats over the span its windows tile: 365 days for twelve monthly means, against the 364.5 inferred from node spacing.
- Nodes sit half a window inside the span the data covers, so `FieldTimeSeries` warns under `Cyclical` — which fills the ends by wrapping — and throws under `Linear`, which extrapolates without bound past the last node. `Clamp`, instantaneous products and single samples are untouched.

`all_dates` still returns the first of the month: it is the file key, and the centers are not a `Month(1)` range (16 June 00:00 to 16 July 12:00 is 30.5 days).

Not covered here: the coupled atmosphere applies one time interpolator, built from `u.times`, to the precipitation fluxes as well, so ERA5's shifted rain nodes have no effect there (radiation interpolates on its own axis and does pick them up); and the daily products (`ECCO2Daily`, `AVISODaily`, `GLORYSDaily`) need their `time_bnds` checked before a window is declared.

#424 adds the same functions for MODIS composites; whichever lands first, the other drops its copy.


src/DataWrangling/ECCO/ECCO.jl
```diff
@@ -70,6 +70,11 @@ DataWrangling.all_dates(dataset::ECCO4Monthly, variable) = metadata_epoch(datase
 DataWrangling.all_dates(dataset::ECCO2Monthly, variable) = metadata_epoch(dataset) : Month(1) : DateTime(2024, 12, 1)
 DataWrangling.all_dates(dataset::ECCO2Daily,   variable) = metadata_epoch(dataset) : Day(1)   : DateTime(2024, 12, 31)
 
+# A V4r4 monthly file holds the mean over a calendar month and carries that window's center
+# in its own `time` variable, so its value belongs half a month after the date naming it.
```
simone-silvestri: I don't think this comment is related to the line of code below
```suggestion
```

src/DataWrangling/EN4/EN4.jl
```diff
@@ -26,6 +26,12 @@ struct EN4Monthly end
 
 Base.size(::EN4Monthly, variable) = (360, 173, 42)
 DataWrangling.all_dates(::EN4Monthly, variable) = DateTime(1900, 1, 1) : Month(1) : DateTime(2024, 12, 1)
+
+# An analysis file holds the mean over a calendar month and carries that window's center in
+# its own `time` variable, so its value belongs half a month after the date naming it.
```
simone-silvestri:
```suggestion
```
same thing

simone-silvestri (approved):

simone-silvestri (approved):

PR #393, opened 2026-06-30
Implement MultiYearERA5 dataset type for yearly ERA5 files

aklocker42:
## Summary
Adds `MultiYearERA5` dataset type for using yearly ERA5 files (8784 hours per file) instead of hourly files. Integrates with the new `yearly()` function in CopernicusClimateDataStore.jl.

## Changes
- Add `MultiYearERA5` dataset type in ERA5_single_levels.jl
- Implement ERA5_field_time_series.jl for reading from yearly files
- Fix `build_era5_area()` to return array format compatible with CDS API
- Map FieldTimeSeries time indices to yearly file time indices correctly
- Support both "time" and "valid_time" NetCDF dimension names
- Remove verbose logging for cleaner output
- Prevent date restriction leakage via `additional_kw`

## API
```julia
# Use MultiYearERA5 instead of ERA5Hourly
atmosphere = ERA5PrescribedAtmosphere(arch;
    dataset=MultiYearERA5(),  # Downloads yearly files
    start_date, end_date, dir, region
)
```

## Key Fixes
1. **Area format bug**: `build_era5_area()` now returns `[south, west, north, east]` array instead of NamedTuple
2. **Time indexing**: Correctly maps simulation time indices to yearly file indices
3. **Variable names**: Handles both "time" and "valid_time" dimension names from CDS

## Testing
- ✅ Full bouveco ocean simulation ran 21+ hours successfully
- ✅ All 8 ERA5 variables (6 atmosphere + 2 radiation) downloaded correctly
- ✅ Simulation completed without errors or crashes

## Dependencies
- Requires corresponding changes in CopernicusClimateDataStore.jl

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

PR #484, opened 2026-07-31
Add ERA5-Land datasets (ERA5HourlyLand, ERA5MonthlyLand)

xkykai:
Supersedes #400 (same dataset types and grid conventions, plus a working monthly-means download path, `time_averaged_field`, the example, docs, and tests) and adopts for land the yearly-file storage layout that #393 introduced for single levels.

Adds the ERA5-Land reanalysis (0.1°, ~9 km, land-only) to `DataWrangling`, so soil state can be loaded as `Field`s / `FieldTimeSeries` like any other dataset. The motivating use is the slab land model's lower thermal boundary: `WaterCoupledEnergy` accepts `deep_temperature` as a `Field`, and a multi-year mean of ERA5-Land soil temperature level 4 (`stl4`, 100–289 cm) drops straight into it.

## API

- `ERA5HourlyLand` — hourly `reanalysis-era5-land` (1950–present): `soil_temperature_level_1..4`, `volumetric_soil_water_layer_1..4`, `skin_temperature`, `temperature` (2 m), `dewpoint_temperature`, `snow_water_equivalent`, `snow_depth`. Ocean cells load as `NaN`.
- `ERA5MonthlyLand` — the `monthly_averaged_reanalysis` companion, one field per calendar month; suited to climatological soil state.
- `time_averaged_field(metadata, arch_or_grid; kw...)` — generic time mean over a multi-date `Metadata`, on the native grid or interpolated onto a model grid. Missing (`NaN`) cells stay `NaN` unless filled with `inpainting`.

Both datasets subtype `ERA5Dataset`, so the filename utilities, netCDF reader, latitude flip, regional cropping, and regridding are all inherited. The overrides are the 0.1° grid (`(3600, 1800, 1)`, with the same poleward-row `AverageNorthSouth` convention as the 0.25° single-level grid), the date ranges, the variable-name dictionaries, and the horizontal padding (two native 0.1° cells).

## Storage layout: one file per variable per year

ERA5-Land uses the `ERA5YearlySingleLevel` / `MultiYearJRA55` layout: every date of a year maps to the same per-variable yearly file, and timesteps are located by matching the file's own time axis. `retrieve_data` reads single dates out of the yearly file, and a dedicated `FieldTimeSeries` `set!` fills in-memory windows that straddle year boundaries by resolving the needed files per slot through the metadata's datewise filenames (the `MultiYearJRA55` pattern — the existing `ERA5YearlySingleLevel` reader is single-file). For hourly data this is 8760× fewer files than one-file-per-date.

## CDS download path

ERA5-Land requests differ from ERA5 single-levels in two ways, handled by two new dispatch hooks in `build_era5_request`:

- `request_product_type!`: the `reanalysis-era5-land` catalogue entry is not keyed by product type (sending `product_type` is rejected); the monthly entry uses `monthly_averaged_reanalysis`.
- `request_dates!`: monthly means carry one field per (year, month) — no `day` key, `time = "00:00"`.

Yearly files are fetched whole. A year of monthly means fits one CDS request (12 fields), issued for all requested variables at once and split per variable; a year of hourly data is rejected by the ERA5-Land per-request cost limit ("cost limits exceeded", verified live), so hourly years are fetched one variable and calendar month at a time — the same request count as a per-date layout — and concatenated locally into the yearly file by a new `concatenate_era5_nc`. Existing single-level / pressure-level request dictionaries are unchanged (covered by the request-construction tests).

Verified against the live CDS: a monthly-means year file (one request, 12 timesteps), an hourly year file (12 month-chunks concatenated to a monotonic 8784-step axis), windowed `FieldTimeSeries` reads across a year boundary bit-identical to full-in-memory reads, and netCDF short names (`stl4`, `stl1`, …) and 0.1° north-to-south coordinates matching the dictionaries.

## Example

`examples/era5_land_deep_temperature.jl` (registered with `build_always = false`) demonstrates the intended end-to-end use over the European Alps:

1. downloads a five-year monthly climatology of all four soil temperature levels and checks the seasonal cycle damps and lags with depth (16.6 K amplitude at 0–7 cm → 9.8 K at 100–289 cm, peak shifted August → September) — the heat-diffusion signature that makes `stl4` the right slowly-varying lower boundary;
2. time-averages `stl4` into a static `Tᵈᵉᵉᵖ` map and lapse-corrects it from ERA5-Land's ~9 km cell-mean elevation to a ~1 km grid, using ETOPO 2022 coarse-grained to the native 0.1° grid as the reference elevation (corrections up to ±11 K across ridges and valleys);
3. runs a 4-day ERA5-forced `SlabLand` (`WaterCoupledEnergy` + `VariablySaturatedHydrology`) restoring to the corrected field, against a control restoring to the hand-set default 280 K. The two runs have nearly identical domain means but the control develops a ±5 K terrain-imprinted error (too cold in the Po valley, too warm on the high ridges) — the imprint of `280 − Tᵈᵉᵉᵖ`.

## Tests

New testsets in `test/test_cds_downloading.jl`: grid size/interfaces and date cadences, variable-dictionary consistency, request construction (no `product_type` for hourly land; `monthly_averaged_reanalysis`, no `day` for monthly), the yearly-file layout (filenames, per-year request batches, per-variable grouping), a synthetic `concatenate_era5_nc` fixture, and a live download testset that reads `stl4` back from the yearly file, checks the 0.1° coordinates, value ranges over the Alps, `time_averaged_field` against a hand-computed two-month mean, and NaN-free interpolation onto a model grid. ExplicitImports checks pass.

## Caveat: lapse rate for the deep-temperature correction

The free-air environmental lapse rate (Γ = 6.5 K km⁻¹, `ElevationCorrection`'s default) overestimates the *ground*-temperature lapse rate in snowy terrain, because snow insulates high-altitude soil from winter air: regressing the coarse 0.1° `Tᵈᵉᵉᵖ` against the 0.1° elevation over the Alps box gives −4.2 K km⁻¹ (correlation −0.89), so a hard-coded 6.5 would over-cool resolved ridges by up to ~3 K. The example therefore fits Γ from the coarse product itself (`ground_lapse_rate = -cov(z, Tᵈ)/var(z)` on the native grid), which is self-consistent and adapts to region and snow regime; the forcing-side `ElevationCorrection` keeps the free-air default since it lifts the air state. Residual caveat: the fit measures the between-cell relationship at 9 km and assumes it holds within cells — aspect and snow-redistribution heterogeneity below that scale is not captured either way.

## Notes

- Out of scope, observed while working here: `ERA5MonthlySingleLevel` / `ERA5MonthlyPressureLevels` CDS downloads still route to the hourly catalogue entries with `product_type = "reanalysis"`, so they fetch an instantaneous snapshot rather than a monthly mean. The new `request_product_type!` / `request_dates!` hooks are the natural place to fix that in a separate PR.
- ERA5-Land accumulated variables (radiation, precipitation, runoff) reset daily rather than hourly, so the single-level accumulation conversions must not be inherited; they are deliberately not in the variable dictionaries yet.


bgroenks96:
@xkykai Let me know when you're ready for a review.

xkykai:
close as no longer relevant

PR #486, opened 2026-07-31
Implement river spread and river mixing

simone-silvestri:
When we have a river inlet in the ocean, the salinity can drop to zero since the incoming freshwater sits on the first cell and creates a lens which never mixes.

For this reason this PR introduces a couple of common strategies in climate modeling
- heightened diffusivity at the river mouth (introduced by default when passing a river routing to the ocean_simulation)
- horizontal spread of freshwater fluxes coming from the river (a tunable parameter, 1.2 degrees spread by default)

It also refactors a bit the river routing to apply also to datasets like JRA55


src/Lands/river_routing.jl
```diff
@@ -153,6 +176,37 @@ function build_river_routing(target_grid, outlet_i, outlet_j, outlet_λ, outlet_
...
+                            outlet_detection_snapshots = 365)
+
+    source_grid = on_architecture(CPU(), flux_time_series.grid)
+    kᴺ = size(source_grid, 3)
+
+    # Scan a full seasonal cycle so intermittent and seasonally frozen rivers stay in the map.
+    outlet_mask = Array(interior(flux_time_series[1]))[:, :, 1] .> 0
+    for n in 2:min(outlet_detection_snapshots, length(flux_time_series.times))
+        outlet_mask .|= Array(interior(flux_time_series[n]))[:, :, 1] .> 0
+    end
```
xkykai: I understand why this is done, but why don't we scan through the entire `flux_time_series`, and only decide the span in which we check for rivers with `Metadata` which is used to create `flux_time_series`?

src/Lands/river_routing.jl
```diff
@@ -161,68 +215,88 @@ end
...
 
-# Index of the entry of sorted vector `a` closest to `x`.
-function searchsortednearest(a, x)
-    i = searchsortedfirst(a, x)
-    i == 1 && return 1
-    i > length(a) && return length(a)
-    return abs(a[i-1] - x) ≤ abs(a[i] - x) ? i - 1 : i
-end
+node_λφ(λc::AbstractVector, φc::AbstractVector, i, j) = (λc[i], φc[j])
+node_λφ(λc::AbstractMatrix, φc::AbstractMatrix, i, j) = (λc[i, j], φc[i, j])
```
xkykai: I wonder if it is possible to not define these functions and hard code at the call site instead since this does something very specific, up to you

src/Lands/river_routing.jl
```diff
@@ -161,68 +215,88 @@ end
...
+    end
+    nearest_distance < reach || return Tuple{Int, Int}[]
+
+    # The footprint is centered on the landing cell, so mouths relocated onto the shelf (the Ob and
+    # Yenisei move 2-3°) keep a full footprint.
+    λ★, φ★ = wet.λ[nearest], wet.φ[nearest]
+    footprint = isnothing(spread_radius) ? reach : spread_radius^2
+    targets = [(squared_distance(λ★, φ★, wet.λ[n], wet.φ[n]), n) for n in reachable]
+    filter!(t -> first(t) ≤ footprint, targets)
+    sort!(targets; by = first)
```
xkykai, line 291: not quite sure what this does, maybe a bit more explanation here could be helpful

xkykai (approved):

src/Lands/river_routing.jl
simone-silvestri: right, I scan the whole window now.

src/Lands/river_routing.jl
simone-silvestri: Implemented in the docstring

src/Lands/river_routing.jl
simone-silvestri: Applied Oceananigans' `*node` functions instead

PR #489, opened 2026-08-01
Replace GEBCO2024 with GEBCO2026 using direct CEDA download

aklocker42:
## Summary

- The old `GEBCO2024` downloaded via the BODC web-form URL (`bodc.ac.uk/data/open_download/.../zip/`), which intermittently returns corrupt ZIP archives rather than a valid file.
- Replace with `GEBCO2026`, downloading the NetCDF directly from the CEDA DAP server — no ZIP handling, no `ZipFile` dependency in the GEBCO module.
- Follows the same pattern as `IBCSOv2` (direct `Downloads.download` to the cache path).

The CEDA URL structure is stable across releases:
```
https://dap.ceda.ac.uk/bodc/gebco/global/gebco_{year}/ice_surface_elevation/netcdf/GEBCO_{year}.nc
```

## Test plan

- [ ] `test_polar_bathymetry.jl` — `GEBCO2026` metadata tests (size, variable name, filename suffix)
- [ ] Manual: `regrid_bathymetry(grid; dataset=GEBCO2026())` downloads and regrids without error


glwagner (approved):
seems strictly better.

should we remove the year from the dataset title? eg just `dataset=GEBCO()`?

that way updating this will not introduce a breaking change for future years

Issue #480, opened 2026-07-30
[WIP] Lacking land data downloading tests for now


xkykai:
This is a note to self for the missing tests for (land) data downloading which I have not included since downloading tests and pipeline are flakey for now. 

Once test infrastructure has been cleaned up (https://github.com/NumericalEarth/NumericalEarth.jl/pull/440) and data downloading testing has been moved to regular schedule rather than per commit I will include regularly-scheduled downloading tests for datasets below (will continually be updated as new land data ingestion is being merged):

1. Copernicus Albedo (https://github.com/NumericalEarth/NumericalEarth.jl/pull/423)
2. OpenLandMap-SoilDB (https://github.com/NumericalEarth/NumericalEarth.jl/pull/426)

aklocker42:
Adding ERA5-Land (PR #490) to the list — live downloading tests now exist in `test_copernicus_climate_datastore_downloading.jl` (excluded from per-commit CI, same pattern as the others here):

3. ERA5-Land (https://github.com/NumericalEarth/NumericalEarth.jl/pull/490)

PR #490, opened 2026-08-01
Add ERA5-Land support via the pure-Julia CopernicusClimateDataStore client

aklocker42:
## Summary

Adds `ERA5HourlyLand` / `ERA5MonthlyLand` (`reanalysis-era5-land`, 0.1° land-surface
reanalysis: skin/soil temperature, soil moisture, 2 m temperature/dewpoint, snow
depth/water-equivalent) to `DataWrangling.ERA5`, downloaded through the pure-Julia
`CopernicusClimateDataStore.jl` client rather than the Python `cdsapi` path used by
`NumericalEarthCDSAPIExt`.

- **Storage layout**: one file per (variable, year), matching the `ERA5YearlySingleLevel`
  layout introduced in #393 — `retrieve_data`/`set!` are shared via a widened
  `ERA5YearlyFileDataset` type alias rather than duplicated.
- **Download path**: a year of monthly means fits one CDS request; a year of hourly
  data exceeds ERA5-Land's per-request cost limit, so it's fetched one calendar month
  at a time via `CopernicusClimateDataStore.retrieve()` and the chunks are concatenated
  locally (`concatenate_era5_nc`) into the year's file.
- **Zip-wrapped responses**: the CDS gateway sometimes wraps ERA5-Land's netcdf output
  in a zip archive even when `format=netcdf` is requested; discovered while validating
  against real CDS (see below) and handled by unwrapping in place before concatenation.

## Relationship to #484

@xkykai's #484 (open, not yet merged) implements the same `ERA5HourlyLand`/`ERA5MonthlyLand`
dataset types and yearly-file layout against the Python-`cdsapi`-based
`NumericalEarthCDSAPIExt`. This PR is an independent implementation on top of the
pure-Julia `CopernicusClimateDataStore.jl` extension instead, so it carries no Python
runtime dependency. The month-chunked batching, retry-wrapped retrieval, and NetCDF
concatenation approach for assembling a year of hourly data from CDS's per-request
cost-limited chunks are adapted from #484's `download_era5_land_year` /
`group_by_calendar_month` / `concatenate_era5_nc` / zip-detection pattern — thank you
@xkykai for working through that design first. Maintainers will need to decide which
download backend (this pure-Julia path vs. #484's Python path) to keep going forward;
this PR doesn't touch `NumericalEarthCDSAPIExt` and can coexist with #484 as-is.

## Testing

- 102 offline unit tests (`test/test_copernicus_climate_datastore.jl`): dataset type
  properties, grid interfaces, variable-name dictionaries, and download dispatch —
  all pass.
- Live validation against real CDS: downloaded a full year (2020) of hourly
  `skin_temperature` for a Scandinavia bounding box, produced spatial snapshots and
  an hourly time series at a point near Oslo. Land/sea masking matches the coastline,
  January/July fields show the expected seasonal reversal (mild Norwegian west coast
  vs. cold interior in winter), and the point time series shows a realistic seasonal
  and diurnal cycle.


src/DataWrangling/ERA5/ERA5_land.jl
```diff
@@ -0,0 +1,112 @@
...
+
+function DataWrangling.metadata_filename(dataset::ERA5LandDataset, name, date, region)
+    var = ERA5Land_dataset_variable_names[name]
+    year = Dates.year(date)
+    suffix = region_suffix(region)
+    return string(var, "_", dataset_name(dataset), "_", year, suffix, ".nc")
+end
+
+DataWrangling.build_filename(dataset::ERA5LandDataset, name, dates::AbstractArray, region) =
+    DataWrangling.metadata_filename(dataset, name, dates[1], region)
```
xkykai: I think this causes multi-year data to be mapped into the first year's file. We should probably use datewise filenames and check https://github.com/NumericalEarth/NumericalEarth.jl/blob/9691d45c3b81fa84dc9b8b838e20575ad9185ee4/src/DataWrangling/ERA5/ERA5_field_time_series.jl#L148-L203 for `set!`

src/DataWrangling/ERA5/ERA5_field_time_series.jl
xkykai: I think with the `build_filename` this will go to the wrong year if it is not the first

xkykai: `set!` might error if it is downloading beyond the year of the first date. try running it to verify

src/DataWrangling/ERA5/ERA5_land.jl
xkykai: note that https://github.com/NumericalEarth/NumericalEarth.jl/blob/9991db381164235a90401d7716323625211eac47/src/DataWrangling/metadata.jl#L800-L802 uses a `DatewiseFilename` per date formulation

xkykai: The solution I think is to remove this method (so it uses the default and then apply `set!` with a pattern similar to https://github.com/NumericalEarth/NumericalEarth.jl/pull/490/changes#r3700653976

test/test_copernicus_climate_datastore.jl
```diff
@@ -142,5 +144,122 @@ const CDSExt = Base.get_extension(NumericalEarth, :NumericalEarthCopernicusClima
...
+        @test first(hourly_dates) == DateTime("1950-01-01")
+        @test step(hourly_dates) == Hour(1)
+
+        monthly_dates = NumericalEarth.DataWrangling.all_dates(monthly_land, :skin_temperature)
+        @test first(monthly_dates) == DateTime("1950-01-01")
+        @test step(monthly_dates) == Month(1)
+    end
+
+    @testset "ERA5-Land download dispatch" begin
+        date = DateTime(2020, 1, 1)
```
xkykai, line 233: can we also exercise a multi-year download test?

also cf https://github.com/NumericalEarth/NumericalEarth.jl/pull/440 we should probably move downloading tests out of the per-commit ci runs

ext/NumericalEarthCopernicusClimateDataStoreExt.jl
```diff
@@ -246,6 +278,187 @@ function Downloads.download(meta::NumericalEarth.DataWrangling.Metadatum{<:Union
...
+end
+
+function era5_land_request(variable_name, ::ERA5HourlyLand, dts, area)
+    request = Dict{String, Any}(
+        "variable" => [variable_name],
+        "year"     => unique(string.(Dates.year.(dts))),
+        "month"    => unique(lpad.(string.(Dates.month.(dts)), 2, '0')),
+        "day"      => unique(lpad.(string.(Dates.day.(dts)), 2, '0')),
+        "time"     => unique([lpad(string(Dates.hour(dt)), 2, '0') * ":00" for dt in dts]),
+        "format"   => "netcdf",
```
xkykai: if you do 
```suggestion
	"data_format"     => "netcdf",
	"download_format" => "unarchived"
```
does the "sometimes downloads in zip" problem go away? seems to be an old CDS API key problem

ext/NumericalEarthCopernicusClimateDataStoreExt.jl
```diff
@@ -246,6 +278,187 @@ function Downloads.download(meta::NumericalEarth.DataWrangling.Metadatum{<:Union
...
+    magic = open(io -> read(io, 2), path, "r")
+    length(magic) == 2 && magic == UInt8[0x50, 0x4b] || return path
+    mktempdir(dirname(path)) do tmp_dir
+        run(`unzip -o -q $path -d $tmp_dir`)
+        extracted = filter(isfile, readdir(tmp_dir; join=true))
+        isempty(extracted) && error("CDS response at $path was a zip archive but contained no files")
+        mv(first(extracted), path; force=true)
+    end
+    return path
+end
```
xkykai: seems dangerous

ext/NumericalEarthCopernicusClimateDataStoreExt.jl
```diff
@@ -246,6 +278,187 @@ function Downloads.download(meta::NumericalEarth.DataWrangling.Metadatum{<:Union
...
+        catch e
+            attempt < max_retries || rethrow(e)
+            @warn "CDS retrieve attempt $attempt/$max_retries failed for $product; retrying..." exception=(e, catch_backtrace())
+            sleep(5.0 * attempt)
+        end
+    end
+end
+
+# Append single-variable NetCDF chunks along the time dimension into one yearly file.
+function concatenate_era5_nc(src_paths, dst_path, nc_varname)
```
xkykai: is `nc_varname` used?

ext/NumericalEarthCopernicusClimateDataStoreExt.jl
```diff
@@ -246,6 +278,187 @@ function Downloads.download(meta::NumericalEarth.DataWrangling.Metadatum{<:Union
...
+        isempty(extracted) && error("CDS response at $path was a zip archive but contained no files")
+        mv(first(extracted), path; force=true)
+    end
+    return path
+end
+
+function group_by_calendar_month(datetimes)
+    keys = unique([(Dates.year(dt), Dates.month(dt)) for dt in datetimes])
+    return Dict(k => filter(dt -> (Dates.year(dt), Dates.month(dt)) == k, datetimes) for k in keys)
+end
```
xkykai, line 280: out of curiosity, what is this used for?

ext/NumericalEarthCopernicusClimateDataStoreExt.jl
```diff
@@ -153,14 +159,38 @@ function date_keywords(::ERA5MonthlyPressureLevels, date)
...
+"""
+    cds_dataset_keyword(dataset)
+
+Select the `dataset` keyword (`:era5` or `:era5_land`) passed to the
+CopernicusClimateDataStore download function, which determines the CDS product id
+and whether `product_type` is included in the request.
+"""
+cds_dataset_keyword(::Union{ERA5HourlySingleLevel, ERA5YearlySingleLevel, ERA5MonthlySingleLevel,
+                             ERA5HourlyPressureLevels, ERA5MonthlyPressureLevels}) = :era5
+cds_dataset_keyword(::Union{ERA5HourlyLand, ERA5MonthlyLand}) = :era5_land
```
xkykai: is this reachable?

xkykai (commented):
nice! consider adding to `docs/src/Metadata/metadata_overview.md` and `docs/src/Metadata/supported_variables.md` as well

aklocker42:
Good call — added `ERA5HourlyLand`/`ERA5MonthlyLand` to both `metadata_overview.md` (Land section table) and `supported_variables.md` (full variable list with units) in 5be3baf.

ext/NumericalEarthCopernicusClimateDataStoreExt.jl
aklocker42: Correct, not reachable — `ERA5LandDataset` has its own dedicated `Downloads.download` method below with a self-contained implementation, so these Union branches on the generic helpers never dispatch. Removed in 42c4063, with `hasmethod` tests added to catch this regressing.

ext/NumericalEarthCopernicusClimateDataStoreExt.jl
aklocker42: Agreed — removed the local duplicate. `CopernicusClimateDataStore.jl` v0.2.2 (released this cycle) does the same unwrap automatically inside `retrieve()`, and does it slightly more safely (warns on multi-file zips instead of silently picking the first). Bumped the compat bound to `0.2.2` to guarantee it's present.

ext/NumericalEarthCopernicusClimateDataStoreExt.jl
aklocker42: No — confirmed it was dead, the function copies whatever variable(s) exist in the source files regardless of the argument. Removed the parameter in 42c4063.

ext/NumericalEarthCopernicusClimateDataStoreExt.jl
aklocker42: A full year of hourly ERA5-Land data exceeds CDS's per-request cost limit, so hourly requests are split into one CDS request per calendar month and the resulting files are concatenated locally into the yearly file (`era5_land_year_batches` builds the batches, `concatenate_era5_nc` does the concatenation).

ext/NumericalEarthCopernicusClimateDataStoreExt.jl
aklocker42: Applied — switched both `era5_land_request` methods to the v2 `data_format`/`download_format` keys in 42c4063.

src/DataWrangling/ERA5/ERA5_land.jl
aklocker42: You're right, this is a real gap. It's inherited from the same pattern in `ERA5YearlySingleLevel`/`ERA5MonthlySingleLevel` (pre-existing, marked "for now" there too), not new to this PR, and today it already fails loudly rather than silently — `set!`/`retrieve_data` in `ERA5_field_time_series.jl` both `error()` when a date isn't found in the (possibly wrong) file. In 42c4063 I added an explicit guard in `build_filename` for Land datasets so the error is clearer and earlier, plus a test. A full fix (datewise filenames across all three yearly-file dataset families) is a bigger refactor I'd rather scope separately — happy to file a follow-up issue if that's preferred.

src/DataWrangling/ERA5/ERA5_field_time_series.jl
aklocker42: Same as the reply on the `ERA5_land.jl` thread — this is a real, pre-existing limitation shared with `ERA5YearlySingleLevel`/`ERA5MonthlySingleLevel`, not new here. `set!`/`retrieve_data` already fail loudly if a date isn't found in the (possibly wrong) file; 42c4063 adds an explicit guard at `build_filename` for Land datasets to fail earlier and more clearly. A symmetric fix across all three yearly-file dataset types would need datewise filenames — scoping that as a follow-up rather than doing it in this PR.

test/test_copernicus_climate_datastore.jl
aklocker42: Added an offline test for the guard above instead of a live multi-year download (avoids adding more real CDS calls to per-commit CI). Agreed on moving downloading tests out of per-commit CI generally — that's #440's scope, I'll leave it there rather than restructure CI in this PR.

src/DataWrangling/ERA5/ERA5_field_time_series.jl
xkykai: does this mean that multi-year ERA5Land ingestion wouldn't work for now? if that's the case what should one do in this case if they wish to have multi-year data

test/test_copernicus_climate_datastore.jl
xkykai: I think we can move downloading tests to `*_downloading.jl` files and then delete them for automatic commit ci triggers (see other datasets for example) in this PR

xkykai (commented):
thanks a lot! I am just curious as to what happens to multi-year downloads currently. why is it not advisable to deal with that in this PR directly?

src/DataWrangling/ERA5/ERA5_field_time_series.jl
aklocker42: Sorry — should have just fixed this properly on the first pass instead of adding a guard and punting to a follow-up. I went back and implemented the real fix.

**Root cause:** `build_filename` for the yearly-file ERA5 datasets (`ERA5HourlyLand`, `ERA5MonthlyLand`, `ERA5YearlySingleLevel`) collapsed a whole requested date range down to a single filename based on `dates[1]`'s year. That's fine within one year, but silently pointed at the wrong file for any range spanning a year boundary.

**Fix:** these `build_filename` methods now return a `DatewiseFilename` — one filename per requested date (heavily repeated within a year, since many dates share a file) — instead of collapsing to one string. This is existing, already-wired infrastructure (`Metadata`'s iteration/download path already dispatches on `DatewiseFilename` via `getfilename`), so no changes were needed on the download side at all — `Downloads.download` already iterates per-date and downloads each year's file as needed.

On the read side, `set!` for these `FieldTimeSeries` (in `ERA5_field_time_series.jl`) now groups consecutive same-year dates and opens each yearly file once, via a new `read_era5_yearly_series` helper, instead of assuming a single file. Added offline tests (synthetic NetCDF files spanning a year boundary) covering both the `build_filename` change and the multi-file read/stitch logic.

Net effect: multi-year ERA5-Land / ERA5YearlySingleLevel `FieldTimeSeries` requests now work correctly, reading from as many yearly files as the range spans.

While I was in there I also fixed `ERA5MonthlySingleLevel`, which had the same latent bug at month granularity (a multi-month request collapsed to the first month's file). That one turned out simpler: it's a monthly-means dataset (one file, one timestep, per month) that already goes through the generic per-file `set!` in `dataset_backend.jl`, which maps each time index to its own file — so the `build_filename` fix alone was sufficient there, no reader changes needed. Added the same style of offline test for it.

test/test_copernicus_climate_datastore.jl
aklocker42: Good idea in general, but I looked closely and `test_copernicus_climate_datastore.jl` doesn't actually make any live CDS API calls anywhere in the file — everything in it is offline (`hasmethod`/dispatch/dict-lookup checks, and now synthetic-NetCDF-based tests for the multi-year fix above). So there's nothing in there that needs to move to a `*_downloading.jl` file for CI-triggering purposes; it's already safe to run on every commit.

If you'd like, I can add a genuine live-download smoke test for ERA5-Land in a new `test_copernicus_climate_datastore_downloading.jl`, following the pattern of the other `*_downloading.jl` files — just didn't want to assume that's what you meant vs. flagging that the current file is already offline-only.

xkykai: yeah I think adding genuine downloading smoke test would be useful, just not for every commit trigger. cf https://github.com/NumericalEarth/NumericalEarth.jl/pull/440 and https://github.com/NumericalEarth/NumericalEarth.jl/issues/480 (feel free to add these datasets to the list as well!). Once https://github.com/NumericalEarth/NumericalEarth.jl/pull/440 I can open a new PR that puts the actual downloading test into regularly-triggered ci.

xkykai (approved):
lgtm! just add an actual downloading test (see https://github.com/NumericalEarth/NumericalEarth.jl/pull/490#discussion_r3707315299) and we can look to incorporate that into ci in the future. for now put something like https://github.com/NumericalEarth/NumericalEarth.jl/blob/0a31b427d03d82ab65e473cd296315a4ade84166/test/runtests.jl#L32 in `runtests.jl` for the the live downloading tests.

once you're confident feel free to merge

PR #493, opened 2026-08-04
Support Breeze 0.8

dkytezab:
I hope this is the correct compat syntax!


test/Project.toml
```diff
@@ -38,7 +38,7 @@ NumericalEarth = {path = ".."}
 
 [compat]
 ArchGDAL = "0.10"
-Breeze = "0.7.3"
+Breeze = "0.8"
```
giordano: The syntax is correct, but if no changes are required then you could simply add 0.8
```suggestion
Breeze = "0.7.3, 0.8"
```

giordano (approved):

PR #495, opened 2026-08-04
Try Reactant version 0.2.278, `--check-bounds=auto`

dkytezab:
actually, i think that we may just need to run the Reactant job with `--check-bounds=auto`


giordano (approved):
Didn't check the errors, but looks OK, since we do it elsewhere (and tests are passing)

dkytezab:
I think that the data downloading job is a bit broken unfortunately

xkykai (approved):
seems to work now!

PR #496, opened 2026-08-04
Release 0.6.1

dkytezab:


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "NumericalEarth"
 uuid = "904d977b-046a-4731-8b86-9235c0d1ef02"
 license = "MIT"
-version = "0.6.0"
+version = "0.7.0"
```
giordano: Wait, v0.6.0 -> v0.7.0 designates a breaking change, but is there any since v0.6.0?

xkykai (commented):
if there are breaking changes, would you mind writing them for the release? Here are the steps for a breaking change release in case that is useful:

1. Bump the minor version.

2. Merge, then write breaking change notes into the registration comment. e.g.

```
@JuliaRegistrator register

Release notes:

## Breaking changes

- Land energy flux field net_energy_flux renamed to surface_energy_flux, with its sign flipped to positive upward (out of the slab). Custom land energy closures and any code reading the field must update; legacy BucketHydrology/SlabEnergy/FractionalHumidity configurations are otherwise unchanged (#323).
```

3. Registrator opens a General PR, copying your notes into its body between  markers. It auto-merges after roughly 15 minutes if checks pass. 

5. TagBot does the rest.

Project.toml
dkytezab: Ah a change I thought was breaking is not on a second pass

giordano (approved):

xkykai (approved):

dkytezab:
@giordano i wonder if we could force this through... some other stuff is waiting on version release and this change shouldn't break anything

PR #499, opened 2026-08-05
Reject regional reads whose file does not cover the native grid

### LINKED ISSUE #498: Regional reads are off by one cell when the on-disk file spans the region rather than the padded native grid

xkykai:
A regional read placed the file's data at an offset clamped to `≥ 0` with an upper bound that
collapsed when the file was smaller than the grid, so a file spanning only the requested box —
rather than the native grid, which `restrict` widens by up to one native cell per side — had every
value written one cell off, with the trailing cells duplicating the file's edge. Fixes #498.

- `region_info` now resolves the offset from the file's coordinate labels and rejects a window that
  falls outside the file, naming both spans.
- Files that do cover the grid are unaffected: an exact window still gives offset `0`, an
  over-fetched file a positive offset, and the global-file seam continuation is unchanged.
- `GHSL`'s writer warps to the requested box, so its reads now raise until it materializes the file
  on `native_grid(metadatum)`.

```julia
λc = φc = collect(0.5:1.0:3.5)                                     # file: 1° cells over (0, 4)
grid = LatitudeLongitudeGrid(CPU(); size = (6, 6), longitude = (-1, 5), latitude = (-1, 5),
                             topology = (Bounded, Bounded, Flat))  # centers -0.5 … 4.5
metadatum = Metadatum(:bottom_height; dataset = ETOPO2022(),
                      region = BoundingBox(longitude = (0, 4), latitude = (0, 4)))
set_region_data!(Field{Center, Center, Nothing}(grid), zeros(4, 4, 1), λc, φc, metadatum)
# ERROR: ArgumentError: the file does not cover the native grid in longitude: the read needs
# 6 cells from -0.5 to 4.5, but the file labels 4 cells from 0.5 to 3.5. …
```

### DIFF
--- a/src/DataWrangling/set_region_data.jl
+++ b/src/DataWrangling/set_region_data.jl
@@ -17,13 +17,16 @@ end
 # promoted to Float64; `tolerance` absorbs the promotion drift, capped at a quarter cell because
 # `eps(Float32) * |h|` alone exceeds half a cell on arc-second axes near 180°. On an axis whose
 # labels sit half a cell from the grid's nodes (regional ERA5 latitude), the tie resolves to the
-# label north/east of the node.
+# label north/east of the node. A node more than half a cell before the first label extrapolates to
+# an index below 1, marking a file that begins east/north of where the read starts.
 function file_cell_index(h, hc)
     Nh = length(hc)
     Nh > 1 || return 1
     j = clamp(searchsortedlast(hc, h), 1, Nh - 1)
-    tolerance = min(8 * eps(Float32) * max(1, abs(h)), abs(hc[j+1] - hc[j]) / 4)
-    return clamp(searchsortedfirst(hc, h - tolerance), 1, Nh)
+    Δ = hc[j+1] - hc[j]
+    tolerance = min(8 * eps(Float32) * max(1, abs(h)), abs(Δ) / 4)
+    h ≥ hc[1] && return clamp(searchsortedfirst(hc, h - tolerance), 1, Nh)
+    return 1 + ceil(Int, (h - tolerance - hc[1]) / Δ)
 end
 
 # Periodic only when the restricted span equals the full native span.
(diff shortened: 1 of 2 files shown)


xkykai:
close this for now, will be absorbed into https://github.com/NumericalEarth/NumericalEarth.jl/pull/465

xkykai:
Folded into #465, which carries the seam continuation this guard depends on. Closing in favor of that PR.

PR #501, opened 2026-08-06
Write the GHSL regional file on its native grid

### LINKED ISSUE #500: GHSL regional files are written on the requested box, not the native grid, so their values land off their coordinates
`ghsl_tiles_to_netcdf` writes its regional file on the box the user asked for, while the read locates that file's cells on GHSL's declared global lattice. The two disagree by a fraction of a cell in origin and by whole cells in count, so every GHSL value is placed off its own coordinates.

The writer anchors on the request (`ext/NumericalEarthArchGDALExt/ghsl.jl`):

```julia
Δ = resolution_m / 111320
west, east = region.longitude
"-te", string(west), string(south), string(east), string(north),
"-tr", string(Δ), string(Δ),
```

so the first cell center is `west + Δ/2`, wherever `west` happens to fall. `native_grid` instead brackets the region onto the lattice `Base.size` declares (`Nx = round(360/Δ)`), padding up to one cell per side. The cell *size* agrees exactly — `360/400752` is bit-identical to `100/111320` — so this is purely origin and count.

MWE, offline (no tiles are downloaded; this compares the geometry the writer produces against the grid the read expects). The window is the one in `examples/ghsl_urban_roughness.jl`:

```julia
using NumericalEarth.DataWrangling: BoundingBox, native_grid
using NumericalEarth.DataWrangling.GHSL: GHSBuiltH
using Oceananigans.Grids: λnodes, φnodes

region = BoundingBox(longitude = (-0.11, -0.07), latitude = (51.505, 51.525))
metadatum = Metadatum(:building_height; dataset = GHSBuiltH(), region)

grid = native_grid(metadatum)          # where the read expects the file's cells to sit
λ, φ = λnodes(grid, Center()), φnodes(grid, Center())

Δ = 100 / 111320                       # the writer's pixel size
file_λ₁ = region.longitude[1] + Δ / 2  # its first cell center, straight from `-te west …`
file_Nx = round(Int, (region.longitude[2] - region.longitude[1]) / Δ)

(file_λ₁ - λ[1]) / Δ      # 0.548 — the file starts half a cell east of the grid
length(λ) - file_Nx       # 1     — and is one cell short
```

Both axes, for that window:

```
grid lon:  46 cells, first -0.11004312     grid lat:  25 cells, first 51.50421906
file lon:  45 cells, first -0.10955084     file lat:  22 cells, first 51.50568630
     offset 0.548 cells, short 1                offset 1.63 cells, short 3
```

At 100 m cells that is a displacement of 55 m east and 163 m north for the example's own scene. The shortfall then makes it worse rather than better: `region_info` clamps the file offset to `≥ 0` with an upper bound that collapses when the file is smaller than the grid (#498), so the resolved offset is discarded, the data is written from index 1, and the trailing rows and columns repeat the file's edge.

Nothing reports any of this. `test_ghsl.jl` passes because it never reads a file, and `test_ghsl_downloading` is excluded from CI in `test/runtests.jl`.

Fix: materialize the file on `native_grid(metadatum)`, the way `CopernicusDEM.zarr_to_netcdf` does — warp with `-te` from the grid's own domain and `-ts Nx Ny`, so the written `lon`/`lat` are the grid's centers and the read resolves an offset of 0.

The numbers above are measured from a downloaded file. The latitude figures are worse than
`west + Δ/2` alone predicts (1.369 cells): under `-tr` GDAL also adjusts the extent when it is
not an integer number of pixels, which moves the origin as well as the count.

xkykai:
`ghsl_tiles_to_netcdf` warped onto the requested bounding box, but the read path locates the file's cells on `native_grid(metadatum)` — same cell size, different origin and count — so the file landed offset from the grid and short of it, with the trailing rows and columns repeating the edge.

- the warp extent, cell count and coordinates now come from `native_grid`; `-ts` pins the count so GDAL cannot round the extent away
- tiles are selected on that window, since the grid can reach a cell past the request and across a tile boundary that cell would come back empty

Measured for `BoundingBox(longitude = (-0.11, -0.07), latitude = (51.505, 51.525))` with `GHSBuiltH()`:

```
grid   lon:  46 cells, first -0.11004312   lat:  25 cells, first 51.50422
before lon:  45 cells, first -0.10955084   lat:  22 cells, first 51.50569
       offset 0.548 cells, short 1              offset 1.63 cells, short 3
after  lon:  46 cells, first -0.11004312   lat:  25 cells, first 51.50422
       offset 0 cells, short 0                  offset 0 cells, short 0
```

Closes #500.


src/DataWrangling/GHSL/GHSL.jl
```diff
@@ -389,6 +390,40 @@ end
 ##### Download (regional Mollweide tiles → reprojected NetCDF via the ArchGDAL ext)
 #####
 
+"""
+    ghsl_regional_raster(metadatum)
+
+Geometry of the regional raster materialized for `metadatum`, taken from `native_grid`
```
glwagner, line 396: define "ghsl"? (like, wut)

src/DataWrangling/GHSL/GHSL.jl
glwagner, line 27: i guess it's defined here

glwagner (approved):

src/DataWrangling/GHSL/GHSL.jl
xkykai: so should I also include "Global Human Settlement Layer" somewhere in the description of this function?

xkykai: I'll take it as it is ok since I defined it near the top

PR #452, opened 2026-07-17
Use bottom_height_field for immersed bathymetry plots

xkykai:
Fixes #451.

Since Oceananigans 0.110.6 ([#5737](https://github.com/CliMA/Oceananigans.jl/pull/5737)), `grid.immersed_boundary.bottom_height` is a plain `OffsetArray`, not a `Field`, so `heatmap` of it fails — Makie receives a 3-D array. Oceananigans exports `bottom_height_field(grid)` for exactly this: it wraps the array back into a `Field` (zero-copy) that plots via the `Field` Makie recipe.

Fixes both affected examples:
- `near_global_ocean_simulation.jl` (fails the docs build)
- `generate_surface_fluxes.jl`

```diff
-heatmap(grid.immersed_boundary.bottom_height)
+heatmap(bottom_height_field(grid))
```

`near_global` is `build_always=false`; the `build all examples` label is set so CI exercises it. (`generate_surface_fluxes` is not in the docs build, so its change is not CI-covered.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #503, opened 2026-08-07
Build docs on a schedule (dev + pending releases) instead of on every push

glwagner:
## Why

The docs workflow built the **full example set** (hours on the single `gpu-3` runner) on **every push to `main`** (several per day), with `cancel-in-progress: true` killing each in-progress build when the next push landed. Result over the last 40 `main` runs: **0 success, 27 cancelled, 12 failure** — the dev docs haven't had a real content deploy since early July and are missing every example added since (e.g. the ERA5 downscaling example 404s). Even the **v0.6.1 release docs never deployed** — its tag build met the same fate.

## What

Three changes, one theme: docs builds happen on a schedule, run to completion, and deploy.

### 1. Don't cancel non-PR builds

```yaml
cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

Scheduled/dispatched builds queue and run to completion; PR builds still cancel superseded runs.

### 2. Build every other night instead of on pushes to `main`

- `schedule: cron '0 6 */2 * *'` — 06:00 UTC on odd days of the month. Dev docs are at most ~2 days stale, and the runner sees at most one full docs build every other day instead of one per merge.
- **Pushes to `main` no longer trigger docs builds.** PR docs builds (light set by default, full set via the `build all examples` label) remain the pre-merge check.
- **`workflow_dispatch` gains a `deploy` input**: manually trigger a full build + deploy on demand (e.g. right after merging a docs fix) instead of waiting for the next scheduled run.

### 3. Releases ride along with the scheduled build (tag trigger removed)

A cheap `plan` job (ubuntu-latest) decides what each run builds: always the triggering ref, **plus the latest release whenever its docs are missing from NumericalEarthDocumentation's gh-pages** (checked via the GitHub contents API). Build and deploy are matrixed over those refs with `max-parallel: 1` (one GPU); the release leg checks out the tag and points `GITHUB_REF` at `refs/tags/vX.Y.Z` for `deploydocs`, so Documenter files it as a versioned release. `deploydocs` accepts `schedule`/`workflow_dispatch` events natively — no changes to `docs/deploy.jl`.

This is self-healing: the check ran against the live deploy repo and found **v0.6.1's docs are indeed missing**, so the first scheduled run after merge will build `dev` *and* `v0.6.1`.

## Notes

- Release docs now appear at the next scheduled run after tagging (≤2 days) or on demand via dispatch, rather than immediately.
- Only the *latest* release is backfilled; if two releases land between scheduled runs, the older one would need a manual dispatch (rare enough not to engineer for).
- This unblocks deployment but doesn't by itself fix builds that *fail* on an example. The last completed `main` build (2026-07-20) died on `near_global_ocean_simulation.jl` with a Makie `Heatmap`/`CellGrid` conversion error; that line was since fixed by #452 (`bottom_height_field`), which this PR's labeled full-examples build should confirm.


giordano:
But this will make the queue even longer, no?  Feels like the only solution is to make the docs build shorter.

> Over the last 40 `main` runs of the docs workflow: **0 success, 27 cancelled, 12 failure**.

image

glwagner:
the 27 cancelled runs got cancelled I think because of a new commit coming into `main`.

In other words, if a commit always lands before the previous build succeeds, the docs will never be updated. What is the solution?

glwagner:
Hmm, maybe a better way to do this is to build the docs on a schedule (and not at all on merge to main).

giordano:
> What is the solution?

As I said above, I believe the docs build is way too expensive.

> Hmm, maybe a better way to do this is to build the docs on a schedule (and not at all on merge to main).

That may shorten the queue if the average of commits on `main` per day is larger than 1, which seems to be the case at least for the last 20 days.  But we may still want to build releases

giordano (approved):

glwagner:
> > What is the solution?
> 
> As I said above, I believe the docs build is way too expensive.

I appreciate your comment, but this is tangential to the point of this PR. The point of this PR is to figure out how to get the docs to build given the current situation. Making the docs build less expensive is a larger and more disruptive change.

If we come to the conclusion that it _isn't possible_ to have docs that are this expensive, that's another story.

I feel this cannot possibly be correct. We can also deploy more computational resources for the docs. So "expensive" is a relative judgement.

glwagner:
Update: tartarus (the gpu-3 docs runner) is offline — no docs build has executed a step since ~Aug 6 (jobs queue 24 h, then GitHub's queue limit cancels them). Two interim changes added to this PR:

- **Docs build moved to `aws-linux-nvidia-gpu-l4`** with the same container pattern as the CI GPU tests (workspace-EBS depot/TMPDIR, host CUDA-toolkit cleanup); **deploy moved to `ubuntu-latest`** (only needs Documenter + the artifact). Revert to tartarus is a small, isolated diff once the machine is back.
- **Near-global ocean example disabled** in `docs/make.jl` — the heaviest example (¼°, ~35M cells), dropped to keep build time and memory in check on the ephemeral cold-cache L4.

The labeled full-examples build on this PR should now actually execute and tell us whether the remaining example set (including the ERA5 downscaling example) builds end to end.

glwagner:
The previous L4 run got past the ECCO-disabled slots and uncovered a real source bug: the Veros example aborts with `FieldError: type NamedTuple has no field η` in `assemble_net_ocean_fluxes.jl` — #244 added freshwater-flux containers to the Oceananigans and slab-ocean `net_fluxes` but missed the Veros extension. Fix: #506. The Veros example is disabled here until that merges (it has never built in a completed docs run, so this was the first time the path was exercised).

Current temporarily-disabled examples on this branch, and their re-enable conditions:
- near-global — heaviest example; L4 capacity judgment call (revisit when tartarus returns)
- one-degree, global-climate — `ecco.jpl.nasa.gov/drive` HTTP 503 outage (monitoring for recovery)
- Veros — blocked on #506

glwagner:
Consolidated per discussion — the example/extension fixes now live on this PR instead of separate ones (#506, #507 closed):

- **Veros flux-assembly fix**: `net_fluxes(::VerosOceanSimulation)` gains the `η` and `freshwater_heat_content` containers missed in #244; the Veros example is re-enabled.
- **Four-oceans example fix**: model-level state is accessed through `sim.model` (`.dynamics`, `.velocities`, `.microphysical_fields`, Breeze diagnostics), matching `breeze_over_slab_land`.
- **`make.jl` no longer aborts at the first failed example**: each example builds in a try/catch, the site is built from whatever succeeded, and the build fails at the very end listing every broken example. The docs artifact uploads even on failure. This ends the one-failure-per-iteration discovery grind that consumed four build cycles this week.

Still disabled: near-global (L4 capacity call) and the two ECCO-initialized examples (upstream 503 outage).

PR #507, opened 2026-08-12
Fix model access in the four-oceans example

glwagner:
## Why

`atmosphere_simulation` returns a `Simulation` (since #433), but `breeze_over_four_oceans.jl` still reads model-level state directly off the return value: `.dynamics.reference_state`, `.velocities`, `.microphysical_fields.qˡ`, and `liquid_ice_potential_temperature(...)` (whose Breeze methods accept only `AtmosphereModel`). The example dies at the first of these:

```
FieldError: type Oceananigans.Simulations.Simulation has no field `dynamics`
```

aborting the docs build. Caught by the docs build on #503's L4 runner ([failed run](https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/31597314603)); like the Veros bug (#506), it went unnoticed because no full-examples docs build has completed since July.

## What

Access the model through the simulation (`sim.model.…`) at each site, matching the `breeze_over_slab_land` pattern. `set!(sim, …)` and the `ComponentInterfaces`/`AtmosphereOceanModel` calls already accept the `Simulation` and are unchanged.

### DIFF
--- a/examples/breeze_over_four_oceans.jl
+++ b/examples/breeze_over_four_oceans.jl
@@ -64,7 +64,7 @@ nh_ocean_atmos         = atmosphere_simulation(grid; potential_temperature=θᵃ
 # A background zonal wind `U₀` provides a nonzero wind speed for the
 # similarity theory flux computation.
 
-reference_state = slab_ocean_atmos.dynamics.reference_state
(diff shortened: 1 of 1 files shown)


glwagner:
Folded into #503 (commit cherry-picked there) — keeping the docs-recovery work in one PR.

PR #517, opened 2026-08-14
Nested model: sample a terrain-following parent at its reference node

ewquon:
## The bug

`interpolate` locates the vertical of a query through the source grid's 1-D `znodes`, and
`znodes(grid, ℓz) = rnodes(grid, ℓz)` for every grid — so on a terrain-following grid the lookup table
is the **reference** column `r`. But `node`/`_node` on such a grid returns the terrain-deformed
**physical** height — for the Gal-Chen & Somerville (1975) coordinate Breeze uses,
`z = r + h(x, y)·b(r)`, with `b(r) = 1 − r/z_top` under the default `LinearDecay` — and that is what
the query side passes in: `getbc` for an `Interpolated` boundary condition, and `interpolate!`'s kernel.

Asking for a value at physical height and finding it in a table of reference heights is neither
convention: every sample is displaced by `h·b(r)` — the full terrain height at the ground, decaying to
zero at the lid. Identical parent and child terrain does not cancel it, because the mismatch is between
the query convention and the lookup table, not between the grids.

This is invisible in the ERA5 → LAM path, because a `PressureLevelGrid`'s `znodes` *are* physical
(geopotential) heights. It bites wherever the source grid is terrain-following.

## Where it shows up today

`initialize_nested_child!` shifts the interpolated momentum onto the child's own staggered locations
with `interpolate!(u::XFaceField, Field(ρu / ρᵈ))`. Source and target are both the child's
terrain-following grid, so the intended Center → Face horizontal shift also picks up the vertical
displacement — an error in the initial `u`, `v` near the ground, where the shear is strongest.

## The fix

`query_node(source_grid, i, j, k, grid, ℓx, ℓy, ℓz)` is the coordinate at which a child samples a
parent. It defaults to the child's physical `node`; the Breeze extension defines the
`TerrainFollowingGrid` method returning `(ξnode, ηnode, rnode)`, so both sides are read in the same
vertical coordinate. `getbc` and the new `interpolate_from_parent!` go through it, and the two velocity
regrids in `initialize_nested_child!` switch to the latter. The 2-D elevation regrid keeps
`interpolate!` — a `(Center, Center, Nothing)` field has no vertical lookup.

## Why terrain-following (σ-to-σ) sampling

σ is the generic name for a terrain-following vertical coordinate — classically the pressure-based
σ = p/pₛ, WRF's mass-based η, or the height-based `r` here — so "σ-to-σ" means sampling the parent at
the same value of that coordinate. Near the ground that is height above ground; aloft, where `b(r) → 0`,
the levels flatten and it becomes physical height.

Sampling this way is also the convention nesting wants, for three reasons beyond internal consistency.

**It cannot sample below the parent's ground.** Where a coarse parent smooths a ridge into a ramp and
the child resolves the peaks and valleys under it, a child valley floor sits *underground* in the
parent. Querying it at its own physical height finds no valid data: the query clamps to the parent's
lowest level, or reads whatever is stored below the surface. Child peaks are fine — they are above the
parent's ground — so the failure is asymmetric, and valleys are the common case in a refined nest.
Sampling at constant `r` maps every child level to a valid parent level by construction. This repo
already carries the evidence that sampling a terrain-bearing source by physical height is fragile:
`clip_subsurface!` and `column_fractional_z_index` exist to keep ERA5's non-physical sub-surface levels
out of exactly this kind of query.

**It follows the coordinate the physics does.** Constant `r` is height above ground up to a column
compression, `AGL = r·(1 − h/z_top)` — 2 % over 400 m of terrain with a 19.5 km lid — so it is
terrain-following exactly where height above ground governs the flow (surface layer, boundary layer,
flow over terrain). Aloft, where height above sea level governs it (free troposphere, the jet,
isentropic structure), `b(r) → 0` and the levels are already flat.

**It is exact when the levels are shared.** Nests that share a reference discretization — the normal
case for a telescope — map level `k` to level `k`, so only the horizontal is interpolated and there is
no vertical interpolation error anywhere, including in the boundary zone.

## Test

A new testset asserts the round trip on a 400 m plateau: the reference-node query reproduces the source
exactly, the physical-node query is displaced by ≈ h, and `interpolate_from_parent!` reproduces a
terrain-following field exactly on a coincident grid. The existing nested-simulation integration tests
cover the changed initialization call sites.

## Complementary Breeze fix

The general fix belongs in Breeze: a `_fractional_indices(::NTuple{3}, ::TerrainFollowingGrid, …)` that
inverts the column mapping, mirroring what this package already does for `PressureLevelGrid` via
`column_fractional_z_index` (#241). The mapping is column-dependent, so it cannot be expressed in the
1-D `znodes` table `interpolate` consults, and cannot be fixed on the lookup side from here. I found no
Breeze issue or PR covering it, and no such method anywhere in Breeze's history.

It would serve the case this PR does not: a terrain-following parent read by a child that has no
terrain-following coordinate of its own. Nesting down to a flat-bottom, uniform-grid LES is the
concrete example. Its levels are fixed heights, not σ levels, so there is no value of `r` on the child
side to match and "sample at the same level" has no referent; the well-posed question there is the
parent's value at a physical height, which needs the parent's column mapping inverted. The displacement
scales with the parent's terrain height under the nest, so this is harmless over lowland and wrong over
an elevated plateau — and it is quiet either way, because `query_node` dispatches on the source grid, so
such a child falls back to its own physical node and keeps the mismatch. The `query_node` docstring
records the assumption. Nests that share the parent's terrain-following coordinate — the telescoping
case — are covered here.

For reference, the inversion is closed form under the default `LinearDecay`,
`r = (z − h)/(1 − h/z_top)`; `TwoLevelDecay` (SLEVE) would need iteration.

### DIFF
--- a/ext/NumericalEarthBreezeExt/breeze_nested_atmosphere.jl
+++ b/ext/NumericalEarthBreezeExt/breeze_nested_atmosphere.jl
@@ -15,7 +15,7 @@ using NumericalEarth.Atmospheres: PrescribedAtmosphere
 using NumericalEarth.DataWrangling: default_download_directory, default_horizontal_padding,
                                     matching_single_level_dataset
 using NumericalEarth.NestedModels: NestedModel, parent_boundary_conditions, parent_forcings,
-                                   blend_parent_terrain!
+                                   blend_parent_terrain!, interpolate_from_parent!
 using Oceananigans: Oceananigans, WENO
 using Oceananigans.Architectures: architecture
 using Oceananigans.BoundaryConditions: ValueBoundaryCondition, NormalFlowBoundaryCondition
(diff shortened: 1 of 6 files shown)


ewquon:
Closing. `main` has changed underneath this (#454, #604), and it would need a third rebase for code that nothing on `main` uses: every parent is a `PressureLevelGrid`, so `query_node` always takes the default. The displacement in the initial `u`, `v` that this also addressed now comes from a different path, the Center→Face regrid inside `set!`, and is fixed separately in #706. Sampling a terrain-following parent is tracked in #705, with this PR as the prototype for one of the options.

PR #523, opened 2026-08-19
Per-cell land roughness and displacement via Field-valued closure parameters

xkykai:
The roughness lengths and zero-plane displacement of `SimilarityTheoryFluxes` now accept per-cell `Field`s alongside the existing `Number` and formulation options. The land flux kernel collapses them to each cell's values (through `state2dindex`) before the index-free Monin–Obukhov solve, so data-driven aerodynamic fields — `urban_roughness` output, canopy roughness — reach the solver.

```julia
fluxes = SimilarityTheoryFluxes(momentum_roughness_length    = ℓᵐ,        # Field
                                temperature_roughness_length = 0.1 * ℓᵐ,  # lazy per-cell operation
                                water_vapor_roughness_length = 0.1 * ℓᵐ,
                                zero_plane_displacement      = d,         # Field
                                stability_functions          = atmosphere_land_stability_functions(Float64))

interface = atmosphere_land_interface(grid, atmosphere, land; fluxes)
```

**Removed API:** `LandRoughnessLength` and `LandZeroPlaneDisplacement` — markers that told the solver to fetch per-cell values supplied separately. With `Field`s accepted directly, both halves of that indirection go away, including the `atmosphere_land_surface_properties` hooks and the `interior_properties` argument of `iterate_interface_fluxes`. `LandRoughnessLength`'s `multiplier` is replaced by a lazy operation (above, indexed per cell like a `Field`); its `minimum_roughness_length` floor has no solver-side analog — clamp where the roughness field is produced.

Tested end-to-end in `test/test_slab_land.jl`: a two-cell neutral coupled model with per-cell `(ℓᵐ, d)` fields reproduces `u★ = ϰ ΔU / log((h − d)/ℓᵐ)` cell by cell.


src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
```diff
@@ -252,54 +252,28 @@ end
...
-        ℓ.minimum_roughness_length
-    end
-
-    return max(ℓ.multiplier * candidate, ℓ.minimum_roughness_length)
-end
+# Localize the flux closure to cell (i, j) before the index-free MOST iteration:
+# `Field`-valued roughness and displacement slots collapse to their per-cell values,
+# while `Number`s and formulations (e.g. `MomentumRoughnessLength`) pass through.
+@inline local_parameter(ℓ, i, j) = ℓ
+@inline local_parameter(ℓ::AbstractArray, i, j) = @inbounds ℓ[i, j, 1]
```
glwagner: there's a lot of helper functions like this, eg `stateindex`. Should we just reuse `stateindex` or one of the other helpers instead of creating a new one?

src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
```diff
@@ -253,10 +253,12 @@ end
...
 # Localize the flux closure to cell (i, j) before the index-free MOST iteration:
-# `Field`-valued roughness and displacement slots collapse to their per-cell values,
-# while `Number`s and formulations (e.g. `MomentumRoughnessLength`) pass through.
+# `Field`-valued roughness and displacement slots collapse to their per-cell values
+# via `state2dindex`, while `Number`s and formulations (e.g. `MomentumRoughnessLength`)
+# pass through. Not a `stateindex` method: there a callable means a coordinate function
+# `a(λ, φ, z, time)`, whereas a callable closure slot is a formulation of `u★`.
 @inline local_parameter(ℓ, i, j) = ℓ
-@inline local_parameter(ℓ::AbstractArray, i, j) = @inbounds ℓ[i, j, 1]
+@inline local_parameter(ℓ::AbstractArray, i, j) = state2dindex(ℓ, i, j)
```
glwagner: just delete `local_parameter` all together and use `state2dindex` at the callsite

xkykai: done

src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
```diff
@@ -179,19 +197,23 @@ function SimilarityTheoryFluxes(FT::DataType = Oceananigans.defaults.FloatType;
                                 momentum_roughness_length = MomentumRoughnessLength(FT),
                                 temperature_roughness_length = ScalarRoughnessLength(FT),
                                 water_vapor_roughness_length = ScalarRoughnessLength(FT),
+                                minimum_roughness_length = 1e-5,
```
simone-silvestri: does this apply to all the similarity fluxes? Why did you choose this value or why is this needed?

src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
```diff
@@ -179,19 +197,23 @@ function SimilarityTheoryFluxes(FT::DataType = Oceananigans.defaults.FloatType;
...
                                 solver_stop_criteria = nothing,
                                 solver_tolerance = 1e-8,
                                 solver_maxiter = 100)
 
-    roughness_lengths = SimilarityScales(momentum_roughness_length,
-                                         temperature_roughness_length,
-                                         water_vapor_roughness_length)
+    # Convert scalar slots to `FT` so the kernel does not promote the whole MOST chain to
+    # a wider type (nor hit an `InexactError` flooring an integer roughness). Fields carry
+    # their own eltype and formulations were already built with `FT`.
```
simone-silvestri:
```suggestion
```

src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
```diff
@@ -252,73 +275,310 @@ end
...
+# `stateindex`'s fallback.
+@inline local_flux_formulation(flux_formulation, i, j) = flux_formulation
+@inline local_flux_formulation(flux_formulation, i, j, zᵃᵗ) = local_flux_formulation(flux_formulation, i, j)
+
+# "The similarity profile can evaluate this value": `log(Δh / ℓ)` needs ℓ > 0, and a
+# displacement raises the surface so d ≥ 0. `validate_flux_formulation` (setup),
+# `local_*` below (kernel) and `fill_aerodynamic_roughness_gaps!` (Lands) all decide the
+# same contract, so they share these two predicates rather than restating them.
+@inline evaluable_roughness_length(ℓ) = isfinite(ℓ) & (ℓ > 0)
+@inline evaluable_zero_plane_displacement(d) = isfinite(d) & (d ≥ 0)
```
simone-silvestri: All these guards, and the ones below, are much more easily solved by doing
```julia
@inline function similarity_profile(::LogarithmicSimilarityProfile, stability_function, h, ℓ, L)
    Φn = log(h / ℓ)
    ζ  = h / L
    ψh = stability_profile(stability_function, ζ)
    ψℓ = stability_profile(stability_function, ℓ / L)
    Φm = Φn / 1000 # stability correction cannot overwhelm the log profile
    return max(Φn - ψh + ψℓ, Φm)
end

@inline function similarity_profile(::COARELogarithmicSimilarityProfile, stability_function, h, ℓ, L)
    Φn = log(h / ℓ)
    ζ  = h / L
    ψh = stability_profile(stability_function, ζ)
    Φm = Φn / 1000 # stability correction cannot overwhelm the log profile
    return max(Φn - ψh, Φm)
end
```

src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
```diff
@@ -12,20 +12,18 @@ struct SimilarityTheoryFluxes{FT, UF, R, D, B, S, SV}
...
-                           adapt(to, fluxes.roughness_lengths),
-                           adapt(to, fluxes.zero_plane_displacement),
-                           adapt(to, fluxes.similarity_form),
-                           adapt(to, fluxes.solver_stop_criteria))
+# `local_flux_formulation` rebuilds this struct positionally inside a kernel, where a
+# name-keyed rebuild would not be type-stable, so a reordered or inserted field would
+# silently mis-wire the closure rather than fail to compile: every slot is a free type
+# parameter. `test_slab_land.jl` pins the field order. The host-side `adapt_structure`
+# needs no hand-written list.
+Adapt.@adapt_structure SimilarityTheoryFluxes
```
simone-silvestri, line 20: what does this do?

src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
```diff
@@ -160,12 +159,31 @@ Keyword Arguments
 - `stability_functions`: The stability functions. Default: `default_stability_functions(FT)` that follow the
                          formulation of [edson2013exchange](@citet).
 - `roughness_lengths`: The roughness lengths used to calculate the characteristic scales for momentum, temperature and
-                       water vapor. Default: `default_roughness_lengths(FT)`, formulation taken from [edson2013exchange](@citet).
+                       water vapor. Each may be a formulation, a `Number`, or — at the
+                       atmosphere--land interface only — a `Field{Center, Center, Nothing}` of
+                       per-cell values, strictly positive and finite everywhere (see
+                       [`atmosphere_land_interface`](@ref)).
+                       Default: `default_roughness_lengths(FT)`, formulation taken from [edson2013exchange](@citet).
+- `minimum_roughness_length`: Floor [m] applied to a per-cell roughness length that the
```
simone-silvestri: this might have a very large impact....

src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
```diff
@@ -252,73 +275,310 @@ end
...
+@inline evaluable_roughness_length(ℓ) = isfinite(ℓ) & (ℓ > 0)
+@inline evaluable_zero_plane_displacement(d) = isfinite(d) & (d ≥ 0)
+
+# Only an `AbstractArray` slot can change after `validate_flux_formulation` has run. Its
+# localized value falls back in the kernel when it is no longer evaluable or no longer
+# clears the displaced profile height. Numbers and formulations pass through untouched.
+@inline local_roughness_length(ℓ, ℓmin, i, j) = ℓ
+
+@inline function local_roughness_length(ℓ::AbstractArray, ℓmin, i, j)
+    ℓᵢⱼ = state2dindex(ℓ, i, j)
```
simone-silvestri: don't we have a `stateindex` function?

src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
```diff
@@ -252,73 +275,310 @@ end
...
+    end
+
+    return nothing
+end
+
+reject_per_cell_slots(flux_formulation, interface) = nothing
+
+# The four slots that may carry per-cell values, each with the predicate its values must
+# satisfy and the requirement to quote when they do not. `reject_per_cell_slots` and
+# `validate_flux_formulation` both walk this table, so adding a slot is one edit.
```
simone-silvestri: `per_cell_slots` is not really a terminology we use anywhere else no?

simone-silvestri:
The title of the PR says that this PR introduces `Field` roughness lengths and displacement. I would expect maximum 20 lines change for such an objective. What justifies changing 300 lines in the `SimilarityTheoryFluxes` machinery?

xkykai:
> The title of the PR says that this PR introduces `Field` roughness lengths and displacement. I would expect maximum 20 lines change for such an objective. What justifies changing 300 lines in the `SimilarityTheoryFluxes` machinery?

I think this PR has suffered from unintended scope creep, because original implementation of `Field` for roughness and displacement exposes bugs in the urban roughness closure when plugged into MOST. @kaiyuan-cheng do you mind filing issues you found with the roughness fields/closure, and I can simplify this down and fix other issues that emerge from this in separate PRs? This would mean reverting some of the changes you put in here I think, but we can relocate them into other PRs.

src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
kaiyuan-cheng: It applied to every `SimilarityTheoryFluxes`, and the value was not derived from anything physical: it was a last-resort floor that existed only because a `Field` slot can be mutated after the interface has been validated. Removed in 688d6c25 — the floor, and the clearance check it belonged to, are now tracked in #582.

src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
kaiyuan-cheng: Done, removed in 688d6c25. Converting numeric roughness slots to the closure's float type is tracked separately in #584.

src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
kaiyuan-cheng: The guards are out of this PR (688d6c25). Your `max(Φn - ψh + ψℓ, Φn/1000)` version is recorded in #582, which is where `displaced_profile_height` gets fixed.

One caveat for that PR: flooring the profile keeps the transfer coefficients finite, but it does not cover `Δh - d ≤ 0`, which is the case that motivated the clearance check — `MorphometricRoughness` under `VariableHeight` returns displacements above the default 10 m surface-layer height, and there `log((Δh - d) / ℓ)` is not just small, it is undefined.

src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
kaiyuan-cheng: It generates the `adapt_structure` method that was spelled out above it: rebuild the struct with `adapt` applied to every field, which is what moves a `Field`-valued roughness or displacement onto the device when the closure is passed to a GPU kernel. The macro version cannot drift out of sync with the field list. Added a comment saying so in 35b9a1de.

src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
kaiyuan-cheng: Agreed, and that is why it is gone (688d6c25) — a floor applied inside every `SimilarityTheoryFluxes` was far too blunt for what it was guarding.

With that and the clearance work stripped out, nothing left in this PR changes a value an existing closure computes: `Number` and formulation slots pass through localization untouched, and the only new behavior is reading a `Field` slot per cell plus setup-time validation of such a slot.

src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
kaiyuan-cheng: Yes — `state2dindex(a, i, j) = stateindex(a, i, j, 1)` (`InterfaceComputations.jl:122`), the 2-D wrapper the flux kernels already use for `surface_layer_height` and `h_bℓ`. Localization calls it directly now; the `local_roughness_length` / `local_zero_plane_displacement` wrappers this comment was pointing at are gone.

src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
kaiyuan-cheng: Renamed in 35b9a1de: `reject_per_cell_slots` → `reject_field_valued_slots` and `is_per_cell` → `is_field_valued`. I kept `slot`, which is used elsewhere in this file (e.g. the `mesoscale` slot of `SubgridVelocities`), and let the name say what is special about these ones: they may hold a `Field`.

kaiyuan-cheng:
Looks like #587 is blocking this PR.

xkykai:
> The title of the PR says that this PR introduces `Field` roughness lengths and displacement. I would expect maximum 20 lines change for such an objective. What justifies changing 300 lines in the `SimilarityTheoryFluxes` machinery?

I've reworked this PR, ready for another round of reviews

src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
xkykai: removed

src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
xkykai: used `stateindex`

src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
xkykai: removed "slots", using roughness lengths and zero-plane displacement now

src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
xkykai: this is no longer done

xkykai:
> Looks like #587 is blocking this PR.

#587 is now merged

xkykai:
@simone-silvestri let me know what you think of this now!

kaiyuan-cheng (approved):

src/EarthSystemModels/InterfaceComputations/InterfaceComputations.jl
```diff
@@ -115,12 +114,13 @@ function interface_kernel_parameters(grid)
...
-# (e.g. a prescribed measurement height or the 600 m BL-height fallback) passes through,
-# and a 2-D `Field` (Breeze's per-column surface- or boundary-layer height) is read at
-# column `(i, j)`. Used by the atmosphere–surface flux kernels to consume
-# `surface_layer_height` / `h_bℓ` uniformly.
-@inline state2dindex(a, i, j) = stateindex(a, i, j, 1)
+# 2-D (surface) specialization of `NumericalEarth.stateindex`, pinning k = 1: a 2-D
+# `Field` (Breeze's per-column surface- or boundary-layer height, a per-cell roughness
+# length) is read at column `(i, j)`, and anything uniform over grid points — a scalar
+# height, a flux formulation such as `MomentumRoughnessLength` — passes through.
+# Used by the atmosphere–surface flux kernels.
```
simone-silvestri:
```suggestion
# 2-D (surface) specialization of `NumericalEarth.stateindex`, pinning k = 1
```

simone-silvestri (approved):

PR #524, opened 2026-08-19
Pad CopernicusClimateDataStore ERA5 requests by two native cells

xkykai:
A regional read needs every native cell covering the requested `BoundingBox`, but the CopernicusClimateDataStore downloads requested the raw box as the CDS `area`, so delivered files (ERA5-Land in particular) came up one native cell short per edge and the read was rejected:

```
ArgumentError: the file does not cover the native grid in longitude: the read needs 39 cells
from -99.4 to -95.6, but the file labels 37 cells from -99.3 to -95.7.
```

Pad the requested area by two native cells per side, the same margin the CDSAPI extension's `era5_request_area` uses; the reader selects the exact cells from the larger file. Verified with regional `ERA5MonthlyLand` soil temperature and volumetric soil water reads.

### DIFF
--- a/ext/NumericalEarthCopernicusClimateDataStoreExt.jl
+++ b/ext/NumericalEarthCopernicusClimateDataStoreExt.jl
@@ -76,7 +76,7 @@ function Downloads.download(meta::ERA5Metadatum;
     hour = Dates.hour(date)
 
     # Build area constraint from region
-    area = padded_era5_area(meta.region, meta.dataset, meta.name)
+    area = build_era5_area(meta.region, meta.dataset, meta.name)
 
     # Build output prefix (filename without extension)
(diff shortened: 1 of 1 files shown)


ext/NumericalEarthCopernicusClimateDataStoreExt.jl
```diff
@@ -76,7 +76,7 @@ function Downloads.download(meta::ERA5Metadatum;
     hour = Dates.hour(date)
 
     # Build area constraint from region
-    area = build_era5_area(meta.region)
+    area = padded_era5_area(meta.region, meta.dataset, meta.name)
```
glwagner: I don't think there is any need to change the name of the function. What's wrong with `build_era5_area`?

glwagner (approved):

PR #525, opened 2026-08-19
Enable caching for `Field(metadatum, grid)` and rework `regrid_bathymetry` caching into the same framework

xkykai:
`Field(metadatum, grid)` re-runs the native materialization and regrid on every call — for high-resolution windows (OpenLandMap texture, WorldCover fractions) that is minutes per read of data that never changes. This adds an opt-in disk cache for the regridded result and rekeys `regrid_bathymetry`'s existing cache onto the same machinery.

- `Field(metadatum, grid; cache = true)` stores the regridded interior in a `field_cache` scratch space and reuses it when the key matches. The key covers everything that determines the result: dataset (with parameters), variable, date, region, target-grid geometry, field location, forwarded read keywords, and a size/mtime stamp of the local dataset file so a re-download invalidates the entry. Streaming datasets (Zarr, `/vsicurl/` windows) have no file to stamp, which is why this cache is opt-in. `overwrite_cache = true` skips the lookup and overwrites the entry with a fresh result — the refresh path for those datasets.
- `regrid_bathymetry` uses the same key and load/save path, carrying its processing parameters (`minimum_depth`, `interpolation_passes`, …) in the keyword slot; `BathymetryRegridding` is deleted. The bathymetry cache thereby gains re-download invalidation and parameter-aware dataset keys. Existing `bathymetry_cache` entries are orphaned and recompute once into `field_cache`.
- Load verifies the stored key against the request and falls back to recomputing on any mismatch or corrupt file.

```julia
sand = Field(Metadatum(:sand_fraction; dataset = OpenLandMapSoilDB(), region), grid; cache = true)
```

Round-trip and invalidation are tested in `test/test_field_cache.jl` (EN4) and `test/test_bathymetry.jl` (ETOPO), CPU.

### DIFF
--- a/src/Bathymetry/regrid_bathymetry.jl
+++ b/src/Bathymetry/regrid_bathymetry.jl
@@ -1,129 +1,15 @@
 
-# Scratch space for cached regridded bathymetry files
-bathymetry_cache_dir::String = ""
-function __init__()
-    global bathymetry_cache_dir = DataWrangling.download_cache("bathymetry_cache")
-end
-
-#####
-##### BathymetryRegridding configuration
-#####
-
-struct BathymetryRegridding
-    grid_type            :: String
-    grid_size            :: Tuple{Int, Int}
-    longitude            :: Tuple{Float64, Float64}
-    latitude             :: Tuple{Float64, Float64}
-    topology             :: Tuple{Symbol, Symbol}
-    float_type           :: Symbol
-    height_above_water   :: Union{Nothing, Float64}
-    minimum_depth        :: Float64
-    interpolation_passes :: Int
-    major_basins         :: Float64
-    dataset              :: String
-end
-
-function BathymetryRegridding(grid, metadata;
-                              height_above_water = nothing,
-                              minimum_depth = 0,
-                              interpolation_passes = 1,
-                              major_basins = 1)
-
-    Nx, Ny, _ = worksize(grid)
-    TX, TY, _ = topology(grid)
-    lon = x_domain(grid)
-    lat = y_domain(grid)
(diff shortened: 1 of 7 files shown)


src/Bathymetry/regrid_bathymetry.jl
```diff
@@ -177,7 +63,8 @@ Keyword Arguments
                   If `Inf` then no basins are removed. Default: 1.
 
 - `cache`: If `true` (default), caches the regridded bathymetry to disk and reuses it on subsequent
-           calls with the same grid and parameters. Set to `false` to force recomputation.
+           calls with the same grid, parameters, and dataset file; a re-download of the dataset
+           invalidates the entry. Set to `false` to force recomputation.
```
glwagner: I was going to say "we need a way to force recomputation". The kw `cache` isn't obvious to me. I thought `cache=false` would mean "don't cache this". But it seems, we actually always cache, is that right? I think there are two independent things needed: one `overwrite_cache` to force recomputration, and then `cache=true/false` to indicate whether to generate a cache at all

xkykai: Right — `cache = false` disables the cache entirely: nothing is read or written. The "force recomputation" wording was misleading (inherited from the old docstring); reworded.

Also added your second knob: `overwrite_cache = false` on both `regrid_bathymetry` and `Field(metadatum, grid)` skips the lookup and overwrites the entry with a freshly regridded result. With the file stamp in the key a re-download already invalidates automatically, so `overwrite_cache` is mainly for streaming datasets that have no file to stamp.

glwagner (approved):

Issue #526, opened 2026-08-19
The ECCO drive's `ECCO2` collection returns 403 for every path


xkykai:
`https://ecco.jpl.nasa.gov/drive/files` lists only `NearRealTime`, `Version4`, `Version5`, and a `README`. There is no `ECCO2` entry, yet that README still documents one:

```
  Version4                            ECCO Version 4 estimate
  NearRealTime                        ECCO-JPL Near Real-Time Kalman Filter and Smoother Solutions
  Version5                            ECCO Version 5 estimate
  ECCO2                               ECCO2 estimate
```

Every path beneath `files/ECCO2/` answers `403 Forbidden`, including to an account the drive otherwise accepts:

| path | status |
|---|---|
| `files` | 200 |
| `files/Version4/Release4/interp_monthly/THETA/1993/` | 200 |
| `files/Version5/` | 200 |
| `files/ECCO2/cube92_latlon_quart_90S90N/…` | 403 |
| `files/ECCO2/LLC270/…`, `files/ECCO2/LLC90/…` | 403 |

The 403 arrives with no `WWW-Authenticate` challenge even when `Authorization: Basic` is sent preemptively, and CI reproduces it with different credentials: the most recent `Data Downloading` run on `main` 403s on all 96 ECCO URLs it requests, split across `cube92_latlon_quart_90S90N` (40), `LLC270` (28), and `LLC90` (28), and none elsewhere.

Affected: `ECCO2Monthly`, `ECCO2Daily`, `ECCO2DarwinMonthly`, and `ECCO4DarwinMonthly` — both ECCO-Darwin datasets are stored under `files/ECCO2/` regardless of the grid their name refers to. `ECCO4Monthly` reads from `files/Version4/` and is unaffected.

```julia
using NumericalEarth, Dates, Downloads
Downloads.download(Metadata(:temperature; dataset=ECCO2Daily(), dates=[DateTime(1992, 1, 2)], dir=mktempdir()))
# RequestError: HTTP/2 403 while requesting .../ECCO2/cube92_latlon_quart_90S90N/daily/THETA/...
```

CI stays green because `test/download_utils.jl` falls back to NumericalEarthArtifacts, which holds the test fixtures; that predates the outage and masks it rather than fixing it, since the mirror carries only those files and not arbitrary dates.

[...]

PR #527, opened 2026-08-19
Name the collection when the ECCO drive refuses an ECCO2 download

xkykai:
Addresses #526.

The ECCO drive stopped serving its `ECCO2` directory, which holds `ECCO2Monthly`, `ECCO2Daily`, and both ECCO-Darwin datasets — those live there whatever grid their name refers to. Downloading one raised a bare `RequestError: HTTP/2 403`, which reads as bad credentials and sends people back to the README to re-copy a password that will not help.

```
The ECCO drive refused ECCO4DarwinMonthly, whose files live in the `ECCO2` directory it no longer
serves. This is not a problem with `ECCO_USERNAME` or `ECCO_WEBDAV_PASSWORD`. `ECCO4Monthly` reads
from the drive's `Version4` directory and is unaffected.
```

The download is still attempted, so these datasets recover on their own if the directory comes back. Other failures, and refusals from the collections the drive still serves, pass through untouched. The README gains the same note.

### DIFF
--- a/src/DataWrangling/ECCO/ECCO.jl
+++ b/src/DataWrangling/ECCO/ECCO.jl
@@ -273,20 +273,19 @@ function DataWrangling.metadata_url(m::Metadata{<:ECCO4Monthly})
     return ECCO4_url * dataset_variable_name(m) * "/" * year * "/" * m.filename
 end
 
-# The ECCO drive serves `Version4` and `Version5`, but every path under `ECCO2` — the quarter-degree
-# `cube92` fields and both ECCO-Darwin grids — answers `403 Forbidden` even to a signed-in account.
-# A bare `403` reads as bad credentials, so name the collection instead.
-is_ecco2_url(url) = occursin("/drive/files/ECCO2/", url)
+# The drive stopped serving its `ECCO2` directory, which holds the quarter-degree fields and both
+# ECCO-Darwin datasets whatever grid their name refers to. The download is still attempted in case it
+# returns, but a bare 403 reads as bad credentials, so say what it means instead.
+const ECCO2DriveDataset = Union{ECCO2Monthly, ECCO2Daily, ECCO2DarwinMonthly, ECCO4DarwinMonthly}
 
-function ecco_download_error(err, metadatum, url)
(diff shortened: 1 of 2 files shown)


glwagner (approved):
hmm ok. so we will remove these comments when ECCO2 comes back? is there a timeline for that?

xkykai:
> hmm ok. so we will remove these comments when ECCO2 comes back? is there a timeline for that?

yes, but also I'm not sure when ECCO2 is coming back, if at all. Don't seem to be able to find them on Earthdata for now as well.

Looking around we may need to rework the dataset ingestion for alternative sources like http://apdrc.soest.hawaii.edu/dods/public_data/ECCO/ECCO2. 

I've also opened https://github.com/NumericalEarth/NumericalEarth.jl/issues/528 for the record in the case of future migration of all ECCO datasets to Earthdata.

glwagner:
> > hmm ok. so we will remove these comments when ECCO2 comes back? is there a timeline for that?
> 
> yes, but also I'm not sure when ECCO2 is coming back, if at all. Don't seem to be able to find them on Earthdata for now as well.
> 
> Looking around we may need to rework the dataset ingestion for alternative sources like http://apdrc.soest.hawaii.edu/dods/public_data/ECCO/ECCO2.
> 
> I've also opened #528 for the record in the case of future migration of all ECCO datasets to Earthdata.

Should we throw an error when someone tries to build an `ECCO2` metadata then?

xkykai:
> > > hmm ok. so we will remove these comments when ECCO2 comes back? is there a timeline for that?
> > 
> > 
> > yes, but also I'm not sure when ECCO2 is coming back, if at all. Don't seem to be able to find them on Earthdata for now as well.
> > Looking around we may need to rework the dataset ingestion for alternative sources like http://apdrc.soest.hawaii.edu/dods/public_data/ECCO/ECCO2.
> > I've also opened #528 for the record in the case of future migration of all ECCO datasets to Earthdata.
> 
> Should we throw an error when someone tries to build an `ECCO2` metadata then?

maybe a warning? so that it is not blocking in the odd case where it does come back

PR #539, opened 2026-08-19
Document the frame contract of the matched retention parameters

xkykai:
The `(α, n)` from `soil_hydraulic_properties` pass exactly through each column's mean water contents at the matching heads, in that column's own `(θʳ, ν)` frame — so a stress closure reading the same column's saturation with them is exactly consistent, and the previous warning steering stress closures toward texture-class values misdiagnosed the failure.

- Rewrites the docstring warning around the real traps: collapsing the fields to a domain-wide scalar while each cell's saturation lives on its own curve, and reading moisture transplanted from another model's parameter space through these curves without rescaling.
- Notes the implied stress window is the pedotransfer family's own (Weynants places wilting wetter than lab texture tables), not a matching artifact.

Docs-only.

### DIFF
--- a/src/Lands/properties/soil_hydraulic_properties.jl
+++ b/src/Lands/properties/soil_hydraulic_properties.jl
@@ -187,6 +187,14 @@ range the slab operates in. 1 m rather than the textbook 3.3 m follows
 capacity for van Genuchten parameters of this family; the choice is worth little averaged
 over contrasting columns, hence a keyword rather than a constant.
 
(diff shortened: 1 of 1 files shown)


xkykai:
Superseded by #538: the stress now takes per-cell parameters (or the hydrology's retention curve object), so the consistent configuration exists in the API and the frame contract is documented on `PlantAvailableWaterStress` itself, where the parameters are consumed.

PR #533, opened 2026-08-19
(0.7.0) Rename `ElevationCorrection` to `AltitudeCorrection` and convert its lapse rate to the exchange-grid float type

xkykai:
Renames `ElevationCorrection` to `AltitudeCorrection`, and converts its lapse rate to the exchange-grid float type. `materialize_correction` converted the gravitational acceleration and gas constant but passed the stored `Float64` lapse rate, so building a coupled model with a correction on a `Float32` grid threw a `MethodError` against the struct's uniform `FT` slots.

```julia
using NumericalEarth, Oceananigans

grid = LatitudeLongitudeGrid(Float32; size = (2, 2, 1), latitude = (36, 37), longitude = (-98, -97),
                             z = (-1, 0), topology = (Bounded, Bounded, Bounded))
atmosphere = PrescribedAtmosphere(grid; surface_layer_height = 10, boundary_layer_height = 512)
land = SlabLand(grid)
AtmosphereLandModel(atmosphere, land; radiation = nothing,
                    exchanger_correction = AltitudeCorrection(300, 250))
# before: MethodError: no method matching AltitudeCorrection(::Int64, ::Int64, ::Float64, ::Float32, ::Float32, ::Field...)
```

A new `test_slab_land.jl` testset checks the materialized constants and the applied `T`/`p`/`q` shift on both `Float32` and `Float64` grids.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #547, opened 2026-08-20
Clamp exchange-grid fractional indices to the readable range

### LINKED ISSUE #546: Exchange-grid halo fractional indices fall off a regional prescribed grid
`initialize!` computes fractional indices over the exchange grid's halo columns as well as its interior (boundary-face stresses need them). Two distinct defects make those columns read outside a regional prescribed grid:

1. **Longitude folding**: `fractional_x_index` wraps longitude into `[λ₀ - Δλ/2, λ₀ - Δλ/2 + 360)`, so a halo node just west of λ = 0 lands near λ = 360, far past a regional source grid — out of bounds even counting halos, and asymmetric (the eastern neighbor doesn't fold). Fixed upstream in CliMA/Oceananigans.jl#5887 (Oceananigans 0.110.18).

2. **Reads past the allocated halo**: interpolation reads `⌊f⌋` and `⌊f⌋ + 1`, and a halo column's index can exceed what the source grid allocates. Halos are capped at the grid size, so a one-cell-deep direction has `H = 1` and the column at `N + 1` reads `N + 2`; a source grid finer than the exchange grid puts the column several source cells past the edge. Filling halos supplies data only inside the allocation, so no halo width fixes this in general.

MWE on `main` (no downloads):

```julia
using NumericalEarth
using Oceananigans
using Oceananigans.TimeSteppers: update_state!

grid = LatitudeLongitudeGrid(size = (6, 6, 3),
                             longitude = (0, 20), latitude = (0, 10), z = (-1, 0),
                             halo = (6, 6, 3), topology = (Bounded, Bounded, Bounded))

atmosphere = PrescribedAtmosphere(grid, [0.0])
model = AtmosphereOceanModel(atmosphere, ocean_simulation(grid))
update_state!(model)
```

```
ERROR: BoundsError: attempt to access 18×18×9×1 OffsetArray(::Array{Float64, 4}, -5:12, -5:12, -2:6, 1:1)
       with indices -5:12×-5:12×-2:6×1:1 at index [107, 0, 1, 1]
 [6] interp_atmos_time_series
```

The west halo column gets `fi = 107` on a 6-cell grid (defect 1). The remaining halo cells land just outside the source interior (`fi = Nx + 1`, `fj = 0`, `fj = Ny + 1`), which the halo here happens to absorb — on a grid whose halo cannot hold the read (e.g. `Ny = 1`, so `Hy = 1`) they throw too (defect 2). `PrescribedRadiation`'s regridder has the same shape.

Fixed in #547.

xkykai:
Closes #546.

Fractional indices are computed over the exchange grid's halo as well as its interior, so a halo column indexes outside the prescribed component's interior by design — it should read the component's own halo, and hence its boundary conditions. Clamping onto the interior would pin every halo column to the edge cell, so a boundary condition could never reach the exchange grid.

- Halo columns are clamped to `[1 - H, N + H)`. Interpolation reads `⌊f⌋` and `⌊f⌋ + 1`, so the upper bound is the largest representable index below `N + H`; `N + H - 1` would give the outermost halo cell zero weight and discard the boundary condition it holds.
- Interior columns are never clamped, so a component that does not cover the exchange grid keeps its out-of-range index and fails loudly instead of being interpolated over.
- Periodic directions are untouched, so seam interpolation on global grids is unchanged.
- `set_prescribed_field!` fills halos, so those reads see the field's boundary conditions rather than whatever initialization left behind. A dataset-backed series already fills them as snapshots are paged in.

Requires Oceananigans 0.110.18 (CliMA/Oceananigans.jl#5887), where a longitude just west of a regional grid indexes just west of it instead of folding to the far east.


glwagner:
Shouldn't we fill halos instead? If we clamp we cannot implement boundary conditions.

xkykai:
Filling halos alone can't fix the reported case. `fractional_x_index` folded a longitude just west of a regional grid to the far *east* of it, so on the two-cell grid in #546 the western halo column got index **36** while the array spans `-1:4`. That's out of range even with halos, so there's nothing to fill. It was also asymmetric — the eastern halo column didn't fold at all. Fixed upstream in CliMA/Oceananigans.jl#5887 by centering the fold window on the grid for non-periodic longitude; with that, the western column indexes `0` and reads the halo, as you'd expect.

The second one is arithmetic rather than folding. The exchanger iterates to `N+1`, and interpolation reads `⌊f⌋ + 1`, so the component grid needs **two** halo cells per interpolated direction. On that grid `Hy = 1`, so the northern column asks for y-index 3 out of `0:2`. The old interior clamp was hiding both of these at once, which is why one line appeared to fix a test with two distinct failures.

So this now clamps to the readable range, `[1 - H, N + H - 1]` where `H` is the number of halos and `N` is the number of grid points, instead of the interior. Halo columns are left alone wherever the component carries enough halo to hold them — so the read reaches the halo, and the boundary condition — and it falls back to the nearest cell only where there is no halo to read, as in latitude above. Periodic directions are untouched.

Also added `fill_halo_regions!` when a prescribed field is set: a dataset-backed series already has halos filled as snapshots are paged in, but the set-a-constant path did not, so those reads were seeing whatever initialization left behind.

xkykai:
@glwagner this is ready for your new round of reviews. once this is merged I can merge https://github.com/NumericalEarth/NumericalEarth.jl/pull/533 and then resolve some bugs found when testing https://github.com/NumericalEarth/NumericalEarth.jl/pull/401

simone-silvestri:
I also do not fully understand why only changing halos does not solve this issue. the longitude field is a periodic field where below 0 there is 360. Is it the correct choice to clamp an index? Is it not falsifying the interpolation?

xkykai:
> I also do not fully understand why only changing halos does not solve this issue. the longitude field is a periodic field where below 0 there is 360. Is it the correct choice to clamp an index? Is it not falsifying the interpolation?

the longitude wrap only applies to Periodic topology now. Before [CliMA/Oceananigans.jl#5887](https://github.com/CliMA/Oceananigans.jl/pull/5887) a `Bounded` grid folded into a window of `[λᵂ, λᵂ + 360)`, where `λᵂ` is the grid's western edge. This causes a point just west of a 2-cell grid to be folded to the far east (index 36).

filling halos alone doesn't really cover it, because interpolation reads `⌊f⌋` and `⌊f⌋+1`. 
For example with Ny = 1 there's only 1 halo point on each side, so the halo column at index 2 reads indices 2 and 3 which is out of bounds. In the past it has been run with `@inbounds` so it is silently incorrect.

On the interpolation side, the clamp doesn't touch regions within which we want to interpolate, since it is triggered only for the exchange grid's halo columns in non-periodic directions where the read goes out of range.

xkykai:
`xk/slab-canopy` (#401) had this branch merged into it, so #401's diff carried these changes. That is now reverted on `xk/slab-canopy` (11ea5f1d) and #401 is canopy-only again. This PR is unaffected and still merges into `main` on its own.

One thing to watch if #401 lands first: its history will then contain both this branch's older commits and that revert, so git sees `xk/clamp-halo-fractional-indices` as already merged and an ordinary merge here comes out empty (or conflicts against the revert). Branch fresh off `main` and re-apply the current tip's state (c1c34283, halo-only clamping) as a new commit instead of merging this branch.

simone-silvestri:
ok got it

src/EarthSystemModels/InterfaceComputations/InterfaceComputations.jl
```diff
@@ -115,6 +115,21 @@ function interface_kernel_parameters(grid)
     return kernel_parameters
 end
 
+# A halo column of the exchange grid indexes outside a regional component's interior by design:
+# the value there comes from the component's own halo, and so from its boundary conditions.
+# Interpolation reads cells `⌊f⌋` and `⌊f⌋ + 1`, so a halo column is held within `1 - H` and
+# `N + H`. Interior columns are never clamped, so a component that does not cover the exchange
+# grid fails instead of being interpolated over.
```
simone-silvestri:
```suggestion
# A grid with N < n limits to H < n which might not be enough for interpolation
```
is this what this comment means? Sorry I always find it very hard to read

src/EarthSystemModels/components.jl
```diff
@@ -24,9 +24,16 @@ exchange_grid(atmosphere, ocean, sea_ice, land=nothing) = grid(ocean)
 exchange_grid(atmosphere, ::Nothing, ::Nothing, land) = land.grid
 
 # Prescribed fields are FieldTimeSeries; set a `Number` into every time slice.
-# `nothing` leaves the field untouched.
+# `nothing` leaves the field untouched. Halos are filled because the exchanger interpolates
+# into them: a dataset-backed series has them filled as snapshots are paged in, so do the
+# same here.
```
simone-silvestri:
```suggestion
# `nothing` leaves the field untouched. 
```

simone-silvestri (approved):

src/EarthSystemModels/InterfaceComputations/InterfaceComputations.jl
xkykai: Agreed it's always quite hard to read.

reworded to

```julia
# Halo columns read the component's own halo, so the index is held where interpolation can
# reach: `⌊f⌋` and `⌊f⌋ + 1` must both lie within `1 - H` and `N + H`.
```

The rest (why interior columns are deliberately left unclamped) moved to the PR description.

src/EarthSystemModels/components.jl
xkykai: Taken, thanks — dropped the rationale, kept it to the one line. (Your suggestion had a trailing space, which trips the whitespace check, so I trimmed that.)

xkykai:
One fix beyond the review comments, in 716eba00: `halo_column` was derived from `size(exchange_grid)`, but `initialize!` launches the kernel over `interface_kernel_parameters`, which uses `worksize`. Those differ on a tripolar grid — `worksize(::RFTRG)` is `(Nx, Ny + 1, Nz)` — so the fold row `j = Ny + 1` is a worked interior row that `size` called a halo column and clamped. Both regridders now use `worksize`.

PR #564, opened 2026-08-24
Optional texture-class retention curve for PlantAvailableWaterStress

xkykai:
Evaluating the stress endpoints on the land's own retention curve breaks down when that curve comes from a pedotransfer reduction: fits matched through the very field-capacity and wilting heads the stress spans push `n` toward 1, and every moderate saturation then maps deep into stress (β ≈ 0.15 domain-wide was the HI-SCALE symptom).

- `PlantAvailableWaterStress(FT; texture = :loam)` (or explicit `inverse_air_entry_head`/`pore_size_uniformity`) gives the closure its own Carsel–Parrish curve and drops the retention-curve requirement on the hydrology, so it also pairs with `BucketHydrology`
- passing neither keeps the current behavior: endpoints on the land's per-cell curve
- new export `van_genuchten_texture_parameters`; docstring carries the pedotransfer warning

```julia
canopy = CanopyConductanceHumidity(moisture_stress = PlantAvailableWaterStress(texture = :loam))
```

### DIFF
--- a/src/EarthSystemModels/EarthSystemModels.jl
+++ b/src/EarthSystemModels/EarthSystemModels.jl
@@ -14,6 +14,7 @@ export
     FractionalHumidity,
     CriticalSaturation,
     PlantAvailableWaterStress,
+    van_genuchten_texture_parameters,
     DryLayerHumidity,
     StorageBasedDryLayerDepth,
     DryLayerVaporPistonVelocity,
--- a/src/EarthSystemModels/InterfaceComputations/InterfaceComputations.jl
+++ b/src/EarthSystemModels/InterfaceComputations/InterfaceComputations.jl
@@ -48,6 +48,7 @@ export
     FractionalHumidity,
     CriticalSaturation,
     PlantAvailableWaterStress,
+    van_genuchten_texture_parameters,
     DryLayerHumidity,
     StorageBasedDryLayerDepth,
     DryLayerVaporPistonVelocity,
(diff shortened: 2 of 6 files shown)


xkykai:
Closing after a second look at the mechanism; two findings, either one disqualifying.

**The motivation doesn't hold.** `soil_hydraulic_properties` fits (α, n) through the water contents at `matching_heads = (1, 150)` m, the same heads the stress evaluates, so both endpoint saturations are reproduced exactly whatever `n` the fit returns. β reads only 𝒮 and those two endpoints, so a fitted `n` near 1 cannot push moderate saturations into stress. Low `n` at loamy textures is the pedotransfer regression's own output (Weynants gives n ≈ 1.18 for an SGP silt loam, before any matching), and β ≈ 0.16 at θ ≈ 0.20 is genuinely near-wilting on that curve (field capacity 0.379, wilting point 0.166).

**The texture path mixes saturation frames.** It compares the land's 𝒮 = (θ − θʳ)/(ν − θʳ) against Se endpoints evaluated on the Carsel–Parrish curve's own frame (θʳ = 0.078, θs = 0.43). On a Weynants silt loam that gives β = 0.79 at the soil's wilting content and β = 0 only at θ = 0.013: transpiration never shuts down, with fluxes that look plausible while it happens.

There is a real question underneath: Weynants puts θ(150 m) at 0.166 where Carsel–Parrish's comparable loam puts 0.088. If that dry-end disagreement turns out to matter, the workable design takes Carsel–Parrish's water *contents* and converts them through the hydrology's own ν and θʳ per cell, keeping the wilting content the source intended. Follow-up material, once the domain-wide low-β symptom is re-attributed with per-cell retention fields (#538).

PR #354, opened 2026-06-16
Adds bathymetry downscaling by averaging

jagoosw:
This PR modifies `regrid_bathymetry` to take an argument `method` (instead of `interpolaiton_passes`) which is `Interpolate(1)` by default, and adds `MedianAveraging` which takes the average of the dataset depths within each (coarser) target cell.

This is with 20 interpolation passes:
image

The new method is almost smoother in the spiky bits and sharper on ridges.

I believe this is how they make the ORCA bathymetries: https://www.drakkar-ocean.eu/publications/reports/orca025-g70.pdf and this SCOR report claims median is better but still flawed for coarse models: https://scor-int.org/WG107Report.pdf

I haven't checked it works on GPU yet but hopefully.

PR #566, opened 2026-08-24
Read dataset windows at a target-matched resolution and regrid them in tiles

xkykai:
`Field(metadatum, grid)` downloaded every native pixel of a window and held it, plus a field on
it, in memory. A CONUS window of the 30 m OpenLandMap product is ~150 GB of `Float32` per depth
layer, so the dataset's resolution set the ceiling rather than the model domain.

This is two independent changes, and a dataset can take either on its own.

The first bounds the bytes read, and is opt-in because it changes the values you get.
`OpenLandMapSoilDB(aggregation_factor = nothing)` reads the coarsest lattice that still
oversamples the target twofold, served from a cloud-optimized GeoTIFF's own average-resampled
overview pyramid. The CONUS window matched to a 12 km grid is 2297 × 1016 cells rather than
232000 × 100000, and reads in 15 s. This needs a source that ships a pyramid, so
`OpenLandMapSoilDB` is the only dataset offering it today.

The second bounds resident memory, and is on by default because it changes nothing observable. The regrid runs a tile at a time, reading only the window each
tile covers, so peak residency follows the tile rather than the window. This needs only a file
that can be read in hyperslabs, which any plain lon/lat NetCDF can. Each tile interpolates from a
windowed `Field` over the native grid, so the result is bitwise identical to an untiled regrid at
any tiling.

Reading the pyramid also averages where the old path sampled the four pixels bracketing a cell
center, four of the ~1100 a 1 km cell covers. Against the exact area-weighted mean, RMS sand
error falls from 2.04 % to 0.66 % (Grand Canyon) and 1.77 % to 0.50 % (Iowa).

```julia
using NumericalEarth, Oceananigans, ArchGDAL

# ~1 km target over the Grand Canyon. The source is a global 30 m soil raster.
grid = LatitudeLongitudeGrid(size = (111, 111, 3), longitude = (-112.5, -111.5), latitude = (35.5, 36.5),
                             z = [-1.0, -0.6, -0.3, 0.0])
region = BoundingBox(grid)

# Default: native 30 m, regridded a tile at a time so the window need not be resident.
sand = Field(Metadatum(:sand_fraction; dataset = OpenLandMapSoilDB(), region), grid)

# Opt in to a target-sized read, which also averages rather than samples.
matched = OpenLandMapSoilDB(aggregation_factor = nothing)
sand_matched = Field(Metadatum(:sand_fraction; dataset = matched, region), grid)
```

| read | factor | source | tiles | peak memory | elapsed |
| --- | --- | --- | --- | --- | --- |
| matched (opt-in) | 16 | 252 × 252 × 3 = 0.7 MB | 1 × 1 | 175 MB | 44 s |
| native (default) | 1 | 4002 × 4002 × 3 = 183.3 MB | 2 × 2 | 691 MB | 157 s |

With the matched read the source already fits one tile, so on a coarse target the opt-in change
usually removes the need for the default one. Holding the native read fixed and varying
`tile_bytes` isolates the default:

| `tile_bytes` | tiles | peak memory | elapsed | surface sand |
| --- | --- | --- | --- | --- |
| 64 MB (default) | 2 × 2 | 535 MB | 63 s | mean 0.525, 11189 finite |
| 64 GB | 1 × 1 | 1045 MB | 70 s | mean 0.525, 11189 finite |

At the scale this is for, the 30 m Southern Great Plains window (8002 × 8002 × 3) falls from
5470 MB to 598 MB. Peak memory is the process's high-water resident size, so it counts the
Julia runtime and GDAL buffers as well as the source.

Bounding the bytes read:

- `DataWrangling.matching_resolution_dataset(dataset, grid)` selects the read resolution; the
  default returns `dataset` unchanged. `Field(metadatum, grid)` and `set!(field, metadatum)`
  rebuild the metadatum on it, so the cached filename follows the realized lattice.
- `OpenLandMapSoilDB(; aggregation_factor = 1)`: an integer pins the read at that many 30 m
  pixels per cell side, and the default reads at full resolution under its existing filename.
  `nothing` sizes the read to the target grid; `Field(metadatum)`, which has no target, still
  reads at full resolution.

Bounding resident memory:

- `DataWrangling.retrieve_window(metadata, longitude_indices, latitude_indices)` with the
  `windowed_retrieval(dataset)` trait; the default reads the whole file and views into it.
  `netcdf_retrieve_window` implements it for any plain lon/lat NetCDF, north-first files included,
  so a dataset opts in with two lines. `GHSL` takes it, and `CopernicusDEM` does once #587
  corrects its coordinate-variable names.
- `tile_bytes` bounds one tile's source.


src/DataWrangling/metadata.jl
```diff
@@ -815,6 +815,42 @@ cadence — e.g. the surface companion of a pressure-level reanalysis product.
...
+is refused.
+"""
+matching_resolution_dataset(dataset, grid) = dataset
+
+"""
+    categorical(dataset)
+
+Whether `dataset` stores class codes rather than continuous quantities. Averaging class codes is
+meaningless, so a categorical dataset is never read at a coarsened resolution: it is aggregated
+by counting instead, over its native pixels.
```
simone-silvestri: I am not sure what "class code" means when compared to continuous quantities

src/DataWrangling/metadata.jl
```diff
@@ -815,6 +815,42 @@ cadence — e.g. the surface companion of a pressure-level reanalysis product.
 """
 function matching_single_level_dataset end
 
+"""
+    minimum_horizontal_spacing(grid)
+
+Return the smallest horizontal cell spacing of `grid` in degrees — the finest scale a dataset
+has to resolve to fill every cell of `grid`. For a distributed grid the result is the global
+minimum, identically on every rank.
```
simone-silvestri: this function is just the minimum horizontal spacing. I think this docstring wants to convey more than what this function actually is, so it results confusing.

src/DataWrangling/WorldCover/WorldCover.jl
```diff
@@ -431,6 +435,9 @@ Oceananigans.Fields.location(::ESAWorldCoverMetadatum) = (Center, Center, Center
 ##### hook, dispatched on the metadatum.
 #####
 
+# Counted, not interpolated: tiling drives the bilinear kernel directly, so it must not run here.
```
simone-silvestri: how does this comment apply to 
`DataWrangling.default_regrid(::ESAWorldCoverMetadatum) = false`?

simone-silvestri: maybe a `default_regrid` should have a type rather than `true/false`?

simone-silvestri:
How does this PR connect to #354? I think they want to do the same thing?
Do we maybe want to clean this up a bit by defining regridding types like in #354?

xkykai:
Split into #585 (target-matched reads) and #586 (tiled regridding, based on #585) — they are independent changes and were not reviewable together.

src/DataWrangling/metadata.jl
xkykai: removed

xkykai:
> How does this PR connect to #354? I think they want to do the same thing? Do we maybe want to clean this up a bit by defining regridding types like in #354?

They aim to do separate things. #354 is concerned with the averaging part, here I'm trying to implement something that allows for data-efficient *loading* of datasets into a target grid that is much coarser than the raw dataset by not materializing the dataset on the native grid itself. This is useful for some very high resolution datasets (soil types, land use types etc) where their resolutions are O(10m) when we want to run simulations at O(km) resolutions.

I'm hoping that this is a more generic framework that would work for a broad(er) range of datasets

src/DataWrangling/WorldCover/WorldCover.jl
xkykai: this is a useless comment, removed

xkykai:
I will close this for now in the spirit of reviewability as this implements two separate functionality which are now split into https://github.com/NumericalEarth/NumericalEarth.jl/pull/586 and https://github.com/NumericalEarth/NumericalEarth.jl/pull/585.

PR #568, opened 2026-08-24
Cache MODIS granules beside the reprojected files

xkykai:
MODIS granules were downloaded into a temporary directory and discarded after each warp, so a new region or date over the same sinusoidal tiles refetched granules it already had (~200 MB per composite).

- Granules are kept under `granules/` beside the reprojected NetCDFs and skipped when already there, the way the ASTER GED tiles already are.
- `granule_urls` goes through `cmr_granules`, which pages the CMR query rather than assuming one page of 2000 holds every granule; `cmr_granules` takes a `date` for products whose granules are dated.
- `earthdata_download_cached` stages its download, so an interrupted fetch cannot leave a truncated file where the next call expects a complete one.

Based on #424.

### DIFF
--- a/ext/NumericalEarthArchGDALExt/NumericalEarthArchGDALExt.jl
+++ b/ext/NumericalEarthArchGDALExt/NumericalEarthArchGDALExt.jl
@@ -12,7 +12,7 @@ using Oceananigans.Grids: λnodes, φnodes
 
 using NumericalEarth.DataWrangling: BoundingBox, native_grid, native_region_grid,
                                     dataset_variable_name, bounding_box_intersects,
-                                    cmr_granules, earthdata_download, earthdata_download_cached,
+                                    cmr_granules, earthdata_download_cached,
                                     figshare_article_url, write_atomically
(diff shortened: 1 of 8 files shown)


xkykai:
Ran the network-gated `test_modis_land_downloading.jl` locally (excluded from the default suite): 36/36 pass. The new testset warps a second window from the cached granule in 0.4 s against 30.6 s for the first composite, and the paged `cmr_granules` query resolves to the protected LP DAAC endpoints, which authenticate through the same netrc downloader the ASTER GED tiles use.

xkykai:
CI red is not from this branch: `(CPU)` fails in `test_speedy_coupling.jl` on a `RingGrids.Field` `zeros` MethodError, which the base branch carries (its SpeedyWeather compat is behind main's), and `(GPU)` fails opening a truncated ECCO `SALT_1993_01.nc` in the runner's depot. The suites this touches pass: `test_earthdata` 11/11, `test_modis_land` 350/350, and Data Downloading green on both platforms.

PR #569, opened 2026-08-24
Cache the seasonal gap fill beside its lattice

xkykai:
`fill_seasonal_gaps!` runs for minutes over a continental native lattice, and the model grid that reads its result is no part of the fill: a 12 km, a 6 km and a 3 km configuration built on the same series each repeat the identical work.

- New `cache` keyword: the path of a JLD2 file holding the filled series with its provenance and reach.
- The file is read back when the series, the class map, and the fill parameters still match what produced it; anything else refills and rewrites.

```julia
fill_seasonal_gaps!(leaf_area_index_series, land_cover;
                    cache = joinpath(dir, "leaf_area_index_filled.jld2"))
```

Based on #424.

### DIFF
--- a/src/DataWrangling/gap_filling.jl
+++ b/src/DataWrangling/gap_filling.jl
@@ -227,6 +227,58 @@ end
 @inline clamp_to_range(value, ::Nothing) = value
 @inline clamp_to_range(value, range) = clamp(value, first(range), last(range))
 
+#####
+##### Caching a filled series
+#####
+##### The chain is expensive over a continental lattice and deterministic in its inputs and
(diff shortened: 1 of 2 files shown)


xkykai:
`test_seasonal_gap_fill` passes 83/83 in CI. The two red jobs are elsewhere: `(CPU)` fails in `test_speedy_coupling.jl` on a `RingGrids.Field` `zeros` MethodError the base branch carries (its SpeedyWeather compat is behind main's), and Data Downloading fails spawning the `copernicusmarine` binary in `test_aviso_downloading.jl` with ETXTBSY. Both pass on the sibling PR off the same base.

PR #572, opened 2026-08-24
Adds an OceanBioME extension to streamline coupling

jagoosw:
- [x] Light shading by sea ice
- [x] Gas exchange driven by wind speed from the atmosphere
- [x] Gas exchange blocked by sea ice
- [x] Make freshwater inputs for tracers

This PR also adds the ability to pass generic tracers into the prescribed atmospheres which get interpolated onto the exchange grid. It can handle constant fields (no-op) and reduced fields. For example:
```julia
O₂ = Oceananigans.Fields.ConstantField(100)
pCO₂ = FieldTimeSeries((nothing, nothing, nothing), grid, 0.0f0:10800.0f0:3.15252f7, time_indexing = Cyclical())
pCcO₂ .= reshape(250 .+ 200 .* [1:2920;] ./ 2920, 1, 1, 1, 2920)

atmosphere = JRA55PrescribedAtmosphere(arch; time_indices_in_memory=2, tracers = (; O₂, pCO₂))
```
In the coupled model we get:
```julia
julia> coupled_model.interfaces.exchanger.atmosphere.state.pCO₂
1×1×1 Field{Nothing, Nothing, Nothing} reduced over dims = (1, 2, 3) on ImmersedBoundaryGrid on CPU
├── grid: 50×50×10 ImmersedBoundaryGrid{Float64, Periodic, RightCenterFolded, Bounded} on CPU with 7×7×7 halo
├── boundary conditions: FieldBoundaryConditions
│   └── west: Nothing, east: Nothing, south: Nothing, north: Nothing, bottom: Nothing, top: Nothing, immersed: ZeroFlux
└── data: 1×1×1 OffsetArray(::Array{Float64, 3}, 1:1, 1:1, 1:1) with eltype Float64 with indices 1:1×1:1×1:1
    └── max=250.068, min=250.068, mean=250.068

julia> coupled_model.interfaces.exchanger.atmosphere.state.O₂
ConstantField(100)
```

To change the parameters for the gas exchange you pass `biogeochemistry_interface_kwargs` to `EarthSystemModel`, which puts them into `CarbonDioxideGasExchangeBoundaryCondition` etc.

~This PR relies on https://github.com/OceanBioME/OceanBioME.jl/pull/399~ - merged
~This PR relies on https://github.com/OceanBioME/OceanBioME.jl/pull/419~ - merged


ext/NumericalEarthOceanBioMEExt/state_exchange_interface.jl
```diff
@@ -0,0 +1,20 @@
...
+
+    return PARFromShortwave(surface_PAR; photosynthetic_fraction_of_shortwave)
+end
+
+exchanged_tracers(biogeochemistry) = 
+    (required_biogeochemical_tracers(biogeochemistry.oxygen)...,
+     dissolved_inorganic_carbon_names(biogeochemistry.inorganic_carbon))
+
+dissolved_inorganic_carbon_names(::CarbonateSystem{1}) = (:DIC, )
+dissolved_inorganic_carbon_names(::CarbonateSystem{N}) where N = map(n->Symbol(:DIC, n), 1:N)
```
simone-silvestri:
```suggestion
dissolved_inorganic_carbon_names(::CarbonateSystem{N}) where N = map(n->Symbol(:DIC, n), 1:N)

```

ext/NumericalEarthOceanBioMEExt/assemble_net_fluxes.jl
```diff
@@ -0,0 +1,41 @@
...
+import NumericalEarth.Radiations: apply_air_sea_biogeochemical_radiative_fluxes!
+import NumericalEarth.Oceans: update_net_ocean_biogeochemical_fluxes!
+
+function update_net_ocean_biogeochemical_fluxes!(coupled_model, biogeochemistry::CompleteBiogeochemistry, ocean, grid)
+    update_net_ocean_biogeochemical_fluxes!(coupled_model, biogeochemistry.underlying_biogeochemistry, biogeochemistry, ocean, grid)
+    update_net_ocean_biogeochemical_fluxes!(coupled_model, biogeochemistry.light_attenuation, biogeochemistry, ocean, grid)
+    update_net_ocean_biogeochemical_fluxes!(coupled_model, biogeochemistry.sediment, ocean, biogeochemistry, grid)
+    update_net_ocean_biogeochemical_fluxes!(coupled_model, biogeochemistry.particles, ocean, biogeochemistry, grid)
+    update_net_ocean_biogeochemical_fluxes!(coupled_model, biogeochemistry.modifiers, ocean, biogeochemistry, grid)
+
```
simone-silvestri:
```suggestion
```

ext/NumericalEarthOceanBioMEExt/assemble_net_fluxes.jl
```diff
@@ -0,0 +1,41 @@
...
+    return nothing
+end
+
+update_net_ocean_biogeochemical_fluxes!(coupled_model, component, ocean, biogeochemistry, grid) = nothing
+
+apply_air_sea_biogeochemical_radiative_fluxes!(coupled_model) =
```
simone-silvestri: Isn't this type piracy from the same function defined in the main src? You probably want to qualify this coupled model as a coupled model with an Oceananigans ocean containing a biogeochemistry
(you can actually qualify it in main, then extend it here for the particular biogeochemistry)

ext/NumericalEarthOceanBioMEExt/assemble_net_fluxes.jl
```diff
@@ -0,0 +1,41 @@
...
+    set!(PAR.surface_shortwave, penetrating_shortwave)
+
+    return nothing
+end
+
+function update_ocean_biogeochemical_boundary_conditions!(coupled_model, biogeochemistry, ocean, grid)
+    fluxes = 0
+end
+
+@inline net_fluxes(condition::GasExchange) = condition
```
simone-silvestri:
```suggestion
@inline net_fluxes(condition::GasExchange) = condition

```

ext/NumericalEarthOceanBioMEExt/assemble_net_fluxes.jl
jagoosw: yes some mistakes I'll fix!

jagoosw:
@simone-silvestri are you happy with how I've added the tracers to the prescribed atmospheres? In the normal case I think there is no change to anything.

ext/NumericalEarthOceanBioMEExt/assemble_net_fluxes.jl
```diff
@@ -0,0 +1,14 @@
+#####
+##### radiation from atmosphere blocked by sea ice
+#####
+apply_air_sea_biogeochemical_radiative_fluxes!(coupled_model, biogeochemistry::DiscreteBiogeochemistry{<:NutrientsPlanktonDetritus}) =
```
simone-silvestri:
```suggestion
#####

apply_air_sea_biogeochemical_radiative_fluxes!(coupled_model, biogeochemistry::DiscreteBiogeochemistry{<:NutrientsPlanktonDetritus}) = 
```

src/Atmospheres/interpolate_atmospheric_state.jl
```diff
@@ -88,6 +83,9 @@ end
 @inline get_fractional_index(i, j, ::Nothing) = nothing
 @inline get_fractional_index(i, j, frac) = @inbounds frac[i, j, 1]
 
+@inline underlying_data(f) = f.data
+@inline underlying_data(::ConstantField) = nothing
```
simone-silvestri: Shouldn't it be passed through?
```suggestion
@inline underlying_data(c::ConstantField) = c
```

src/Atmospheres/interpolate_atmospheric_state.jl
```diff
@@ -128,12 +124,27 @@ end
...
         surface_atmos_state.v[i, j, 1] = vᵃᵗ
-        surface_atmos_state.T[i, j, 1] = Tᵃᵗ
         surface_atmos_state.p[i, j, 1] = pᵃᵗ
-        surface_atmos_state.q[i, j, 1] = qᵃᵗ
         surface_atmos_state.Jʳⁿ[i, j, 1] = Mr
         surface_atmos_state.Jˢⁿ[i, j, 1] = Ms
     end
+
+    for (tn, tv) in pairs(atmos_tracers)
+        update_tracer_state!(i, j, surface_atmos_state[tn], tv, atmos_args, t_itp)
```
simone-silvestri:
```suggestion
        @inbounds update_tracer_state!(i, j, surface_atmos_state[tn], tv, atmos_args, t_itp)
```

src/Oceans/ocean_simulation.jl
```diff
@@ -433,37 +447,40 @@ function hydrostatic_ocean_simulation(grid;
...
-    default_additional_fluxes = (u=nothing, v=nothing, T=nothing, S=nothing)
+    tracers = (tracers..., required_biogeochemical_tracers(biogeochemistry)...)
+
+    default_additional_fluxes = merge((u=nothing, v=nothing), NamedTuple(name => nothing for name in tracers))
     additional = merge(default_additional_fluxes, additional_surface_fluxes)
 
-    # Freshwater heat content is `Σᵢ Tᵢ Jʷᵢ`, the Freshwater salinity content is assumed to be 0 for the moment (no salinity for incoming freshwater)
     freshwater_heat_content = Field{Center, Center, Nothing}(grid)
-    freshwater_salt_content = ZeroField()
+    default_freshwater_tracer_content = NamedTuple(name => name === :T ? freshwater_heat_content : ZeroField()
```
simone-silvestri, line 524: I think you need to figure out if the additional tracer content is required also for passive tracers instead of only T, i.e. if incoming freshwater like rivers or rain carries tracers.

src/Oceans/ocean_simulation.jl
```diff
@@ -433,37 +447,40 @@ function hydrostatic_ocean_simulation(grid;
...
     end
 
-    # Merge user-supplied additional fluxes with defaults
-    default_additional_fluxes = (u=nothing, v=nothing, T=nothing, S=nothing)
+    tracers = (tracers..., required_biogeochemical_tracers(biogeochemistry)...)
+
+    default_additional_fluxes = merge((u=nothing, v=nothing), NamedTuple(name => nothing for name in tracers))
     additional = merge(default_additional_fluxes, additional_surface_fluxes)
 
-    # Freshwater heat content is `Σᵢ Tᵢ Jʷᵢ`, the Freshwater salinity content is assumed to be 0 for the moment (no salinity for incoming freshwater)
```
simone-silvestri, line 511: I would rather keep this comment (in an updated fashion), so we understand what is this additional freswater_tracer_content

simone-silvestri: because in the future I wanted to aqdd salinity of freshwater sources

src/Oceans/ocean_simulation.jl
```diff
@@ -528,3 +545,6 @@ function EarthSystemModels.heat_capacity(::TEOS10EquationOfState{FT}) where FT
     cₚ⁰ = SeawaterPolynomials.TEOS10.teos10_reference_heat_capacity
     return convert(FT, cₚ⁰)
 end
+
+# setup helpers
+biogeochemistry_surface_exchanged_tracers(biogeochemistry) = tuple()
```
simone-silvestri:
```suggestion
biogeochemistry_surface_exchanged_tracers(biogeochemistry) = tuple()

```

src/Oceans/ocean_simulation.jl
jagoosw: Oh yes I'll try to work that out. Is this where it would be set for rivers from the land enters too or is this just the atmosphere freshwater input?

src/Oceans/ocean_simulation.jl
jagoosw: Sorry didn't mean to remove

src/Oceans/ocean_simulation.jl
simone-silvestri: it's also from the rivers. At the moment they are all collected together, but if we need to separate the different sources, we can in the `assemble_net_fluxes!`. Note that we add the content for temperature, because the freshwater input it _neutral_ in temperature, so it carries the same T as the model. It is not the same for salinity which is reduced because freshwater comes in with 0 content. If you leave 0 content for tracers it means that a freshwater source effectively dilutes away that tracer near the inlet.

simone-silvestri:
>  Light shading by sea ice

I guess this is what we need also for the ocean, at the moment all the shortwave is collected on top of the sea ice (weighted by concentration) if there is sea ice, but shortwave should penetrate sea ice depending on the thickness. Should be open another PR to implement this in general?

jagoosw:
> I guess this is what we need also for the ocean, at the moment all the shortwave is collected on top of the sea ice (weighted by concentration) if there is sea ice, but shortwave should penetrate sea ice depending on the thickness. Should be open another PR to implement this in general?

What I ended up doing was just using the shortwave what reaches the sea surface from the existing interface so I didn't implement anything new, but yes it probably would be better to have the ice attenuate depending on the thickness which we can do in a different PR

simone-silvestri:
> What I ended up doing was just using the shortwave what reaches the sea surface from the existing interface so I didn't implement anything new, but yes it probably would be better to have the ice attenuate depending on the thickness which we can do in a different PR

Ok, then I guess we can tackle this issue once this PR is merged

src/Atmospheres/interpolate_atmospheric_state.jl
jagoosw: We don't actually need to pass it because this is the field that is getting update but we can incase we need it elsewhere

src/Oceans/ocean_simulation.jl
jagoosw: I think from rain its fine to assume 0 for most tracers right? But for rivers its going to be more challenging and probably will need a dataset

jagoosw: Maybe at some point we should make the content part of the land model? For now we can manually pass content to freshwater_tracer_content.

jagoosw: Just checking, are the freshwater fluxes from land model meant to be conservatively interpolated onto the ocean grid? I guess maybe its fine on higher than dataset resolution but on low resolution it looks like the integral freshwater flux is about an order of magnitude lower in the interface state for a 50x50 Tripolar grid

jagoosw:
@lgloege do you have any thoughts on this PR? The behaviour now is that if you have DIC and oxygen they automatically get the default gas exchange boundary condition. ~To change from the default parameters you just need to change or add a new method for:
`biogeochemical_interface`~ You can also pass `biogeochemistry_interface_kwargs` to `EarthSystemModel` to pass any of the parameters for `CarbonDioxideGasExchangeBoundaryCondition` and `OxygenGasExchangeBoundaryCondition` (they will all get the same for now, if you want different ones for different DIC realisations you can add a new `biogeochemical_interface` method).

Light should work out of the box you just have to set the surface PAR to `PARFromShortwave(grid)`.

After this PR the only major component that will need some work is rivers (I think).

Are you happy with this interface (at least as a starting point)? And have you found any other setup difficulties that I should address here?

jagoosw:
I think the relevant tests now pass

ext/NumericalEarthOceanBioMEExt/gas_exchange.jl
```diff
@@ -0,0 +1,152 @@
...
+     biogeochemistry_surface_exchanged_tracers(biogeochemistry.underlying_biogeochemistry.inorganic_carbon)...)
+
+biogeochemistry_surface_exchanged_tracers(::Oxygen) = (:O₂, )
+biogeochemistry_surface_exchanged_tracers(::AbstractInorganicCarbon{1}) = (:DIC, )
+biogeochemistry_surface_exchanged_tracers(::AbstractInorganicCarbon{N}) where N = carbon_replicate_names(Val(N))
+
+@inline carbon_replicate_names(::Val{N}) where N = ntuple(n -> Symbol(:DIC, n), Val(N))
+
+@inline surface_wind_speed(exchanger) =
+    sqrt(exchanger.atmosphere.state.u^2 + exchanger.atmosphere.state.v^2)
```
simone-silvestri:
```suggestion
@inline surface_wind_speed(exchanger) = sqrt(exchanger.atmosphere.state.u^2 + exchanger.atmosphere.state.v^2)
```

src/Atmospheres/interpolate_atmospheric_state.jl
```diff
@@ -40,13 +41,7 @@ function EarthSystemModels.interpolate_state!(exchanger, grid, atmosphere::Presc
...
     # Simplify NamedTuple to reduce parameter space consumption.
     # See https://github.com/CliMA/NumericalEarth.jl/issues/116.
-    atmosphere_data = (u   = atmosphere_fields.u.data,
-                       v   = atmosphere_fields.v.data,
-                       T   = atmosphere_fields.T.data,
-                       p   = atmosphere_fields.p.data,
-                       q   = atmosphere_fields.q.data,
-                       Jʳⁿ = atmosphere_fields.Jʳⁿ.data,
-                       Jˢⁿ = atmosphere_fields.Jˢⁿ.data)
+    atmosphere_data = NamedTuple(k=>underlying_data(v) for (k, v) in pairs(atmosphere_fields))
```
simone-silvestri, line 44: I think this line is type unstable

src/Oceans/assemble_net_ocean_fluxes.jl
```diff
@@ -13,8 +13,14 @@ using ..EarthSystemModels.InterfaceComputations: computed_fluxes
 # Fallback for an ocean-only model (it has no interfaces!)
 EarthSystemModels.update_net_fluxes!(coupled_model::Union{NoOceanInterfaceModel, NoInterfaceModel}, ocean::OceananigansModelSimulations) = nothing
 
-EarthSystemModels.update_net_fluxes!(coupled_model, ocean::OceananigansModelSimulations) =
+function EarthSystemModels.update_net_fluxes!(coupled_model, ocean::OceananigansModelSimulations)
     update_net_ocean_fluxes!(coupled_model, ocean, ocean.model.grid)
+    update_net_ocean_biogeochemical_fluxes!(coupled_model, ocean.model.biogeochemistry, ocean, ocean.model.grid)
+
```
simone-silvestri:
```suggestion
```

src/Oceans/ocean_simulation.jl
```diff
@@ -486,49 +500,50 @@ function hydrostatic_ocean_simulation(grid;
...
+    tracers = (tracers..., required_biogeochemical_tracers(biogeochemistry)...)
+
+    default_additional_fluxes = merge((u=nothing, v=nothing), NamedTuple(name => nothing for name in tracers))
     additional = merge(default_additional_fluxes, additional_surface_fluxes)
 
     # Freshwater heat content is `Σᵢ Tᵢ Jʷᵢ`, the Freshwater salinity content is assumed to be 0 for the moment (no salinity for incoming freshwater)
     freshwater_heat_content = Field{Center, Center, Nothing}(grid)
-    freshwater_salt_content = ZeroField()
+    default_freshwater_tracer_content = NamedTuple(name => name === :T ? freshwater_heat_content : ZeroField()
+                                                    for name in tracers)
```
simone-silvestri:
```suggestion
    default_freshwater_tracer_content = NamedTuple(name => name === :T ? freshwater_heat_content : ZeroField() for name in tracers)
```

src/Oceans/ocean_simulation.jl
```diff
@@ -486,49 +500,50 @@ function hydrostatic_ocean_simulation(grid;
...
     u_top_bc = build_top_bc(τˣ, λˣ, additional.u)
     v_top_bc = build_top_bc(τʸ, λʸ, additional.v)
-    T_top_bc = build_tracer_top_bc(Jᵀ, Jʷ, freshwater_heat_content, additional.T, :T)
-    S_top_bc = build_tracer_top_bc(Jˢ, Jʷ, freshwater_salt_content, additional.S, :S)
+
+    flux_tracers = (:T, :S, surface_exchanged_tracers...)
+    tracer_top_fluxes = NamedTuple(name => name ∈ flux_tracers ? Field{Center, Center, Nothing}(grid) : ZeroField()
+                                    for name in tracers)
+    tracer_top_bcs = NamedTuple(name => build_tracer_top_bc(tracer_top_fluxes[name], Jʷ, freshwater_tracer_content[name],
+                                                             additional[name], name)
```
simone-silvestri:
```suggestion
    tracer_top_bcs = NamedTuple(name => build_tracer_top_bc(tracer_top_fluxes[name], Jʷ, freshwater_tracer_content[name], additional[name], name)
```

src/Oceans/ocean_simulation.jl
```diff
@@ -486,49 +500,50 @@ function hydrostatic_ocean_simulation(grid;
...
 
-    # Construct ocean boundary conditions including surface forcing and bottom drag
     u_top_bc = build_top_bc(τˣ, λˣ, additional.u)
     v_top_bc = build_top_bc(τʸ, λʸ, additional.v)
-    T_top_bc = build_tracer_top_bc(Jᵀ, Jʷ, freshwater_heat_content, additional.T, :T)
-    S_top_bc = build_tracer_top_bc(Jˢ, Jʷ, freshwater_salt_content, additional.S, :S)
+
+    flux_tracers = (:T, :S, surface_exchanged_tracers...)
+    tracer_top_fluxes = NamedTuple(name => name ∈ flux_tracers ? Field{Center, Center, Nothing}(grid) : ZeroField()
+                                    for name in tracers)
```
simone-silvestri:
```suggestion
    tracer_top_fluxes = NamedTuple(name => name ∈ flux_tracers ? Field{Center, Center, Nothing}(grid) : ZeroField() for name in tracers)
```

src/Oceans/ocean_simulation.jl
simone-silvestri: with this change the freshwater comes in at zero tracer content. I guess if it is ok?

src/Atmospheres/interpolate_atmospheric_state.jl
```diff
@@ -127,12 +123,40 @@ end
...
+        name = names[n]
+        update_tracer_state!(i, j, surface_atmos_state[name], atmos_tracers[name], atmos_args, t_itp)
+    end
+
+    return nothing
+end
+
+@inline function update_tracer_state!(i, j, state, tracer, atmos_args, t_itp)
+    @inbounds state[i, j, 1] = interp_atmos_time_series(tracer, atmos_args...)
+
```
simone-silvestri:
```suggestion
```

src/Atmospheres/interpolate_atmospheric_state.jl
```diff
@@ -127,12 +123,40 @@ end
...
+    return nothing
+end
+
+@inline update_tracer_state!(i, j, state, ::ConstantField, atmos_args, t_itp) = nothing
+# `t_itp` comes from the atmosphere's time axis, so the tracer must share it.
+@inline function update_tracer_state!(i, j, state, zero_D_fts::FTS0, atmos_args, t_itp)
+    fi = FractionalIndices(nothing, nothing, nothing)
+    @inbounds state[1, 1, 1] = interpolate(fi, t_itp, zero_D_fts, zero_D_fts.backend, zero_D_fts.time_indexing)
+
```
simone-silvestri:
```suggestion
```

src/Atmospheres/interpolate_atmospheric_state.jl
```diff
@@ -127,12 +123,40 @@ end
...
+# on a literal symbol and the kernel stays statically resolvable on GPU.
+@inline function update_tracer_states!(i, j, surface_atmos_state, atmos_tracers::NamedTuple{names},
+                                       atmos_args, t_itp) where names
+
+    ntuple(Val(length(names))) do n
+        name = names[n]
+        update_tracer_state!(i, j, surface_atmos_state[name], atmos_tracers[name], atmos_args, t_itp)
+    end
+
+    return nothing
```
simone-silvestri:
```suggestion
@inline function update_tracer_states!(i, j, surface_atmos_state, atmos_tracers::NamedTuple{names}, atmos_args, t_itp) where names
    ntuple(Val(length(names))) do n
        name = names[n]
        update_tracer_state!(i, j, surface_atmos_state[name], atmos_tracers[name], atmos_args, t_itp)
    end
    return nothing
```

ext/NumericalEarthOceanBioMEExt/radiation.jl
```diff
@@ -0,0 +1,22 @@
...
+
+apply_air_sea_biogeochemical_radiative_fluxes!(coupled_model, biogeochemistry::DiscreteBiogeochemistry{<:NutrientsPlanktonDetritus}) = 
+    apply_air_sea_biogeochemical_radiative_fluxes!(coupled_model, biogeochemistry.light_attenuation)
+
+function apply_air_sea_biogeochemical_radiative_fluxes!(coupled_model, light::AbstractPhotosyntheticallyActiveRadiation{<:PARFromShortwave})
+    penetrating_shortwave = coupled_model.radiation.interface_fluxes.ocean.downwelling_shortwave
+    PAR = surface_PAR(light)
+
+    set!(PAR.surface_shortwave, penetrating_shortwave)
+
```
simone-silvestri:
```suggestion
    set!(PAR.surface_shortwave, penetrating_shortwave)
```

ext/NumericalEarthOceanBioMEExt/gas_exchange.jl
```diff
@@ -0,0 +1,151 @@
...
+
+    compute_gas_exchange!(i, j, grid, clock, biogeochemistry.oxygen, fluxes, ℵ, ocean_tracers, exchangers)
+    compute_gas_exchange!(i, j, grid, clock, biogeochemistry.inorganic_carbon, fluxes, ℵ, ocean_tracers, exchangers)
+end
+
+@inline compute_gas_exchange!(i, j, grid, clock, ::Nothing, fluxes, ℵ, ocean_tracers, exchangers) = nothing
+
+@inline function compute_gas_exchange!(i, j, grid, clock, ::Oxygen, fluxes, ℵ, ocean_tracers, exchangers)
+    @inbounds fluxes.O₂[i, j, 1] = exchangers.O₂(i, j, grid, clock, ocean_tracers) * (1 - ℵ)
+
```
simone-silvestri:
```suggestion
```

ext/NumericalEarthOceanBioMEExt/gas_exchange.jl
```diff
@@ -0,0 +1,151 @@
...
+    return nothing
+end
+
+@inline function compute_gas_exchange!(i, j, grid, clock, ::AbstractInorganicCarbon{1}, fluxes, ℵ, ocean_tracers, exchangers)
+    @inbounds fluxes.DIC[i, j, 1] = exchangers.DIC(i, j, grid, clock, ocean_tracers) * (1 - ℵ)
+
```
simone-silvestri:
```suggestion
```

ext/NumericalEarthOceanBioMEExt/gas_exchange.jl
```diff
@@ -0,0 +1,151 @@
...
+    return nothing
+end
+
+# `Symbol(:DIC, n)` can not be constructed on the GPU so the replicates are unrolled here
+@generated function compute_gas_exchange!(i, j, grid, clock, ::AbstractInorganicCarbon{N}, fluxes, ℵ, ocean_tracers, exchangers) where N
+    exprs = map(1:N) do n
+        DIC = Symbol(:DIC, n)
+
```
simone-silvestri:
```suggestion
```

ext/NumericalEarthOceanBioMEExt/gas_exchange.jl
```diff
@@ -0,0 +1,151 @@
...
+        :(@inbounds fluxes.$DIC[i, j, 1] = exchangers.$DIC(i, j, grid, clock, ocean_tracers) * (1 - ℵ))
+    end
+
+    return quote
+        $(exprs...)
+
+        return nothing
```
simone-silvestri, line 144: This cannot be written out explicitly?

ext/NumericalEarthOceanBioMEExt/gas_exchange.jl
```diff
@@ -0,0 +1,151 @@
...
+
+biogeochemical_interface(exchanger, ocean, ::AbstractInorganicCarbon{1}; kwargs...) =
+    (; DIC = carbon_dioxide_exchange(exchanger, :DIC, :Alk; kwargs...))
+
+function biogeochemical_interface(exchanger, ocean, ::AbstractInorganicCarbon{N}; kwargs...) where N
+    names = carbon_replicate_names(Val(N))
+
+    exchanges = ntuple(n -> carbon_dioxide_exchange(exchanger, Symbol(:DIC, n), Symbol(:Alk, n); kwargs...), Val(N))
+
+    return NamedTuple{names}(exchanges)
```
simone-silvestri, line 48:
```suggestion
    names = carbon_replicate_names(Val(N))
    exchanges = ntuple(n -> carbon_dioxide_exchange(exchanger, Symbol(:DIC, n), Symbol(:Alk, n); kwargs...), Val(N))
    return NamedTuple{names}(exchanges)
```

src/Oceans/ocean_simulation.jl
jagoosw: I think this is an okay default but will need to be configured. Maybe I should have it default to a function that I can extend in the extension.

ext/NumericalEarthOceanBioMEExt/gas_exchange.jl
jagoosw: I don't know a different way? We basically have a bunch of tracers called DIC1, DIC2, ..., DICN and Alk1, Alk2, ..., AlkN that get the same BGC tendencies but exchange with the atmosphere depending on the pairs chemistry, and e.g. some could get an alkalinity addition. Its a parameter of the model like InorganicCarbon{N} which is set by the user when the model is constructed

src/Atmospheres/interpolate_atmospheric_state.jl
jagoosw: Is that a problem for the parameter space issue?

jagoosw: I can write a generator function if thats the only way to conserve the parameter space but allow user specified fields?

jagoosw: So the version on main would be the default then we have a generator otherwise?

jagoosw:
is there a strategy to get other tests to run when the download tests fail?

jagoosw:
The problems were from data downloading again now, is there a way to fix them?

jagoosw:
@simone-silvestri I think this PR is ready now, the testset I added is passing and as far as I can see the other testsets are fine too except something about data downloading:
```
ERROR: LoadError: The offline test run downloaded data: /tmp/jl_ZMi1Ud/ETOPO/ETOPO_2022_v1_60s_N90W180_surface.nc, /tmp/jl_ZMi1Ud/JRA55/RYF.friver.1990_1991.nc, /tmp/jl_ZMi1Ud/JRA55/RYF.huss.1990_1991.nc, /tmp/jl_ZMi1Ud/JRA55/RYF.licalvf.1990_1991.nc, /tmp/jl_ZMi1Ud/JRA55/RYF.prra.1990_1991.nc, /tmp/jl_ZMi1Ud/JRA55/RYF.prsn.1990_1991.nc, /tmp/jl_ZMi1Ud/JRA55/RYF.psl.1990_1991.nc, /tmp/jl_ZMi1Ud/JRA55/RYF.rlds.1990_1991.nc, /tmp/jl_ZMi1Ud/JRA55/RYF.rsds.1990_1991.nc, /tmp/jl_ZMi1Ud/JRA55/RYF.tas.1990_1991.nc, /tmp/jl_ZMi1Ud/JRA55/RYF.uas.1990_1991.nc, /tmp/jl_ZMi1Ud/JRA55/RYF.vas.1990_1991.nc
```
I'm not sure how to fix these tests. Could we merge it soon?

simone-silvestri:
Ah, now the tests should not download external data, they should use synthetic data when testing

jagoosw:
> Ah, now the tests should not download external data, they should use synthetic data when testing

I've set them to run again, for the docs there is a problem with the Veros extension because its not finding the CubicSplines package. I checked and I haven't touched anything like that in the Project.toml?

simone-silvestri (approved):

jagoosw:
@simone-silvestri in fixing the type stability for the atmosphere stuff I/claude found a problem with interpolating x-only or y-only FTSs, and Face located ones (which #700 uses because they provide data from -90:1:90. Are you still happy with these changes? Part of this also meant using each tracers own grid and times

PR #575, opened 2026-08-24
Publish the Breeze RTM's surface downwelling fluxes to the interface radiation state

xkykai:
Under a `RadiativeTransferModel`, the `AirLandRadiationState` every surface formulation reads was identically `(0, 0, 0, 0, 0)`: `ComponentExchanger(::BreezeRTM, grid)` returned `nothing` and `kernel_radiation_properties(::BreezeRTM)` an empty `surface_properties`. The RTM now fills the same contract a `PrescribedRadiation` does.

- `ComponentExchanger(::BreezeRTM, grid)` holds `ℐꜜˢʷ`/`ℐꜜˡʷ` on the exchange grid
- `interpolate_state!` copies the RTM's surface faces across each step, negated: Breeze stores fluxes positive-up, the interface contract wants positive-down magnitudes
- `kernel_radiation_properties(::BreezeRTM)` returns σ and the RTM's surface albedo and emissivity as the land's `SurfaceRadiationProperties`
- `apply_air_land_radiative_fluxes!` reads the published downwelling rather than the RTM's own fields, so one representation feeds both paths


simone-silvestri:
> The formulations that do read it are not here.

Does this mean we need to thread radiation through the interface solver? I think this is done for the ice and for `SkinTemperature`

xkykai:
> > The formulations that do read it are not here.
> 
> Does this mean we need to thread radiation through the interface solver? I think this is done for the ice and for `SkinTemperature`

Land is already threading radiation through the interface solver where it is required, so this comment is Claude's fluff. I've fixed the PR body

xkykai:
@kaiyuan-cheng do we need this for the Breeze radiation--land coupling?

kaiyuan-cheng:
> do we need this for the Breeze radiation--land coupling?

The existing radiation-land coupling is working correctly, but I think this PR makes the coupling more robust. I will take a closer look later today.

test/test_breeze_coupling.jl
```diff
@@ -250,8 +251,8 @@ end
 
         expected_surface_energy(temperature) =
             initial_surface_energy_flux +
-            emissivity * stefan_boltzmann_constant * temperature^4 +
-            emissivity * downwelling_longwave +
+            emissivity * stefan_boltzmann_constant * temperature^4 -
+            emissivity * downwelling_longwave -
             (1 - albedo) * downwelling_shortwave
```
glwagner, line 256: consider using normal notation for this?

PR #576, opened 2026-08-25
New rules to make the generated code more sustainable

simone-silvestri:
Recently claude coding has become quite the standard for NumericalEarth. This is not a negative, it allows to accelerate writing code and NumericalEarth has grown quite quickly thanks to it. What I am afraid of, however, is the sustainability of the codebase. I feel like generated code it quite obscure to a human reviewer because of several reasons: (1) the excess of prose which, counterintuitively hampers readability, (2) comments that refer to different part of the codebase, very useful for a machine, not so much for a human reader, (3) the insane amount of bespoke features that a machine introduces rather than finding a more general design or a lower-impact change in the code that would result in the same feature implementation, (4) the thousands of lines of code written for implementing rather simple changes (maybe same as number 3), (5) the excessive input validation which now makes up most of the codebase.

To be practical and quantitative, the codebase grew from 10k lines in January to 26k lines now with the share of comments increasing from ~8.5% to ~13%.

For this reason, I tried to synthesize this in a "restriction" rule that tries to make the code more sustainable and human-readable.

### DIFF
--- a/.claude/rules/restraint-rules.md
+++ b/.claude/rules/restraint-rules.md
@@ -18,6 +18,11 @@ effect**, and they outrank any instinct toward completeness.
 > **The test behind every rule below:** could a reviewer who knows the physics read this hunk once, in order,
 > and say yes?
 
+**They apply to existing code too.** Every rule below is also a license to delete: when you are already editing
+a file and a rule catches something that is *already there*, removing it is in scope, not scope creep. What is
+out of scope is going looking — do not open unrelated files to hunt for violations, and keep a cleanup that
+grows past the change it started from to its own PR.
+
 ## Rule 1 — The feature is the diff; everything else is scope creep
 
 Before opening a PR, ask what the smallest diff is that delivers the stated change. That diff is the PR.
@@ -66,7 +71,10 @@ A **loud failure is better than a silent fallback.** A `NaN` that reaches the co
 first output; a roughness silently floored at `1e-5` produces plausible wrong fluxes forever.
 
 - If a guard's own docstring has to say *"this is a guard, not a physical parameter"* — delete the guard.
-- Never grow a struct, a keyword argument, or a `show` line to hold a value that is not physics.
+- A keyword argument must express **physics or numerics** — a rate, a length, a tolerance, an iteration
+  cap, a search radius. What it must not express is a defense: a floor, a clamp, or a fallback whose
+  only job is to keep malformed input from reaching the solver. Never grow a struct or a `show` line
+  for one either.
 
 ❌ `minimum_roughness_length` — a new public keyword, struct field, `FT` conversion and `show` row, existing
    only so a field mutated after validation cannot produce `NaN`.
@@ -228,6 +236,7 @@ Any **yes** means cut before presenting.
 9. Is any error message longer than one sentence?
 10. Does any test pin an internal helper, a field order, or a fallback value?
 11. Does any identifier carry an invented sub/superscript, or mix math with English?
-12. Is there a new public keyword argument that the docs describe as non-physical?
+12. Is there a new public keyword argument that is neither physics nor numerics — a floor, clamp, or
+    fallback guarding against malformed input?
 13. Does the diff contain anything a reader would not expect from the title — including another
     branch merged into the head?
--- a/.claude/rules/restraint-rules.md
+++ b/.claude/rules/restraint-rules.md
@@ -216,7 +216,7 @@ The protocol:
    *"Self-review: cut the `minimum_roughness_length` guard (Q3) and two forwarding methods (Q5); 41 lines
    removed."* If it caught nothing, say *"Self-review: clean."* Never claim clean without having read the diff.
 5. **Never delete a rule to pass the checklist.** If a rule seems wrong for the case at hand, say so to the PR author
-   and let him decide. Editing `restraint-rules.md` to accommodate your own diff is the one forbidden move.
+   and let them decide. Editing `restraint-rules.md` to accommodate your own diff is the one forbidden move.
 
 For a three-line edit this takes seconds. For a large one it is the difference between review and no review.
 
--- a/.claude/rules/restraint-rules.md
+++ b/.claude/rules/restraint-rules.md
@@ -215,7 +215,7 @@ The protocol:
 4. **Report the result in one line.** State that the checklist ran and what it caught, e.g.
    *"Self-review: cut the `minimum_roughness_length` guard (Q3) and two forwarding methods (Q5); 41 lines
    removed."* If it caught nothing, say *"Self-review: clean."* Never claim clean without having read the diff.
-5. **Never delete a rule to pass the checklist.** If a rule seems wrong for the case at hand, say so to Simone
+5. **Never delete a rule to pass the checklist.** If a rule seems wrong for the case at hand, say so to the PR author
    and let him decide. Editing `restraint-rules.md` to accommodate your own diff is the one forbidden move.
 
 For a three-line edit this takes seconds. For a large one it is the difference between review and no review.
--- a/.claude/rules/restraint-rules.md
+++ b/.claude/rules/restraint-rules.md
@@ -215,6 +215,23 @@ number.
    `interface.flux_formulation.zero_plane_displacement[2, 1, 1] == 4.0`
 ✅ A doctest that shows the object, via `show`/`summary`, or nothing at all.
 
+## Rule 13 — Search before you write; do not reinvent what exists
+
+Before adding a function, search Oceananigans and NumericalEarth for one that already does the job. Reuse it,
+extend it, or dispatch on it. A second implementation of an existing capability is the most expensive kind of
+scope creep: the duplicate is correct on the day it lands and diverges from the original forever after.
+
+- If an existing function is close but not exact, **widen it** — a new method, a keyword argument, a relaxed
+  type annotation. A near-copy under a new name is never the answer.
+- If you cannot find it, say where you looked. "I searched `Oceananigans/src/Fields` and `DataWrangling` for a
+  masked fill and found none" is a reviewable claim; silence is not.
+- The same holds for patterns, not just functions: grid construction, the `launch!` idiom, `Adapt` rules,
+  `show` methods, dataset hooks — follow the way the codebase already does it.
+
+❌ A new `fill_gaps!` writing a fresh flood-fill kernel while `inpaint_mask!` sits in `DataWrangling`.
+❌ A hand-rolled bilinear interpolation in an example, next to `Oceananigans.Fields.interpolate`.
+✅ `NumericalEarth.DataWrangling.inpaint_mask!(field, mask)`, or a new method on it.
+
 ## Mandatory self-review
 
 **Claude: run the checklist below over every change you make, before you present it.** Not at PR time — at
@@ -224,7 +241,7 @@ The protocol:
 
 1. **Read your own diff.** `git diff` (or `git diff --stat` first if it is large). Review what you actually
    wrote, not what you remember intending. The checklist is answered from the diff, never from memory.
-2. **Answer all thirteen questions.** Every **yes** is a defect you introduced. There is no "yes, but it is
+2. **Answer all fourteen questions.** Every **yes** is a defect you introduced. There is no "yes, but it is
    justified here" — that judgment is exactly the one that produced the drift.
 3. **Cut before presenting.** Removing your own scaffolding is the default action and needs no permission.
    Do not ask "should I remove the validation?" — remove it, and say that you did.
@@ -256,3 +273,5 @@ Any **yes** means cut before presenting.
     fallback guarding against malformed input?
 13. Does the diff contain anything a reader would not expect from the title — including another
     branch merged into the head?
+14. Does the diff reimplement something that already exists in Oceananigans or NumericalEarth, or add a
+    new name where a new method on an existing function would do?
--- a/.claude/rules/restraint-rules.md
+++ b/.claude/rules/restraint-rules.md
@@ -30,6 +30,15 @@ and belongs in a separate PR, or nowhere.
 - Adding lines to `src/` is a cost paid forever. The bar is not "is this useful?" but "is the codebase better
   off carrying this than not?"
 
+**A PR must show only its own change.** Reviewability is a property of the diff GitHub renders, not of the
+commits you know are yours. Never open a PR whose head has merged another feature branch into it: base the PR
+on that branch instead, so the diff shows your commits alone. If the base has to be a feature branch, keep it
+one level deep and rebase — never merge sideways.
+
+❌ #570: a 5-line forwarding method and an 11-line test, presented as **+6186 / −1926 across 119 files**,
+   because the head merged `xk/slab-canopy` (50 commits of unrelated canopy work) into itself.
+✅ Base the PR on `xk/slab-canopy` directly. Diff renders 16 lines. Review takes one minute.
+
 ❌ Feature lands in commit 1 (+142/−143); commits 4 and 7 add +680 lines of validation nobody asked for.
 ✅ Feature lands in commit 1. PR ends. Validation, if genuinely needed, is proposed separately on its merits.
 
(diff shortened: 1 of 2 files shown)


.claude/rules/restraint-rules.md
```diff
@@ -0,0 +1,188 @@
...
+
+Never add machinery to defend against a state that is unreachable through the public API. "A user might mutate
+this field after construction" and "a downstream package might pass a wrong type" are hypotheticals, not
+requirements. Code written against hypotheticals is permanent; the hypothetical usually never arrives.
+
+A **loud failure is better than a silent fallback.** A `NaN` that reaches the coupled state is visible in the
+first output; a roughness silently floored at `1e-5` produces plausible wrong fluxes forever.
+
+- If a guard's own docstring has to say *"this is a guard, not a physical parameter"* — delete the guard.
+- Never grow a struct, a keyword argument, or a `show` line to hold a value that is not physics.
```
navidcy: but sometimes the implementation introduces a kwarg that is not physics but does make sense though, right?

.claude/rules/restraint-rules.md
```diff
@@ -0,0 +1,188 @@
...
+
+❌ `minimum_roughness_length` — a new public keyword, struct field, `FT` conversion and `show` row, existing
+   only so a field mutated after validation cannot produce `NaN`.
+✅ Validate at construction, or let it fail loudly. No new API.
+
+## Rule 4 — Don't invent an API to solve a problem the same PR created
```
navidcy, line 91: omg, that's so true!

navidcy (approved):

.claude/rules/restraint-rules.md
simone-silvestri: I guess physics / numerics? I can rephrase this

navidcy:
does it worth making Claude go through the `.md` file and apply it over the whole codebase with this in mind and open a new PR just to see if it captures changes we are happy with or we need to tweak this ruleset?

I guess the way the rule is phrased it's only applicable to a new PR?

.claude/rules/restraint-rules.md
```diff
@@ -0,0 +1,188 @@
...
+effect**, and they outrank any instinct toward completeness.
+
+> **The test behind every rule below:** could a reviewer who knows the physics read this hunk once, in order,
+> and say yes?
+
+## Rule 1 — The feature is the diff; everything else is scope creep
+
+Before opening a PR, ask what the smallest diff is that delivers the stated change. That diff is the PR.
+Anything that is not the feature — validation, fallbacks, gap-fillers, helper APIs — is a **separate concern**
+and belongs in a separate PR, or nowhere.
```
simone-silvestri: I am also not super sure of this maybe I relax it a bit? Validation does not really belong in another PR, but the over-validation claude does is definitely overwhelming

.claude/rules/restraint-rules.md
```diff
@@ -0,0 +1,188 @@
...
+- teach the language (`==` vs `===`, how `@inbounds` works, why `Adapt` needs field order);
+- carry a multi-line design memo or `TODO` about a different module's construction order.
+
+If the reasoning genuinely cannot be dropped, it belongs in the PR description — which reviewers read once and
+which does not age in the source tree.
+
+❌ Five lines above `Adapt.@adapt_structure SimilarityTheoryFluxes` explaining that a positional rebuild
+   elsewhere would mis-wire the struct and that a test pins the field order.
+❌ `# Grids are compared with `==` … rather than `===`: two references to one grid are not guaranteed to be egal`
+✅ `# ifelse, not ?:, so the kernel stays branch-free on GPU`
```
simone-silvestri: Also this is not worth a comment probably

simone-silvestri:
> does it worth making Claude go through the .md file and apply it over the whole codebase with this in mind and open a new PR just to see if it captures changes we are happy with or we need to tweak this ruleset?

We can try to make a wide pass and see what comes out

simone-silvestri:
@glwagner @xkykai @giordano what do you think? Something to add / remove?

.claude/rules/restraint-rules.md
```diff
@@ -0,0 +1,242 @@
...
+1. **Read your own diff.** `git diff` (or `git diff --stat` first if it is large). Review what you actually
+   wrote, not what you remember intending. The checklist is answered from the diff, never from memory.
+2. **Answer all thirteen questions.** Every **yes** is a defect you introduced. There is no "yes, but it is
+   justified here" — that judgment is exactly the one that produced the drift.
+3. **Cut before presenting.** Removing your own scaffolding is the default action and needs no permission.
+   Do not ask "should I remove the validation?" — remove it, and say that you did.
+4. **Report the result in one line.** State that the checklist ran and what it caught, e.g.
+   *"Self-review: cut the `minimum_roughness_length` guard (Q3) and two forwarding methods (Q5); 41 lines
+   removed."* If it caught nothing, say *"Self-review: clean."* Never claim clean without having read the diff.
+5. **Never delete a rule to pass the checklist.** If a rule seems wrong for the case at hand, say so to Simone
```
xkykai:
```suggestion
5. **Never delete a rule to pass the checklist.** If a rule seems wrong for the case at hand, say so to the PR author
```
Can't wait for all Claude agents to flood Simone's inbox

.claude/rules/restraint-rules.md
```diff
@@ -0,0 +1,242 @@
...
+   and let him decide. Editing `restraint-rules.md` to accommodate your own diff is the one forbidden move.
```
xkykai:
```suggestion
   and let them decide. Editing `restraint-rules.md` to accommodate your own diff is the one forbidden move.
```

xkykai (approved):

xkykai:
I wonder if it is useful for human reviewers to fill out a simple checklist of the most difficult parts of Claude coding to steer, doesn't have to be too complex but perhaps three points like:

1. Is the code understandable?
2. Does it reinvent methods that are/should be reused?
3. 

Just as a sounding board. Perhaps we can add our most common annoyances when using Claude. Then all the bullet points have to be checked before it counts as an approval. These points also shouldn't be too difficult to check though, and importantly has to be subjective and related to taste/design/judgement that AI agents still can't do quite well

xkykai:
but for now we can try this out to see how effective it is!

.claude/rules/restraint-rules.md
simone-silvestri: ahahah

xkykai:
If it's helpful about PR Issue verbosity here's what I have been experimenting with in `CLAUDE.md`:

## Issues and PRs
- Keep issues and PRs as concise as possible. Cut to essentials: a 1–2 sentence summary, tight bullets for the public API, and at most one minimal MWE. Omit validation prose, full downstream wiring, and multi-section scaffolding unless asked — aim for a body readable in ~20 seconds.
- Keep issue/PR descriptions developer/user-facing: describe the change in terms of the code and its behavior, not the internal process. No references to internal planning artifacts (e.g. "plan 08b", "phase 2", task numbers). This applies equally to commit messages and code comments — and note a bare `#2` also renders as a link to an unrelated GitHub issue/PR.
- When writing issues, ideally they should come with an MWE which is as minimal (involves as few components as possible), simple, and brief as possible.

simone-silvestri:
> I wonder if it is useful for human reviewers to fill out a simple checklist of the most difficult parts of Claude coding to steer, doesn't have to be too complex but perhaps three points like:

I like the idea, where should we put it, in AGENTS.md?

xkykai:
> > I wonder if it is useful for human reviewers to fill out a simple checklist of the most difficult parts of Claude coding to steer, doesn't have to be too complex but perhaps three points like:
> 
> I like the idea, where should we put it, in AGENTS.md?

My intention is that it is part of the checks a reviewer has to go through before giving an approval, so not `AGENTS.md`. we could perhaps put it into the ruleset and github workflow. Whether it should be a required checklist before approval could be granted is up to folks I guess, maybe we can start without enforcing this requirement but as a strong encouragement. 

@giordano might have tricks up his sleeve to implement this in a beautiful manner.

.claude/rules/restraint-rules.md
```diff
@@ -0,0 +1,242 @@
...
+✅ Base the PR on `xk/slab-canopy` directly. Diff renders 16 lines. Review takes one minute.
+
+❌ Feature lands in commit 1 (+142/−143); commits 4 and 7 add +680 lines of validation nobody asked for.
+✅ Feature lands in commit 1. PR ends. Validation, if genuinely needed, is proposed separately on its merits.
+
+## Rule 2 — One invariant, one mechanism
+
+An invariant is enforced in exactly one place. Enforcing it at setup *and* in the kernel *and* through a
+fallback *and* in a gap-filler does not make it four times as safe — it makes it impossible to tell which
+mechanism is load-bearing, and guarantees the four drift apart.
```
giordano: "load-bearing" 😏

.claude/rules/restraint-rules.md
```diff
@@ -0,0 +1,242 @@
...
+❌ `evaluable_roughness_length` checked in `validate_flux_formulation`, again in `local_roughness_length`,
+   again in `guard_local_roughness_length`, again in `fill_aerodynamic_roughness_gaps!` — while
+   `displaced_profile_height`'s `max(Δh - d, 2ℓ)` already floored it before the PR.
+✅ One check, at the layer that owns the contract. State in one line where that is.
+
+## Rule 3 — No guards for states the code cannot reach
+
+Never add machinery to defend against a state that is unreachable through the public API. "A user might mutate
+this field after construction" and "a downstream package might pass a wrong type" are hypotheticals, not
+requirements. Code written against hypotheticals is permanent; the hypothetical usually never arrives.
```
giordano: Honestly I don't understand exactly what this means.

.claude/rules/restraint-rules.md
```diff
@@ -0,0 +1,242 @@
...
+- teach the language (`==` vs `===`, how `@inbounds` works, why `Adapt` needs field order);
+- carry a multi-line design memo or `TODO` about a different module's construction order.
+
+If the reasoning genuinely cannot be dropped, it belongs in the PR description — which reviewers read once and
+which does not age in the source tree.
+
+❌ Five lines above `Adapt.@adapt_structure SimilarityTheoryFluxes` explaining that a positional rebuild
+   elsewhere would mis-wire the struct and that a test pins the field order.
+❌ `# Grids are compared with `==` … rather than `===`: two references to one grid are not guaranteed to be egal`
+✅ `# ifelse, not ?:, so the kernel stays branch-free on GPU`
```
giordano, line 159: This is something I hate in LLM-generated code: they often describe the previous state of the code, which however isn't there anymore, so that's incredibly confusing for a future reader.  That belongs to the git commit message (which has the context of the change before vs after), not to a code comment (which only has the new code).  Recently I came across a [docstring which described what the code doesn't do](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/fade415f2337874ff75695f516951edfaec08e07/src/ArchiveUtils.jl#L29-L32), which was a bit irritating.

.claude/rules/restraint-rules.md
```diff
@@ -0,0 +1,242 @@
...
+
+❌ `"$name has $(count(...)) cells that are not $requirement (minimum $(minimum(values))). A bad cell either
+   propagates NaN through u★, θ★ and q★ into the coupled state or silently zeroes every turbulent flux there.
+   Fill the gaps first, e.g. with fill_aerodynamic_roughness_gaps!."`
+✅ `"momentum_roughness_length must be finite and positive, found $(minimum(values))"`
+
+## Rule 10 — Test behavior, not scaffolding
+
+Tests pin the physics and the public contract. A test that asserts an internal helper's return type, a struct's
+field order, or the exact value a fallback produces makes the scaffolding **load-bearing**: it can no longer be
```
giordano: "load-bearing" No. 2, sigh.

giordano (commented):
I think the intention is good and generally agree with most of the directives, but it's not like these tools do whatever you tell them in these files, they'll pick some of the bits anyway, giving a large set of instructions only increases the chances that something will be missed.

> I like the idea, where should we put it, in AGENTS.md?

People often put this type of checklists in [pull request templates](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository), although those are more commonly for PR authors rather than reviewers, but if you want to enforce them that's one way to do it.

xkykai:
> > > I wonder if it is useful for human reviewers to fill out a simple checklist of the most difficult parts of Claude coding to steer, doesn't have to be too complex but perhaps three points like:
> > 
> > 
> > I like the idea, where should we put it, in AGENTS.md?
> 
> My intention is that it is part of the checks a reviewer has to go through before giving an approval, so not `AGENTS.md`. we could perhaps put it into the ruleset and github workflow. Whether it should be a required checklist before approval could be granted is up to folks I guess, maybe we can start without enforcing this requirement but as a strong encouragement.
> 
> @giordano might have tricks up his sleeve to implement this in a beautiful manner.

also I am messing around with how that might look like in https://github.com/NumericalEarth/NumericalEarth.jl/pull/581, feel free to comment/edit/close as folks see fit. This sets up a github-actions bot to post it automatically and monitor when the comment is getting edited.

The items on the checklist are only placeholders, so if people want to do this we can then discuss what should go on the checklist.

.claude/rules/restraint-rules.md
navidcy: hahahahhahahahahhaha

.claude/rules/restraint-rules.md
navidcy: Sometimes agents try to failproof against cases that are 1 in a gazillion. Like "let's write an error warning if the user provides a plot type for the value of diffusivity". I think that's what this is trying to capture. Right?

giordano: I once saw this: https://github.com/JuliaTesting/ParallelTestRunner.jl/pull/125#issuecomment-4230145985.  The inner function has a (technically correct) error path which can never be reached when that's called by the outer function, which is what the users are meant to use.  I agree it's excessive (users aren't meant to use the inner function), but honestly I believe this is the least problematic behaviour, although it does add bloat.

xkykai:
Another thing I noticed that Claude likes to do is to reinvent functions which has been implemented elsewhere. Sometimes I add something like

> As much as possible, you should reuse/extend/follow existing patterns and methods in Oceananigans/NumericalEarth. Do not reinvent the wheel.

simone-silvestri:
> Another thing I noticed that Claude likes to do is to reinvent functions which has been implemented elsewhere. Sometimes I add something like

Added a new rule for it

> they'll pick some of the bits anyway, giving a large set of instructions only increases the chances that something will be missed.

you are right, I asked claude how it operates and how to minimize the loss. This is the answer:

## How I actually operate, concretely:

AGENTS.md and matched .claude/rules/*.md are injected as text near the start of the session. From then on they are static, competing with everything that arrives after — tool results, file contents, your messages. Twenty tool calls into a kernel debug, a rule I read at token 3k has far less pull on the next token than the stack trace I read thirty seconds ago. Nothing re-reads the file for me; it just recedes.

What survives that recession, in order of strength:

1. Procedures with a reportable output. "Run git diff, answer the questions, print one line saying what it caught." That is an action with an artifact you can check. If I don't print the line, you know I skipped it. This is the single most effective thing in the file, and it's doing more work than all thirteen rules above it.
2. Concrete ❌/✅ pairs from this repo. I pattern-match much harder than I reason from principle. minimum_roughness_length, #570's merged head, the nine one-liner clamps — those fire. The paragraph explaining why fires much less.
3. Bright lines. "Error messages are one sentence" survives; "keep error messages concise" does not. A threshold I can check against my own output beats a disposition I'm supposed to hold.
4. Everything else — the explanatory prose, the mid-document rationale, the "the test behind every rule" framing — is where the loss concentrates. It's what I'm least likely to act on and the bulk of the 258 lines.

So yes, compact — but cut the essay, not the checklist

My concrete proposal: split the file in two.

- .claude/rules/restraint-rules.md → the 14 questions + one ❌/✅ line each, ~45 lines. This is what loads on every src/ edit. It's the part that changes behavior.
- The essay moves out — to docs/, the PR description, or a wiki page. The case studies (#570, #523, the roughness-length PR) are genuinely valuable for humans deciding whether the ruleset is right. They're near-worthless as per-edit context, and they're most of the weight.

That answers Giordano directly: loaded surface drops ~80%, and the reasoning isn't lost, it just stops competing for attention on every file touch.

Rule 13 that I just added would collapse to two lines under this scheme — which is a fair thing to point at, since I just grew the file you're asking whether to shrink.

The part compaction won't fix

Most of these rules are judgment calls. "Is this guard unreachable through the public API?" needs me to reason about call sites; "is this scope creep?" needs me to know what you meant by the title. Compaction improves retention, not judgment — a 45-line version will still miss those.

What doesn't degrade at all is mechanical enforcement. From the main-branch scan we did earlier, the drift that was actually real and actually measurable was PR size and comment density — not the line-level findings, which ran ~85% false positives. Both of those are CI checks, not prose. A bot that flags "+700 lines, feature commit is +142" or "comment-to-code ratio 3× the repo median" enforces Rule 1 and Rule 7 with zero attention budget and no session-length decay. Same for xkykai's #581 reviewer checklist: a posted comment a human ticks is immune to everything I described above.

Caveat worth stating plainly: I'm not a reliable narrator of my own instruction-following. The above is inference from the same machinery that produces the behavior, so treat it as a hypothesis, not a readout. The wide pass navidcy proposed and you agreed to is the actual test — and it's cheap to make it discriminating: run it once with the full 258-line file, once with a checklist-only version, and compare what each catches. That settles Giordano's question with evidence instead of my introspection.

.claude/rules/restraint-rules.md
ewquon: 100% agree with this sentiment. LLMs do bake in a lot of history and justification into their coding comments.

Something I can appreciate though is a _concise_ note when a piece of code appears to be doing something unusual or is following some unorthodox implementation pattern, but for a defensible reason. It's hard to find the sweet spot for verbosity though. Still being a relatively new developer in this codebase, these super verbose comments have been helpful at times for review — I've tried to be better about reading, digesting, and then pruning the comments afterward.

In the spirit of preserving the before/after context in the commits, can we add to the rules something along the lines of: "make atomic commits" and "prefer adding context to commit messages rather than code comments"?

simone-silvestri:
I ll merge this so it starts taking partial effect. We can figure out if it is useful and how to modify it with use and how to connect it to #581

PR #577, opened 2026-08-25
Some cleanup of drifted code

simone-silvestri:
This is not necessarily cleanup of just AI code, but also of human code that drifted a bit in time, also following what written in PR #576. This is also fully AI driven, but no functionality nor API should change from this PR. This is the claude-generated description of this PR (always very verbose, uff)

 ## Cleanup: dead code, duplicated hooks, two latent bugs

  Cleanup pass over `src/` and `test/` following `.claude/rules/restraint-rules.md`. No new features.
  **50 files, +312 / −937 — net −625 lines.**

  ### Removed as dead

  - `src/DataWrangling/ERA5/ERA5_variables.jl` (98 lines) — its four definitions referenced only each
    other. The one external mention was a comment falsely claiming the CDS extension exported them.
  - `src/InitialConditions/` — the whole module: included, `using`'d, exported nothing, called nowhere.
    Two latent bugs went with it (`construct_grid` used `extent[2]` for `z`; `_apply_tracer_mask!`'s
    `Function` method referenced an out-of-scope `grid`).
  - Eight unreferenced definitions: `fractional_index_type`, `cv_d`, `variable_is_three_dimensional`,
    `JRA55NetCDFFTS`, `JRA55_time_indices`, `ORCAOneMetadatum`, `ORCAQuarterMetadatum`, `default_clock`.

  ### Duplication collapsed

  - **`inpainted_metadata_filename` was seven module-local functions sharing a name**, each paired with an
    identical one-line `inpainted_metadata_path` method. Now one hook with a default:
    9 path methods → 1, 8 generics → 2. Filenames verified byte-identical.
  - **`AtmosphereSeaIceFluxes` merged into `AtmosphereSurfaceFluxes`** — identical apart from three
    write-only diagnostic fields, and it had one call site.
  - **The three atmosphere–surface flux kernels** shared 49 character-identical lines (over half of each);
    the common read/assemble/store moved to `atmosphere_interface_kernels.jl`, per-surface physics stays
    in each kernel.
  - **`ERA5PrescribedRadiation` / `JRA55PrescribedRadiation`** (0.84 similar) → two-line wrappers over a
    shared `prescribed_radiation`, using the existing `default_download_directory` hook.
  - `_compute_radiation_fractional_indices!` was a copy of `_compute_fractional_indices!` minus one branch.
  - `apply_height_distribution`'s 5-arg method repeated the 7-arg body; it now forwards.

  ### Bugs fixed

[...]

PR #578, opened 2026-08-25
Remove deprecated InitialConditions module

ewquon:


simone-silvestri:
the removal of the module should be addressed in #577

simone-silvestri (approved):

ewquon:
I wanted to suggest an edit on #577 and thought this would be simple but this new "stack" feature seems to have made a mess of things.

PR #585, opened 2026-08-25
Read OpenLandMap COG windows at a target-matched resolution

xkykai:
`Field(metadatum, grid)` materialized the full 30 m OpenLandMap window before regridding — a
CONUS-sized box is ~150 GB of `Float32` per depth layer. `OpenLandMapSoilDB(aggregation_factor = nothing)`
instead reads the coarsest lattice that still oversamples the target twofold, served from the
cloud-optimized GeoTIFFs' average-resampled overview pyramid: the CONUS window matched to a 12 km
grid is 2297 × 1016 cells rather than 232000 × 100000, and reads in 15 s. Averaging suits
continuous fields only, and it changes the values a user gets, so it is opt-in.

- `DataWrangling.coarsest_resolving_dataset(dataset, grid)` returns the coarsest variant of
  `dataset` whose cells still resolve `grid`, or `dataset` itself by default. `Field(metadatum, grid)`
  and `set!(field, metadatum)` rebuild the metadatum on it, so the cached filename follows the
  realized lattice.
- `OpenLandMapSoilDB(; aggregation_factor = 1)`: an integer pins the read at that many 30 m
  pixels per cell side, `nothing` sizes the read to the target, and `Field(metadatum)`, which has
  no target, reads at full resolution either way.
- The reader writes one depth layer at a time, so peak memory is one window rather than all three.

```julia
grid = LatitudeLongitudeGrid(size = (600, 270, 3), longitude = (-134, -61), latitude = (21, 53),
                             z = [-1.0, -0.6, -0.3, 0.0])
dataset = OpenLandMapSoilDB(aggregation_factor = nothing)

sand = Field(Metadatum(:sand_fraction; dataset, region = BoundingBox(grid)), grid)
```


xkykai:
Folded back into #566, which carries the review thread. Closing in favor of it.

glwagner:
lots of conflicts!

PR #586, opened 2026-08-25
Regrid dataset fields in tiles so the native window need not be resident

xkykai:
`Field(metadatum, grid)` materialized the whole native window and a field on it before
interpolating, so peak memory scaled with the dataset's native resolution rather than with the
model domain. A regional read now fills the target one tile at a time, reading only the window
each tile covers. On the 30 m Southern Great Plains window (8002 × 8002 × 3) the regrid's memory
growth falls from 5470 MB to 598 MB, and it is no slower.

Each tile's source is a windowed `Field` over the native grid, sharing the grid object, so every
tile interpolates from the same node coordinates the whole field carries: the result is bitwise
identical to an untiled regrid at any tiling, measured 0.0 across 1, 4, 9, 36 and 144 tiles on
the CONUS window, masks included.

The machinery is all in the new `src/DataWrangling/tiled_regridding.jl`; `metadata_field.jl` only
routes `Field(metadatum, grid)` and `set!` through `regrid_from_metadata!`.

- `DataWrangling.retrieve_window(metadata, longitude_indices, latitude_indices)` with the
  `windowed_retrieval(dataset)` trait. It serves the hyperslab of any plain lon/lat NetCDF,
  north-first files included, so a dataset opts in with one line; `OpenLandMapSoilDB` and `GHSL`
  do. `Colon` indices span the file, and `retrieve_data` is now `retrieve_window(metadata, :, :)`,
  so the orientation flips live in one reader rather than two.
- `tile_bytes` bounds one tile's source; `set_region_data!` takes the region offset and kernel
  parameters to use, which is what lets a windowed field be filled over its own indices.

Tiling is declined where it would not be equivalent: a dataset that cannot read windows,
inpainting, an unbounded region, a periodic native window, or a target without one-dimensional
lon/lat axes.

Independent of #585, which bounds the bytes read rather than what stays resident. Both touch
`Field(metadatum, grid)` and `set!`, so whichever lands second needs a small rebase.


xkykai:
Folded back into #566, which carries the review thread. Closing in favor of it.

src/DataWrangling/set_region_data.jl
```diff
@@ -220,15 +220,18 @@ applying mangling, NaN conversion, and unit conversion in a single GPU-friendly
...
 
-    region      = region_info(metadata.region, target, λc, φc)
     FT          = eltype(target)
     grid        = target.grid
     arch        = architecture(grid)
     data        = architecture_ready(arch, data)
     missing_val = missing_value(metadata)
-    launch!(arch, grid, :xyz, _set_region_kernel!, interior(target), data, region, mangling, conversion, missing_val, FT)
+    # `target` rather than `interior(target)`: the two index identically over the interior, and a
+    # windowed field is then filled over its own indices.
```
simone-silvestri:
```suggestion
    # With `parameters`, a windowed field is then filled over its own indices.
```

src/DataWrangling/tiled_regridding.jl
```diff
@@ -0,0 +1,212 @@
...
+        return (data, λ[longitude_indices], φ[latitude_indices])
+    end
+end
+
+#####
+##### Tiled regridding — bounding what is resident by the tile, not by the window
+#####
+
+# Source data one tile may hold.
+const default_tile_bytes = 64 * 1024^2
```
simone-silvestri, line 68: Is this set by the available RAM?

simone-silvestri:
To reduce the scope creep and the duplication of methods, I think in this PR you can set 
```julia
function retrieve_data(metadata::Metadatum)
    data, _, _ = retrieve_window(metadata, :, :)
    return is_three_dimensional(metadata) ? data : dropdims(data, dims=3)
end
```
which needs `retrieve_window` to accept `Colon` (a simple change), the call graph for datawrangling then becomes:
```julia
  Field(metadatum, grid; cache, tile_bytes, kw…)      set!(field, metadatum; kw…)
          metadata_field.jl:372                            metadata_field.jl:398
                      ↘                                      ↙
                regrid_from_metadata!(target, metadata; …)      tiled_regridding.jl:216
                  │
                  ├─ Downloads.download(metadata)
                  │
                  ├─ tiled_native_grid(target, metadata, inpainting, halo)          :186
                  │     ├─ windowed_retrieval(metadata.dataset)     — is it big enough?    :18
                  │     ├─ default_regridding(metadata)             — bilinear hook?  ★NEW :180
                  │     │     └─ which(interpolate_physical!, …)
                  │     ├─ isnothing(inpainting) · region isa BoundingBox
                  │     ├─ horizontal_centers(target.grid)                               :107
                  │     └─ native_grid(…) → topology(native, 1) === Bounded
                  │
                  ├─ validate_vertical_extent(…)                                         :202
                  │
                  ├── nothing ⇒ WHOLE-WINDOW PATH
                  │     ├─ Field(metadata, arch; inpainting, halo, kw…)   metadata_field.jl:215
                  │     │     ├─ native_grid(metadata, arch; halo)
                  │     │     ├─ retrieve_data(metadata)                              :196
                  │     │     │     └─ retrieve_window(metadata, :, :)   ★NEW EDGE
                  │     │     ├─ set_metadata_field!(field, data, metadata)           :403
                  │     │     │     ├─ read_file_coords(metadata)                     :412
                  │     │     │     └─ set_region_data!(field, data, λc, φc, metadata)
                  │     │     │           ├─ region_info(metadata.region, …)   ← derived
                  │     │     │           └─ launch!(…, :xyz, _set_region_kernel!)
                  │     │     └─ inpaint_mask! (optional) · propagate_vertically! (Column)
                  │     └─ interpolate_physical!(target, source, metadata)
                  │
                  └── native  ⇒ regrid_in_tiles!(target, metadata, native, tile_bytes)   :120
                        ├─ read_file_coords · mangling_for
                        ├─ native_file_offset(λ, native_longitude) ×2                    :76
                        │     └─ compute_bounding_indices          (set_region_data.jl:21)
                        ├─ tile_count(native, FT, tile_bytes)                            :100
                        │
                        └─ for each tile:
                             ├─ tile_indices(Nx, ni, tile_i) ×2                          :93
                             ├─ bracketing_indices(native_λ, west, east; margin=1) ×2     :83
                             ├─ retrieve_window(metadata, window_i, window_j)             :29
                             ├─ source = Field(native; indices = (source_i, source_j, :))
                             ├─ set_region_data!(source, data, λ, φ, metadata;
                             │                   region = BoundingBoxOffset(…),  ← injected
                             │                   parameters = KernelParameters(…))
                             └─ interpolate_physical!(view(target, target_i, target_j, :),
                                                      source, metadata)

  Shared leaf, reached from both paths:

  retrieve_window(metadata, longitude_indices, latitude_indices)        tiled_regridding.jl:29
        └─ window_indices(indices, coordinates)   — `:` → eachindex  (To implement)

  interpolate_physical!(to, from, metadata)          metadata_field.jl:353   (dataset hook)
        └─ interpolate_physical!(to, from)                             :332
              ├─ interpolate!(to, from)                    — ordinary grid
              └─ launch!(…, _interpolate_physical!)        — mutable/terrain-following grid  :306
```
(maybe we can write it up somewhere and then simplify it, but it's for another PR)

src/DataWrangling/CopernicusDEM/CopernicusDEM.jl
```diff
@@ -82,6 +82,10 @@ zarr_host_path(::GLO90) = "api.earthdatahub.destine.eu/copernicus-dem/GLO-90-v0.
 
 const CopernicusDEMMetadatum = Metadatum{<:CopernicusDEMDataset}
 
+# GLO30 is 1296000 x 648000 cells globally, so a regional window is read and regridded by window
+# rather than whole.
```
simone-silvestri:
```suggestion
# GLO30 is regridded by window rather than whole.
```

src/DataWrangling/GHSL/GHSL.jl
```diff
@@ -133,6 +133,10 @@ const GHSBuiltHMetadatum = Metadatum{<:GHSBuiltH}
 const GHSBuiltSMetadatum = Metadatum{<:GHSBuiltS}
 const GHSLMetadatum      = Metadatum{<:AbstractGHSLDataset}
 
+# The 10 m built-up rasters are large enough that a regional window is read and regridded by
+# window rather than whole.
```
simone-silvestri:
```suggestion
# The 10 m built-up rasters are large and regridded by window rather than whole.
```

src/DataWrangling/OpenLandMap/OpenLandMap.jl
```diff
@@ -189,6 +189,10 @@ function DataWrangling.retrieve_data(metadata::OpenLandMapSoilDBMetadatum)
     return data
 end
 
+# The 30 m regional window is large enough that it is read and regridded by window rather than
+# whole.
```
simone-silvestri:
```suggestion
# The 30 m regional window is large and regridded by window.
```

simone-silvestri (approved):

src/DataWrangling/tiled_regridding.jl
xkykai: right now it's not. I briefly considered that idea but thought that might be too intrusive, should probably let the user decide for themselves how much to use and for the default it's just 6 MB

PR #587, opened 2026-08-25
Surface `GLO30`/`GLO90` and correct CopernicusDEM's file traits

xkykai:
`GLO30` and `GLO90` were never brought into scope at the top level, so `using NumericalEarth`
could not see them, and three of CopernicusDEM's traits did not match the file its downloader
writes.

- `using .DataWrangling.CopernicusDEM`, plus `GLO30, GLO90` in the export list. Every other
  dataset module is already surfaced this way.
- `longitude_name` and `latitude_name` return `"lon"` and `"lat"`, the names the Zarr downloader
  writes. The defaults, `"longitude"` and `"latitude"`, are not variables in the file.
- `default_inpainting` returns `nothing`. The generic default is `NearestNeighborInpainting(5)`,
  meant for ocean datasets whose land cells are missing; the DEM covers every cell of its window,
  with ocean set to 0.

```julia
using NumericalEarth

GLO30()
```

`Field(metadatum, grid)` on a `GLO30` metadatum still fails further in, on a disk-read indexing
error for the DEM's two-dimensional `z` variable. This makes the traits correct; it does not
finish `Field` support.

### DIFF
--- a/src/DataWrangling/CopernicusDEM/CopernicusDEM.jl
+++ b/src/DataWrangling/CopernicusDEM/CopernicusDEM.jl
@@ -85,6 +85,13 @@ const CopernicusDEMMetadatum = Metadatum{<:CopernicusDEMDataset}
 DataWrangling.dataset_variable_name(data::CopernicusDEMMetadatum) =
     CopernicusDEM_bathymetry_variable_names[data.name]
 
+# The downloaded window names its axes "lon"/"lat".
+DataWrangling.longitude_name(::CopernicusDEMMetadatum) = "lon"
+DataWrangling.latitude_name(::CopernicusDEMMetadatum)  = "lat"
+
+# The DEM covers every cell of its window, with ocean set to 0, so there is nothing to fill.
+DataWrangling.default_inpainting(::CopernicusDEMMetadatum) = nothing
+
 DataWrangling.metadata_filename(dataset::CopernicusDEMDataset, name, date, region) =
     string(dataset_prefix(dataset), "_", bounding_box_suffix(region), ".nc")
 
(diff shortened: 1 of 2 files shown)


src/DataWrangling/CopernicusDEM/CopernicusDEM.jl
```diff
@@ -85,6 +85,13 @@ const CopernicusDEMMetadatum = Metadatum{<:CopernicusDEMDataset}
 DataWrangling.dataset_variable_name(data::CopernicusDEMMetadatum) =
     CopernicusDEM_bathymetry_variable_names[data.name]
 
+# The downloaded window names its axes "lon"/"lat".
+DataWrangling.longitude_name(::CopernicusDEMMetadatum) = "lon"
+DataWrangling.latitude_name(::CopernicusDEMMetadatum)  = "lat"
+
+# The DEM covers every cell of its window, with ocean set to 0, so there is nothing to fill.
```
simone-silvestri:
```suggestion
```

src/DataWrangling/CopernicusDEM/CopernicusDEM.jl
```diff
@@ -85,6 +85,13 @@ const CopernicusDEMMetadatum = Metadatum{<:CopernicusDEMDataset}
 DataWrangling.dataset_variable_name(data::CopernicusDEMMetadatum) =
     CopernicusDEM_bathymetry_variable_names[data.name]
 
+# The downloaded window names its axes "lon"/"lat".
```
simone-silvestri:
```suggestion
```

simone-silvestri (approved):

PR #590, opened 2026-08-25
Reject a zero-plane displacement at or above the surface layer height

### LINKED ISSUE #582: `displaced_profile_height` silently caps a too-tall zero-plane displacement, and floors only on momentum roughness
Split out of #523, which ran into this while giving the roughness slots per-cell values.

`displaced_profile_height` is the only thing standing between a tall zero-plane displacement and a divide-by-zero in the log profile:

https://github.com/NumericalEarth/NumericalEarth.jl/blob/3c8a8f53316794c4535acb6a1d31e16938c48a48/src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl#L313

```julia
@inline displaced_profile_height(Δh, d, ℓ) = max(Δh - d, 2ℓ)
```

Two problems.

**1. When the floor binds, the answer is silently wrong rather than an error.**

Once `d ≥ Δh - 2ℓu`, the profile is evaluated at `2ℓu` regardless of the actual geometry, so the neutral transfer coefficient saturates at `ϰ² / log(2)² ≈ 0.33` — about two orders of magnitude above any real surface. Neutral, `U = 5` m/s: `u★ = ϰ U / log 2 ≈ 2.89` m/s. Nothing throws, nothing warns; the coupled model just runs with a surface that transfers momentum ~100× too efficiently.

This is reachable in the urban workflow we advertise, not just in a pathological configuration. `PrescribedAtmosphere` defaults to `surface_layer_height = 10` m, while `MorphometricRoughness` under `VariableHeight` returns a displacement that (per its own docstring) "is bounded by the tallest element rather than by `h`, and routinely exceeds the mean building height". A city block of 15 m mean height plausibly lands `d > 10` m.

**2. The floor only knows about the momentum roughness length.**

The call site passes `ℓu₀`:

https://github.com/NumericalEarth/NumericalEarth.jl/blob/3c8a8f53316794c4535acb6a1d31e16938c48a48/src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl#L368

but the temperature and water-vapor roughness lengths are independent user inputs (both slots accept a plain `Number`). With `ℓθ > Δh - d` the temperature profile is evaluated *below* its own roughness sublayer: `log(h / ℓθ) < 0`, and the heat transfer coefficient changes sign. E.g. `Δh = 10`, `d = 9.95`, `ℓu = 0.01`, `ℓθ = 0.1`: the momentum floor `2ℓu = 0.02` never binds, `h = 0.05 < ℓθ`, `χθ = ϰ / log(0.5) < 0`.

**Proposed fix**

- Floor on `ℓᵐᵃˣ = max(ℓu₀, ℓθ₀, ℓq₀)` so every profile stays above its own sublayer.
- Reject the misconfiguration at setup instead of silently capping it: when the interface is built, check `zᵃᵗ - d ≥ 2ℓᵐᵃˣ` for the roughness and displacement values that are known then (numbers and per-cell fields; per cell, not extreme-against-extreme), and keep the runtime floor purely as a backstop for state-dependent formulations whose values only exist inside the solve.

An alternative raised in review on #523 is to drop the height floor and instead floor the profile itself, `Φ = max(Φn - ψh + ψℓ, Φn/1000)`, which also keeps the coefficients finite but does not address a displacement that puts `Δh - d` at or below zero.

xkykai:
A zero-plane displacement at or above the atmosphere surface layer height leaves no room for the similarity profiles; the `max(Δh - d, 2ℓu)` floor silently capped it instead, saturating the neutral transfer coefficient at `ϰ²/log²2 ≈ 0.33`.

- `ComponentInterfaces` now throws when a `Number` `zero_plane_displacement` on a `SimilarityTheoryFluxes` reaches `minimum(surface_layer_height)`.
- The runtime floor is removed: the profiles are evaluated at `Δh - d`. A per-cell displacement that violates the bound at runtime fails loudly (`DomainError` on CPU, `NaN` on GPU) instead of producing plausible capped fluxes.

Closes #582.

### DIFF
--- a/src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
+++ b/src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
@@ -1,4 +1,3 @@
-using DocStringExtensions: TYPEDSIGNATURES
 using Oceananigans.Utils: prettysummary
 using Thermodynamics: Thermodynamics as AtmosphericThermodynamics
 
--- a/src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
+++ b/src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
@@ -302,15 +302,17 @@ end
     end
 end
 
-"""
-$(TYPEDSIGNATURES)
+# A zero-plane displacement at or above the surface layer height leaves no room
+# for the similarity profiles.
+validate_zero_plane_displacement(flux_formulation, zᵃᵗ) = nothing
 
-Height `Δh - d` at which the similarity profiles of a surface with zero-plane
-displacement `d` are evaluated, floored at twice the momentum roughness length `ℓ`
-so the profile stays above the roughness sublayer (and the transfer coefficients
-stay finite) when the displacement approaches the atmosphere surface layer height.
-"""
-@inline displaced_profile_height(Δh, d, ℓ) = max(Δh - d, 2ℓ)
+function validate_zero_plane_displacement(fluxes::SimilarityTheoryFluxes, zᵃᵗ)
+    d = fluxes.zero_plane_displacement
+    d isa Number || return nothing
+    zᵐⁱⁿ = minimum(zᵃᵗ)
+    d < zᵐⁱⁿ || throw(ArgumentError("zero_plane_displacement ($d m) must be below the surface layer height ($zᵐⁱⁿ m)"))
+    return nothing
+end
 
 function iterate_interface_fluxes(flux_formulation::SimilarityTheoryFluxes,
                                   Tₛ, qₛ, Δθ, Δq, Δh,
@@ -365,7 +367,7 @@ function iterate_interface_fluxes(flux_formulation::SimilarityTheoryFluxes,
     # Tall roughness elements displace the similarity profiles upward by `d`.
     d = local_zero_plane_displacement(flux_formulation.zero_plane_displacement,
                                       interior_properties)
-    Δh = displaced_profile_height(Δh, d, ℓu₀)
+    Δh = Δh - d
 
     # Transfer coefficients at height `h`
     ϰ = flux_formulation.von_karman_constant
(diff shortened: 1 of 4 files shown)


src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
```diff
@@ -306,11 +306,10 @@ end
...
 
 Height `Δh - d` at which the similarity profiles of a surface with zero-plane
-displacement `d` are evaluated, floored at twice the momentum roughness length `ℓ`
-so the profile stays above the roughness sublayer (and the transfer coefficients
-stay finite) when the displacement approaches the atmosphere surface layer height.
+displacement `d` are evaluated, floored at twice the largest of the three roughness
+lengths `ℓᵐᵃˣ` so every profile stays above its own roughness sublayer.
 """
-@inline displaced_profile_height(Δh, d, ℓ) = max(Δh - d, 2ℓ)
+@inline displaced_profile_height(Δh, d, ℓᵐᵃˣ) = max(Δh - d, 2ℓᵐᵃˣ)
```
simone-silvestri: why is it 2 and not ℓᵐᵃˣ + \epsilon with \epsilon equal to a very small number?

simone-silvestri: Personally, I preferred to have something like this as a guard
```julia
@inline function similarity_profile(::LogarithmicSimilarityProfile, stability_function, h, ℓ, L, ϵ)
    Φn = log(h / ℓ)
    ζ  = h / L
    ψh = stability_profile(stability_function, ζ)
    ψℓ = stability_profile(stability_function, ℓ / L)
    return max(Φn - ψh + ψℓ, ϵ)
end
```
because it is clears that the stability profile cannot change the sign of the log layer. Is this implementation the same guard in spirit? I think it does not guard against a negative ψℓ swamping Φn?

simone-silvestri: However, I guess also this is a valid guard, just we probably need to put that `2` somewhere as a parameter

xkykai:
I'm beginning to think this edge case should not be hit in a correctly set up simulations at all, so we should choose a sensible guard to prevent blowup or let it fail loudly.

When we initialize the model with *terrain* which doesn't take into account surface canopy/buildings, Δh = 10m and if we have displacement height d > 10m the problem is not well defined anyway.

However on the other hand if we use a *surface* model which includes (even if partially) the canopy and building heights when we should just set d = 0 and encode roughness just in the roughness lengths.

simone-silvestri (approved):

PR #596, opened 2026-08-27
Implicit immersed drag (with additional `uᵦ`)

simone-silvestri:
This PR applies the quadratic bottom drag as a semi-implicit flux $J = λ u^{n+1}$, with $λ = -μ |u|^n$ solved in the vertical solver's diagonal, rather than explicitly. On a thin bottom cell, which typically occurs in shallow regions (the worst regions in terms of stability!) the explicit stability number $λ Δt / Δz$ reaches order 10, which the implicit treatment removes completely. Both the bottom and the immersed boundary conditions are affected. This PR is a requirement to run a one degree ocean beyond a 30 minute time step (depending on the vertical grid of course).

Two additional things:

1. The default has changed: every ocean simulation now gets semi-implicit drag. To switch it off: `implicit_bottom_drag = false` goes back to the explicit form.
2. New keyword: `bottom_drag_background_velocity`: adds a parameterization for the "unresolved bottom velocity" which is added to the resolved speed in the drag: $τ = μ u √(uᵦ² + |u|²)$, representing tides the grid does not carry. Defaults to `0`,  so it is inert unless set.

Needs the `IMEXFluxBoundaryCondition` from Oceananigans (0.110.20)


xkykai (approved):
are there existing tidal forcing parameterizations which can be implemented here?

src/Oceans/ocean_simulation.jl
```diff
@@ -317,6 +328,10 @@ defaults on a per-field basis.
 - `rotation_rate`: Planetary rotation rate used for Coriolis forcing.
 - `gravitational_acceleration`: Gravitational acceleration, passed to buoyancy.
 - `bottom_drag_coefficient`: Bottom drag coefficient. May be a `Default` wrapper.
+- `bottom_drag_background_velocity`: unresolved velocity added in quadrature to the resolved speed in the
+  quadratic drag, `τ = -μ u √(uᵦ² + |u|²)`, representing tides and other motions the grid does not carry.
```
navidcy:
```suggestion
- `bottom_drag_background_velocity`: unresolved velocity `uᵦ` added in quadrature to the resolved speed in the
  quadratic drag, `𝛕 = - μ 𝐮 √(uᵦ² + |𝐮|²)`, representing tides and other motions the grid does not carry.
```

docs/src/appendix/notation.md
```diff
@@ -191,6 +192,7 @@ Symbols introduced by [`VariablySaturatedHydrology`](@ref),
 | ``S`` | `S` | salinity | Practical salinity (g kg⁻¹) |
 | ``u`` | `u` | zonal velocity | Eastward ocean velocity (m s⁻¹) |
 | ``v`` | `v` | meridional velocity | Northward ocean velocity (m s⁻¹) |
+| ``u_b`` | `uᵦ` | background bottom velocity | Unresolved velocity, mostly tidal, added in quadrature to the resolved speed in the bottom drag (m s⁻¹) |
```
glwagner: superscript `b`?

glwagner:
A more general design would use the Patankar trick so that the form of the bottom drag were not limited. I think the limitation imposed by the required time discretization could eventually prove to be a fatal limtation

simone-silvestri:
I think the $u_\beta$ does not limit the drag because it modulates the coefficient rather than applying an explicit drag. So this basically translates only in a larger drag coefficient which should be handled easily by the implicit solver.

docs/src/appendix/notation.md
```diff
@@ -192,7 +192,7 @@ Symbols introduced by [`VariablySaturatedHydrology`](@ref),
 | ``S`` | `S` | salinity | Practical salinity (g kg⁻¹) |
 | ``u`` | `u` | zonal velocity | Eastward ocean velocity (m s⁻¹) |
 | ``v`` | `v` | meridional velocity | Northward ocean velocity (m s⁻¹) |
-| ``u_b`` | `uᵦ` | background bottom velocity | Unresolved velocity, mostly tidal, added in quadrature to the resolved speed in the bottom drag (m s⁻¹) |
+| ``u_\beta`` | `uᵦ` | background bottom velocity | Unresolved velocity, mostly tidal, added in quadrature to the resolved speed in the bottom drag (m s⁻¹) |
```
glwagner: how about u^b for this? \beta is weird for "background". or ^{bg}

simone-silvestri: yeah, much better

Issue #595, opened 2026-08-27
Roadmap: calendar time in NumericalEarth simulations


simone-silvestri:
## Goal

Carry model time and dataset time as **dates**, not as floating-point offsets from implicit origins. `Δt` stays `Float64` seconds — that is the combination Oceananigans is already built for, with `time` and `last_Δt` as separate type parameters on `Clock` and `Utils/times_and_datetimes.jl` bridging them.

## Why

Three things the current representation cannot express, each of which has already cost us:

**1. (The reason that motivated this issue): It loses resolution as the run gets long.** In an eORCA1 OMIP run at `Δt = 5400 s`, the `Float32` time axis of the forcing data rounds its nodes by 32 s once model time passes `2^28 s`, so the bracketing weight reaches 1.001484 at iteration 49718. `interpolating_time_indices(::Cyclical, …)` reads "past the end of the record" and wraps the window to record `Nt`. All nine 3-hourly JRA55 series then alternated between the 2017 end of the archive and the correct index, every 12 model hours, for the rest of the run
***TLDR***: Float32 is a not accurate representation of the clock for long-running simulations and the model clock desynchronizes with the forcing times. This is easily fixable with Float64 clocks, but, since we already have to solve this issue, why not directly using dates that is correct always? :)

**2. It has no shared origin.** `native_times(metadata; start_time = first(metadata).dates)` anchors each dataset at *its own* first record. The 3-hourly mean JRA55 fields are first stamped at 01:30 and the instantaneous ones at 00:00, so radiation and precipitation are applied 90 minutes early relative to wind and temperature. Both axes honestly begin at zero, which is why it is invisible.

**3. It cannot express a climatology.** `RepeatYearJRA55` is stamped 1990 and `WOAMonthly` 2018. Their years are metadata, not physics, and there is currently no way to say so — the model cannot be in both.

Widening the axis to `Float64` removes (1) (verified: no wrap anywhere in 1958–2018, against a first wrap at 49718 with `Float32`). It buys thirty bits of headroom, not a guarantee, and does nothing for (2) or (3).
***TLDR***: let's use Dates to solve all our problems!

## Status

Both stages are merged. Still untested against real downloads, a coupled `EarthSystemModel`, the checkpointer and GPU.

### 1 — the axis follows the clock (`ss/axis-clock-agreement`)

Make the axis match the clock, then make disagreement unrepresentable.

[...]

PR #598, opened 2026-08-27
Validate dataset times axis

simone-silvestri:
This PR is the first of three following the roadmap in #595. The first step is to make sure the clock of the model agrees with the time axis of the datasets, otherwise the datsets can drift away from the model in long standing simulations. 
This is a very simple building block that will be build upon to finally switch to a complete `Dates` clock for both model and prescribed components


xkykai (approved):
lgtm

PR #599, opened 2026-08-27
(0.8.0) Implement a time anchor for different datsets

simone-silvestri:
Closes stage 2 of #595, on top of #598.

The `EarthSystemModel`, ocean and sea-ice clocks and the `FieldTimeSeries` axes become `DateTime`, and each dataset says how it wants a model date interpreted. `Δt` stays `Float64` seconds.

### What changed

1. We introduced "Anchoring" as a `time_indexing`. The new options are
- `CalendarDate`
- `CalendarPhase`
- `SimulationStart` 
They plug into `interpolating_time_indices`, which is the single dispatch point behind both `update_field_time_series!` and in-kernel `interpolate`, so one method covers the host and the GPU path, and it arrives through the `time_indexing` keyword that already exists on every `FieldTimeSeries` constructor.

Basically these are the calendar-aware siblings of the `Cyclical` time indexing (for all of these behaviors the time is cycled through the period, see later on the behavior).

2. `native_times` returns dates, and collapses doing it:

```julia
native_times(metadata) = [window_center(datum) for datum in metadata]
```

3. Every component clock derives from its own time axis. Previously `PrescribedAtmosphere` and `PrescribedRadiation` hardcoded `Clock{Float64}`, and `PrescribedLand` used `Clock{eltype(first_flux)}`. All three now go through one `default_component_clock(times)`, and `default_earth_system_clock` inherits the atmosphere's time type and starting time.

4. Added a `start_time` field to the `EarthSystemModel` such that we are always able to `reset!` the simulation clock.

### Behaviour

The time axis carries dates, not floats

```julia
julia> md = Metadata(:temperature; dataset=RepeatYearJRA55(), start_date=DateTime(1990,1,1), end_date=DateTime(1990,1,2));

julia> native_times(md)[1:3]
3-element Vector{DateTime}:
 1990-01-01T00:00:00
 1990-01-01T03:00:00
 1990-01-01T06:00:00
```

The clock follows the axis, and the coupled clock follows the atmosphere (by default, we can also change it):

```julia
julia> atmos = PrescribedAtmosphere(grid, collect(DateTime(1958,1,1) : Hour(3) : DateTime(1958,1,10)));

julia> atmos.clock.time
1958-01-01T00:00:00

julia> default_earth_system_clock(atmos).time
1958-01-01T00:00:00
```

Each dataset says how it wants the model date interpreted (also this is a default that can be changed):

```julia
julia> default_time_indexing.((MultiYearJRA55(), RepeatYearJRA55(), WOAMonthly()))
(CalendarDate(), CalendarPhase(), CalendarPhase())
```

A reanalysis reads the model's own date, and folds by whole calendar years once the run passes the end of the record while a second OMIP cycle repeats the same calendar dates, with the model clock left monotonic:

```julia
julia> lookup_date(CalendarDate(), DateTime(1967,2,12), jra)   # in range
1967-02-12T00:00:00

julia> lookup_date(CalendarDate(), DateTime(2018,6,1), jra)    # past the record
1958-06-01T00:00:00

julia> lookup_date(CalendarDate(), DateTime(2027,2,12), jra)   # second cycle
1967-02-12T00:00:00
```

A climatology reads the model's month and day, resolved in whatever year it happens to be stamped in — so `RepeatYearJRA55` (1990) and `WOAMonthly` (2018) can be used together without the model having to be in either year:

```julia
julia> lookup_date(CalendarPhase(), DateTime(1966,3,15,9), ryf)   # repeat year, stamped 1990
1990-03-15T09:00:00

julia> lookup_date(CalendarPhase(), DateTime(1966,3,15,9), woa)   # climatology, stamped 2018
2018-03-15T09:00:00

julia> lookup_date(CalendarPhase(), DateTime(2043,3,15,9), ryf)   # 77 years later, same phase
1990-03-15T09:00:00
```

### Caveat: LEAP YEARS

The difficult part, which requires a decision, is handling leap years. Some "year-agnostic" datasets (like `RepeatYearJRA55`) do not have the 29th of February. This is the decision we came up with Claude and I:
29 February reads 28 February at the same time of day, since a non-leap nominal year has no counterpart for it.
The opposite case (dataset has a leap year, model does not), the simple solution is to skip the 29th of Feb.

Leap model year onto a non-leap dataset — 29 February replays the 28th:

```julia
2044-02-28T00:00:00   -> 1990-02-28T00:00:00
2044-02-28T12:00:00   -> 1990-02-28T12:00:00
2044-02-29T00:00:00   -> 1990-02-28T00:00:00     <-- replayed, midnight to midnight
2044-02-29T12:00:00   -> 1990-02-28T12:00:00     <-- noon to noon
2044-03-01T00:00:00   -> 1990-03-01T00:00:00
2044-03-01T12:00:00   -> 1990-03-01T12:00:00
```

Non-leap model year onto a leap dataset — the dataset's 29 February is never read:

```julia
2043-02-28T00:00:00   -> 1960-02-28T00:00:00
2043-02-28T12:00:00   -> 1960-02-28T12:00:00
2043-03-01T00:00:00   -> 1960-03-01T00:00:00     <-- 1960-02-29 skipped
2043-03-01T12:00:00   -> 1960-03-01T12:00:00
```

This means that in leap years we effectively repeat the 28th of Feb for datasets like RYF, which is probably better than just interpolating between 28th of Feb and 1st of March since at least it preserves the daily cycle and avoids freezing a certain hour.

### Known sharp edge 

(Claude calls it a sharp edge, I call it a desired feature of the implementation)
`CalendarDate`'s fold period comes from the loaded window, not from the dataset. Setting `start_date = 1967-02-12` on JRA55 trims the metadata, so the cycle becomes 51 years instead of 60:

```
model 2027-02-12  ->  full record 1958..2018: 1967-02-12
                      trimmed     1967..2018: 1976-02-12
```

So, exact leap-year preservation might be off because the cycle can not be a multiple of 4, but given the fact that on non-leap years the 29th of feb maps to the 28th it is not a problem.

***Bump major version, as this is a big change to the clock type.***

### DIFF
--- a/src/DataWrangling/time_anchoring.jl
+++ b/src/DataWrangling/time_anchoring.jl
@@ -79,9 +79,9 @@ a leap year becomes 28 February.
 """
 function substitute_year(date, target_year)
     month(date) == 2 && day(date) == 29 && !isleapyear(target_year) &&
-        return DateTime(target_year, 2, 28, hour(date), minute(date), second(date))
+        return DateTime(target_year, 2, 28, hour(date), minute(date), second(date), Dates.millisecond(date))
 
-    return DateTime(target_year, month(date), day(date), hour(date), minute(date), second(date))
+    return DateTime(target_year, month(date), day(date), hour(date), minute(date), second(date), Dates.millisecond(date))
 end
 
 """
--- a/src/DataWrangling/time_anchoring.jl
+++ b/src/DataWrangling/time_anchoring.jl
@@ -1,11 +1,14 @@
 using Dates: Dates, DateTime, Millisecond, year, month, day, hour, minute, second, isleapyear
+using Oceananigans.OutputReaders: find_time_index
 
 """
     AbstractTimeAnchor
 
-Supertype for the rules mapping a model date onto a dataset's own time stamps.
+Supertype for the `time_indexing` rules that map a model date onto a dataset's own time stamps.
 """
 abstract type AbstractTimeAnchor end
+
+Oceananigans.OutputReaders.interpolating_time_indices(anchor::AbstractTimeAnchor, times, t) = find_time_index(times, lookup_date(anchor, t, times))
 
 """
     CalendarDate()
@@ -49,10 +52,10 @@ function lookup_date(::CalendarDate, model_date, dates)
 
     folded = model_date
     while folded > last_stamp
-        folded = shift_years(folded, -cycle_years)
+        folded = substitute_year(folded, year(folded) - cycle_years)
     end
     while folded < first_stamp
-        folded = shift_years(folded, cycle_years)
+        folded = substitute_year(folded, year(folded) + cycle_years)
     end
 
     return folded
@@ -82,19 +85,12 @@ function substitute_year(date, target_year)
 end
 
 """
-    shift_years(date, n)
+    default_time_indexing(dataset)
 
-`date` shifted by `n` whole calendar years.
+The `time_indexing` `dataset` uses unless the caller overrides it. Climatologies, whose nominal year carries no
+meaning, extend this with `CalendarPhase()`.
 """
-shift_years(date, n) = substitute_year(date, year(date) + n)
-
-"""
-    default_time_anchor(dataset)
-
-The anchor `dataset` uses unless the caller overrides it. Climatologies, whose nominal year carries no meaning,
-extend this with `CalendarPhase()`.
-"""
-default_time_anchor(dataset) = CalendarDate()
+default_time_indexing(dataset) = CalendarDate()
 
 """
     validate_time_anchors(anchored_series)
--- a/src/DataWrangling/time_anchoring.jl
+++ b/src/DataWrangling/time_anchoring.jl
@@ -1,0 +1,111 @@
+using Dates: Dates, DateTime, Millisecond, year, month, day, hour, minute, second, isleapyear
+
+"""
+    AbstractTimeAnchor
+
+Supertype for the rules mapping a model date onto a dataset's own time stamps.
+"""
+abstract type AbstractTimeAnchor end
+
+"""
+    CalendarDate()
+
+Read a dataset at the model's own date, folding back by whole calendar years once the model passes the end of
+the record.
+"""
+struct CalendarDate <: AbstractTimeAnchor end
+
+"""
+    CalendarPhase()
+
+Read a dataset at the model's month, day and time of day resolved in the dataset's own year. A model 29 February
+reads 28 February.
+"""
+struct CalendarPhase <: AbstractTimeAnchor end
+
+"""
+    SimulationStart(start_date)
+
+Read a dataset from its first record at `start_date`, advancing by elapsed time and wrapping by the span of the
+record.
+"""
+struct SimulationStart{D} <: AbstractTimeAnchor
+    start_date :: D
+end
+
+"""
+    lookup_date(anchor, model_date, dates)
+
+The date at which a dataset stamped at `dates` is read for a model at `model_date`.
+"""
+function lookup_date end
+
+function lookup_date(::CalendarDate, model_date, dates)
+    first_stamp, last_stamp = first(dates), last(dates)
+    first_stamp <= model_date <= last_stamp && return model_date
+
+    cycle_years = year(last_stamp) - year(first_stamp)
+    cycle_years < 1 && return model_date
+
+    folded = model_date
+    while folded > last_stamp
+        folded = shift_years(folded, -cycle_years)
+    end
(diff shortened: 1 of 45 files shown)


simone-silvestri:
This PR requires adding a `start_time` to the `EarthSystemModel` so that we can reset the clock of the simulation to the start time (it is part of the testing suite).
I am ok with keeping it here, but it might be more plausible to add it as a field of the `Clock` in Oceananigans.

xkykai:
The leap years are really hard. I think it's reasonable that we skip/repeat 29/28 Feb as proposed. What are the alternatives? What do others usually do? 

As an aside I'm tempted to say that if we demand that we strictly follow the date system of the (atmospheric) forcing we will never run into issues like that, so we might have time from simulation start but we will not have misaligned calendars between ocean and `RepeatYearJRA55` if we follow whatever calendar system the (dataset-derived) dataset has, no matter how odd the calendar might be. But I haven't thought deeply about what implications this may have.

src/DataWrangling/time_anchoring.jl
```diff
@@ -0,0 +1,116 @@
...
+"""
+struct CalendarDate <: AbstractTimeAnchor end
+
+"""
+    CalendarPhase()
+
+Read a dataset at the model's month, day and time of day resolved in the dataset's own year. A model 29 February
+reads 28 February.
+"""
+struct CalendarPhase <: AbstractTimeAnchor end
```
xkykai, line 34: I wonder if we should implement other methods that allow the user to define what kind of calendar it is (eg. leap year or not)

src/EarthSystemModels/earth_system_model.jl
```diff
@@ -4,9 +4,10 @@ using Oceananigans.TimeSteppers: Clock
 using Oceananigans.OutputReaders: extract_field_time_series
 using KernelAbstractions: @kernel, @index
 
-mutable struct EarthSystemModel{R, A, L, I, O, F, C, Arch} <: AbstractModel{Nothing, Arch}
+mutable struct EarthSystemModel{R, A, L, I, O, F, C, T, Arch} <: AbstractModel{Nothing, Arch}
     architecture :: Arch
     clock :: C
+    start_time :: T
```
glwagner, line 10: I don't understand the need for this. What is special about `EarthSystemModel` vs other types of models? There should be nothing special.

glwagner: I think if you want to add the concept of an `epoch` to `Clock`, maybe that is what you are looking for

simone-silvestri: yeah I have mentioned it here:
> This PR requires adding a start_time to the EarthSystemModel so that we can reset the clock of the simulation to the start time (it is part of the testing suite).
I am ok with keeping it here, but it might be more plausible to add it as a field of the Clock in Oceananigans.

I also don't like too much having it here, it should be in the clock

src/DataWrangling/time_anchoring.jl
simone-silvestri: I am not sure about this, because each dataset is anchored to its dates, these anchors just tell the users how the datasets are supposed to interpret the model clock, which can be switched by the users through switching the `time_indexing`.

simone-silvestri:
> The leap years are really hard. I think it's reasonable that we skip/repeat 29/28 Feb as proposed. What are the alternatives? What do others usually do?

I think it is a matter of us choosing a convention and sticking to it. I wouldn't introduce multiple conventions.

> As an aside I'm tempted to say that if we demand that we strictly follow the date system of the (atmospheric) forcing we will never run into issues like that, so we might have time from simulation start but we will not have misaligned calendars between ocean and RepeatYearJRA55 if we follow whatever calendar system the (dataset-derived) dataset has, no matter how odd the calendar might be. But I haven't thought deeply about what implications this may have.

This would be the `SimulationStart` time indexing which is basically equivalent to the float time but with dates instead of numbers, i.e. the dataset and the model only see the seconds elapsed from the starting time, not the actual date

glwagner:
> This would be the `SimulationStart` time indexing which is basically equivalent to the float time but with dates instead of numbers, i.e. the dataset and the model only see the seconds elapsed from the starting time, not the actual date

what does "simulation start" refer to?

glwagner:
> > The leap years are really hard. I think it's reasonable that we skip/repeat 29/28 Feb as proposed. What are the alternatives? What do others usually do?
> 
> I think it is a matter of us choosing a convention and sticking to it. I wouldn't introduce multiple conventions.

I thought that `DateTime` already includes different date types to handle these. There are date types for no leap years, and also for 360 day years. Right?

PR #601, opened 2026-08-27
Implement chlorophyll-based radiation absorption

simone-silvestri:
This PR implements an anchor to make the blue-green absorption coefficient chlorophyll dependent. There will be a follow-up which introduces a chlorophyll climatology which can be used for this scope in DataWrangling.

This PR is motivated by the fact that, coincidentally, places where freshwater is applied to the surface (like river mouths, the bane of my existance...), are not clear water regions! What happens currently is that we apply a "too large" penetrative depth which traps heat below a fresh water surface which cannot escape and these isolated spots (amazon mouth, gange mouth, congo mouth), show maximum temperatures of about 37 - 39 degrees C. 

Introducing a chlorophyll map reduces these maximum temperatures to more tractable 32 - 33 degrees (verified in the OMIP). This PR only introduces the wiring which allows to ingest a chlorophyll map.
It would be awesome, in the future, if we can use OceanBIOME to inform the radiative penetration (@jagoosw)


jagoosw:
We can definitely do this prognostically in the near future!

jagoosw (approved):
Sorry thought I approved this before!

PR #604, opened 2026-08-27
Hand the nested child the (θˡⁱ, moisture) pair its own microphysics inverts

### LINKED ISSUE #560: State exchanger writes qᵛ instead of qᵉ, and forms θˡⁱ by a definition Breeze does not invert
`_compute_child_prognostics!` (`ext/NumericalEarthBreezeExt/breeze_state_exchanger.jl:67-96`) builds the
child's energy/moisture pair on the parent grid. Neither half is built to the definition the child's own
Breeze inversion reads it with. The two mismatches are independent.

## 1. The moisture slot: `qᵛ` written into a `qᵉ` prognostic

```julia
qˡ = qᶜˡ[i, j, k] + qʳ[i, j, k]                             # :74 — cloud liquid + rain, lumped
qⁱ = qᶜⁱ[i, j, k] + qˢ[i, j, k]                             # :75 — cloud ice + snow, lumped
θ  = liquid_ice_potential_temperature(Tᵢ, qˡ, qⁱ, pᵢ, …)    # :80 — ℒqˡ/cᵖ removed for ALL of it
…
ρθ[i, j, k]  = ρᵈ[i, j, k] * θᵢ                              # :83 ┐ the pair the child is handed
ρqᵛ[i, j, k] = ρ * qᵛᵢ                                       # :86 ┘ vapor only
```

Those writes land in a `FieldTimeSeries` the exchanger calls `prognostic.ρqᵛ`, which
`breeze_nested_atmosphere.jl` hands to the child under whatever the microphysics names its moisture
prognostic — `NamedTuple{tuple(moisture_name)}(tuple(ρqᵛ))` at `:259`, relabeled but not converted. It
supplies both the lateral BC values (`:280-283`) and the Davies relaxation targets (`:314`). The two
conventions agree only for some schemes:

| name | schemes | content | exchanger correct? |
|---|---|---|---|
| `:ρqᵛ` | `InstantaneousPrecipitation`, `NonEquilibriumCloudFormation`, `DCMIP2016KM`, `nothing` | vapor | yes |
| `:ρqᵉ` | `SaturationAdjustment`, incl. as the `cloud_formation` of `BulkMicrophysics` | **equilibrium** moisture: `qᵉ = qᵛ + qᶜˡ (+ qᶜⁱ)`, i.e. `qᵗ` less precipitation (`microphysics_interface.jl:640,660` at Breeze v0.11.0) | **no — cloud condensate missing** |

The second row is what `default_nested_microphysics()` returns, so it is the default nesting path. There
the child gets `θˡⁱ` with the condensate's latent heat subtracted, paired with a moisture that never
carried the condensate; saturation adjustment re-partitions too little equilibrium moisture, condenses too
little, and recovers a temperature low by roughly `ℒqᶜ/cᵖ`.

## 2. `θˡⁱ`: a dry-κ formula against a mixture-κ inversion

```julia
# NumericalEarth (breeze_prognostic_state.jl:27) — DRY exponent, cᵖᵈ in the latent term
θˡⁱ = T * (pˢᵗ/p)^(Rᵈ/cᵖᵈ) * (1 - (ℒˡqˡ + ℒⁱqⁱ)/(cᵖᵈ T))

# Breeze `with_temperature` (dynamic_states.jl:138) — MIXTURE exponent, cᵖᵐ
θˡⁱ = (T - (ℒˡqˡ + ℒⁱqⁱ)/cᵖᵐ) / Π,   Π = (p/pˢᵗ)^(Rᵐ/cᵖᵐ)
```

`with_temperature` is the exact algebraic inverse of `temperature` — forward-then-inverse returns the
input to `0.0 K` — so the child recovers exactly the temperature it was sent only if the producer used
that definition.

## Evidence

Both are visible with no grid and no simulation, by inverting the emitted pair with the child's own
saturation adjustment. Sweeping cloud liquid at 283.15 K / 900 hPa, `Breeze.Microphysics.compute_temperature`
returns (**ΔT** = defect 1; `round-trip` = defect 2, the error remaining with the *correct* `qᵉ`):

| `qᶜˡ` g/kg | `ℒqᶜˡ/cᵖᵈ` | `T(qᵉ = qᵛ + qᶜˡ)` | `T(qᵉ = qᵛ)` | **ΔT** | round-trip |
|---|---|---|---|---|---|
| 0.10 | 0.2488 | 283.1540 | 283.0507 | **−0.1033** | +0.0040 |
| 0.25 | 0.6221 | 283.1562 | 282.8972 | **−0.2590** | +0.0062 |
| 0.50 | 1.2442 | 283.1596 | 282.6396 | **−0.5200** | +0.0096 |
| 1.00 | 2.4884 | 283.1650 | 282.1167 | **−1.0483** | +0.0150 |

`ΔT ≈ −1.05 K` per g/kg of cloud liquid — ~42 % of `ℒqᶜˡ/cᵖᵈ`, not 100 %, because withholding moisture
cools the parcel, lowers saturation, and re-condenses part of what remains. The damping
`1/(1 + (ℒ/cᵖ)·dqᵛ⁺/dT)` is 0.414 at this state against 0.415–0.421 measured. It weakens as the air cools,
and 283 K is near its mildest: `ΔT` per g/kg is −1.02 K here, −1.34 K at 275 K and **−2.17 K at 250 K**
(damping 0.874), so a cold or mixed-phase case loses close to the full undamped `ℒqᶜ/cᵖ`.

The round-trip column is defect 2, and it is not solver error: tightening the secant `abstol` from 1e-6 to
1e-10 does not move it. At a fixed partition it decomposes as `+0.034 K` from the exponent less `0.011 K`
from the heat capacity. It vanishes at `p = pˢᵗ` and grows with height:

| 1000 hPa | 900 | 850 | 700 | 500 | 300 |
|---|---|---|---|---|---|
| 0.0000 K | 0.0234 | 0.0589 | 0.0893 | **0.2481** | 0.1826 |

Negligible for a shallow LES near `pˢᵗ`, but ~0.25 K through the middle of a deep nested LAM, as a
systematic height-dependent bias on every lateral boundary value and Davies target. It is ~50× under
defect 1 near the surface and cancels from `ΔT`, which is why that column is a difference of two
inversions rather than either absolute error.

<details><summary>Reproducer</summary>

```julia
using Breeze
using Breeze.Thermodynamics: MoistureMassFractions, LiquidIcePotentialTemperatureState
using Breeze.Microphysics: compute_temperature
using Breeze: ThermodynamicConstants, SaturationAdjustment, WarmPhaseEquilibrium,
              dry_air_gas_constant, adjustment_saturation_specific_humidity

FT = Float64
constants = ThermodynamicConstants(FT)
Rᵈ, cᵖᵈ = dry_air_gas_constant(constants), constants.dry_air.heat_capacity
ℒˡ, pˢᵗ = constants.liquid.reference_latent_heat, FT(100000)
microphysics = SaturationAdjustment(FT; equilibrium = WarmPhaseEquilibrium())

T, p, qᶜˡ = FT(283.15), FT(90000), FT(0.5e-3)
qᵛ = FT(0.008)                                     # saturate with the adjustment's own qsat
for _ in 1:50
    global qᵛ = adjustment_saturation_specific_humidity(T, p, qᵛ + qᶜˡ, constants, WarmPhaseEquilibrium())
end

θˡⁱ = T * (pˢᵗ / p)^(Rᵈ / cᵖᵈ) * (1 - ℒˡ * qᶜˡ / (cᵖᵈ * T))    # as the exchanger forms it
invert(qᵉ) = compute_temperature(LiquidIcePotentialTemperatureState(θˡⁱ, MoistureMassFractions(qᵉ), pˢᵗ, p),
                                 microphysics, constants)

invert(qᵛ + qᶜˡ)   # 283.1596 K — qᵉ as Breeze defines it
invert(qᵛ)         # 282.6396 K — what :86 writes
```
</details>

## Fix

Both reduce to handing the consumer what its own inverse expects.

1. Write `ρ * (qᵛ + qᶜˡ + qᶜⁱ)` when the bound name is `:ρqᵉ`, `ρ * qᵛ` when it is `:ρqᵛ`. The species
   arrive separately and are lumped into `qˡ`/`qⁱ` at `:74-75` — correct for `θˡⁱ` and for `ρ`, but it
   discards the cloud/precipitation split this write needs, so the fix reads `qᶜˡ`/`qᶜⁱ` before that step.
2. Call Breeze's `with_temperature` / `potential_temperature_from_temperature` instead of reimplementing
   the conversion. Both of Breeze's state types use the mixture exponent, so a single dry-κ formula cannot
   be right for either.

## Context

Surfaced from a nested LES whose domain-mean θ drifted ~1.6 K in 3 h identically across runs that
differed in an interior forcing — a drift shared by a run and its own control, so not physics.

The vapor-only convention dates to #107 (Breeze v0.4 dropped the total-moisture variable); schemes carrying
an equilibrium prognostic have since returned. Sibling of #514, a different consistency defect on the same
coupling path.

NumericalEarth 0.6.x, Breeze 0.8.0, Oceananigans 0.110.14.

Addressed by #604. Line references above refreshed against `main` @ `5924f3179`; the density symbols in the
first block are the source's (`ρᵈ` dry, `ρ` total), which the original transcription rendered as `ρᵂ`/`ρᵠ`.

ewquon:
Fixes #560.

The state exchanger handed the child a `(θˡⁱ, moisture)` pair built to definitions the child's own Breeze
inversion does not read it back with. Three independent mismatches, each measured by inverting the
emitted pair with the child's own `compute_temperature`.

**θˡⁱ used a dry-κ formula.** `(pˢᵗ/p)^(Rᵈ/cᵖᵈ)` with `cᵖᵈ` in the latent term, while Breeze's
`temperature` uses the mixture `Rᵐ/cᵖᵐ`. It is now taken from `with_temperature`, that function's exact
inverse. Round-trip residual at 283.15 K with 0.5 g/kg cloud liquid, before → after: −0.0103 K → 0 at
1000 hPa, +0.208 K → 0 at 500 hPa, +0.369 K → 0 at 300 hPa. It does not vanish at the reference pressure,
as both Exner factors are 1 there but the latent term still divides by `cᵖᵈ`.

**The moisture slot always held vapor.** `moisture_prognostic_name` binds it to `:ρqᵉ` for saturation
adjustment — what `default_nested_microphysics` returns, so the default nesting path — where it means
`qᵗ − qʳ − qˢ`, vapor plus cloud condensate. The child re-partitioned too little moisture against a θˡⁱ
that had already given up the condensate's latent heat, recovering a temperature 0.520 K low for 0.5 g/kg
of cloud liquid at 283 K, and roughly twice that at 250 K. Renamed `ρqᵛ` → `ρqᵛᵉ`, matching
`breeze_atmosphere_interface.jl` and Breeze's own `qᵛᵉ`.

**θˡⁱ removed latent heat for precipitation the child never receives.** The kernel lumped all four
condensate species into one pair and gave it to both the density and θˡⁱ. The density wants that; θˡⁱ does
not. The child makes its own precipitation, at its own resolution and dynamics, so the parent's `qʳ`/`qˢ`
cross neither the moisture slot (`qᵉ = qᵗ − qʳ − qˢ`) nor θˡⁱ. Removing their latent heat handed the child
a θˡⁱ it could not repay: −0.205 K at 0.2 g/kg of rain, −1.038 K at 1.0 g/kg, worsening as the air cools. A
vapor-only child lost the entire latent term undamped. θˡⁱ now gives up latent heat for exactly the
condensate the moisture slot carries, so a raining parent and a dry one hand over the same θˡⁱ.

The handoff closes to 3.2e-5 K for `:ρqᵉ` and exactly for `:ρqᵛ`, whatever precipitation the parent
carries.

Three tests encoded the dry-κ definition and now assert that Breeze's `temperature` inverts the emitted
θˡⁱ; one of them was passing only because `≈` absorbed the ~0.02 K mixture-vs-dry gap at qᵛ = 0.005.

## Not addressed here

The child receives precipitation-free air by design, so it holds less water than the parent — 2.8% less for
an equilibrium scheme with the parent at 0.25 g/kg of rain and snow, 9.6% for a vapor-only one — and its
total density is correspondingly lower. That is the intended handoff, not a deficit: the child's air
genuinely has no rain in it, and spins its own up.

One consequence is worth stating. Where the child has developed its own precipitation inside the
relaxation zone, relaxing its θˡⁱ toward a precipitation-free target biases it warm by `ℒqʳ/cᵖ`. That is
inherent to relaxing a child that owns its precipitation toward a parent that does not share it; no choice
of θˡⁱ removes it.

## Testing

`test/test_nested_simulation.jl` passes in full, 21 testsets, with CloudMicrophysics loaded so the
precipitating default child is exercised end to end, except `AtmosphereLandModel (compressible
Breeze) as a NestedSimulation child`, which errors with `FieldError: type Nothing has no field
'condition'` in `net_fluxes`. That reproduces identically on a pristine `main`, so it is unrelated to this
branch. Verified against Breeze 0.9.0 with Oceananigans 0.110.19, and against
NumericalEarth/Breeze.jl#933 with Oceananigans 0.110.20.

## Why the initial condition is keyed by the scheme's moisture name

`initialize_nested_child!` spells the `set!` moisture keyword as
`moisture_specific_name(child.microphysics)` rather than `qᵗ`. That is load-bearing from Breeze 0.11.

`set!` writes `qᵛ`/`qᵉ` as given, but `qᵗ` sets `total_moisture_given`, which gates a pass that re-derives
the prognostic moisture from total water through `specific_prognostic_moisture_from_total`
(`set_atmosphere_model.jl:268-270`, `:328`, `:412-430` at v0.11.0). Through 0.10 all three names shared one
branch and were written verbatim.

The lateral boundary prescribes the prognostic directly and has no such pass, so keying the interior by
`qᵗ` would let the initial condition and the boundary values hold different quantities at the wall — the
standing gradient this construction exists to avoid.

It does not fire yet: the gate probes for a four-argument
`specific_prognostic_moisture_from_total(microphysics, qᵗ, μ_fields, ρ)`, and the one-moment extension
defines only the three-argument state form (`one_moment_microphysics.jl:745-754`), so `hasmethod` is false
for the scheme `default_nested_microphysics` builds. It goes live when that method is added.
NumericalEarth's compat is still `Breeze = "0.10"`, so this is recorded ahead of the bump.

Raised by the session working Breeze 0.11 compat (#689), verified against the v0.11.0 tag.


ext/NumericalEarthBreezeExt/breeze_prognostic_state.jl
```diff
@@ -20,12 +26,27 @@ using Breeze: ThermodynamicConstants, dry_air_gas_constant, vapor_gas_constant
...
-# dynamics' `standard_pressure` (Breeze's `ReferenceState` value) — never hardcoded.
-@inline potential_temperature(T, p, pˢᵗ, Rᵈ, cᵖᵈ) = T * (pˢᵗ / p)^(Rᵈ / cᵖᵈ)
+# `with_temperature` is the exact inverse of the `temperature` the child reconstructs with, so θˡⁱ is
+# taken from it rather than restated here.
+#
+# `qᵛ` has no latent-heat term of its own, but it sets the answer twice. It fixes the mixture gas constant
+# `Rᵐ = qᵈRᵈ + qᵛRᵛ` and the mixture heat capacity `cᵖᵐ = qᵈcᵖᵈ + qᵛcᵖᵛ + qˡcˡ + qⁱcⁱ` (both through
+# `qᵈ = 1 − qᵛ − qˡ − qⁱ` as well), whose ratio is the Exner exponent in `Π = (p/pˢᵗ)^(Rᵐ/cᵖᵐ)`; and `cᵖᵐ`
+# divides the latent term again in `θˡⁱ = (T − (ℒˡqˡ + ℒⁱqⁱ)/cᵖᵐ)/Π`. `pˢᵗ` is the caller's
+# `standard_pressure`, never hardcoded.
```
kaiyuan-cheng, line 36:
```suggestion
# `standard_pressure`.
```

ext/NumericalEarthBreezeExt/breeze_state_exchanger.jl
```diff
@@ -64,26 +69,42 @@ end
...
-        qⁱ  = qᶜⁱ[i, j, k] + qˢ[i, j, k]
-        pᵢ  = p[i, j, k]
-
+        Tᵢ   = T[i, j, k]
+        qᵛᵢ  = qᵛ[i, j, k]
+        qᶜˡᵢ = qᶜˡ[i, j, k]
+        qᶜⁱᵢ = qᶜⁱ[i, j, k]
+        qˡ   = qᶜˡᵢ + qʳ[i, j, k]
+        qⁱ   = qᶜⁱᵢ + qˢ[i, j, k]
+        pᵢ   = p[i, j, k]
```
kaiyuan-cheng, line 83: We may want to use a consistent notation style.

kaiyuan-cheng (approved):

ext/NumericalEarthBreezeExt/breeze_state_exchanger.jl
ewquon: Thanks for catching this, this was pretty bad. `Tᵢ`, `qᵛᵢ` and `pᵢ` had an "i" subscript while `qˡ` and `qⁱ` did not; there's also `qᶜˡᶜ` and `qᶜⁱᶜ` further down...

glwagner (approved):
good but in my opinion the title "Hand the nested child the (θˡⁱ, moisture) pair its own microphysics inverts" is not comprehensible.

ewquon:
Agreed, it's embarassingly bad. I knew there was a reason nobody wanted to touch this PR with a 10-ft pole...

PR #605, opened 2026-08-28
Apply first-cell penetrative radiation as a surface flux

simone-silvestri:
When using penetrative radiation, we apply it as a forcing rather than a surface flux.
This has a pretty large consequence for CATKE which does not see the stabilizing surface flux and believes that the column is convecting, injecting huge amounts of TKE in the surface layer.

Even with the two-color split and a surface cell 1.5 meter thick, all the red light is roughly absorbed in the first layer, so applying the surface cell forcing as a surface flux recovers ~55% of the stabilizing flux which is then threadded through CATKE to the TKE equation.


glwagner:
What is the argument for not applying this fix to CATKE ?

src/Oceans/radiative_forcing.jl
```diff
@@ -231,17 +232,39 @@ end
...
 end
 
 @inline shortwave_radiative_forcing(i, j, grid, Fᵀ, ℐₜˢʷ, ocean_properties) = ℐₜˢʷ
 
+"""
+$(TYPEDSIGNATURES)
+
+Fraction of the net shortwave absorbed within the surface cell, from Beer's law on both colors.
+Vertical closures build their surface buoyancy flux from the tracer boundary conditions, so this
+part of the shortwave has to arrive there rather than as an interior source. Beer's law is evaluated
```
glwagner, line 249: I'm not sure you should mention vertical closures in this docstring

src/Oceans/radiative_forcing.jl
```diff
@@ -231,17 +232,39 @@ end
     dJ₂dz = ∂zᶜᶜᶜ(i, j, k, grid, beers_law_radiation, J₀, κ₂)
 
     # Net radiation flux divergence
-    return ϵ₁ * dJ₁dz + (1 - ϵ₁) * dJ₂dz
+    dJdz = ϵ₁ * dJ₁dz + (1 - ϵ₁) * dJ₂dz
+
+    # The surface cell's share is delivered through the temperature boundary condition instead. 
+    surface_share = surface_absorbed_fraction(i, j, grid, R) * J₀ / Δzᶜᶜᶜ(i, j, Nz, grid)
+    return dJdz - ifelse(k == Nz, surface_share, zero(dJdz))
```
glwagner, line 239:
```suggestion
    return dJdz - (k == Nz) * surface_share
```

simone-silvestri:
I think a CATKE fix is better. We need to add a radiation handler in Oceananigans and modify the $w_\star$ convective scale to see the radiation.  Solar radiation adds to $w_\star$ which slightly increases the convective velocity, the mixing length and the surface bcs of tke as compared to if the solar radiation was all dumped at the surface.

The solution here is a bit of a hack: it only considers part of the radiation as a stabilizing effect (as opposed to main which is completely blind to radiation). The correct formulation discounts the stabilizing effects of solar radiation based on the penetration depth. 

Basically, if we assume that the flux profile is linear during convection we have that
```math
\overline{w^\prime T^\prime}(z) + J_{sw}(z) = J_T(0)\left(1 + \frac{z}{h}\right)
```
which, after integration in the mixed layer leads to (with an absorption coefficient of $\kappa$)
```math
w_\star^3 \sim e^{3/2} \sim J_b h - J_{b, sw}(0) \frac{1 - e^{- \kappa h}}{\kappa}
```
where $J_{b, sw}$ is the (negative, stabilizing) buoyancy flux associated with the impinging solar radiation at the top, which leads to a non-linear equation for $h$ (the mixing length in this case).
This is a small (positive) correction to catke's 
```math
w_\star^3 \sim e^{3/2} \sim J_b h 
```

Also, $w_\star^3$ that enters the tke boundary condition will be slightly higher because of this effect.
It is a bit of a detail compared to just communicating the complete penetrated solar flux as a top flux to catke.

Issue #609, opened 2026-08-28
Field-valued zero_plane_displacement is not validated against the surface-layer height


xkykai:
`validate_zero_plane_displacement(::SimilarityTheoryFluxes, zᵃᵗ)` returns early for anything that is not a `Number`, and the similarity profile computes `Δhᵈ = Δh − d` with no floor. A per-cell displacement field with `d ≥ zᵃᵗ` anywhere therefore produces `log` of a non-positive number in that cell and NaN fluxes, silently, at the first flux computation — where the scalar case throws `ArgumentError` at construction.

```julia
using Oceananigans, NumericalEarth
grid = LatitudeLongitudeGrid(size = (2, 1), longitude = (0, 2), latitude = (0, 1), topology = (Bounded, Bounded, Flat))
d = Field{Center, Center, Nothing}(grid); set!(d, [1.0 40.0])   # 40 m > a 30 m surface-layer height
SimilarityTheoryFluxes(zero_plane_displacement = d)               # accepted; the second cell will NaN
SimilarityTheoryFluxes(zero_plane_displacement = 40.0)            # ArgumentError, as intended
```

Forest classes in `representative_canopy_height` reach 25 m and the Raupach displacement is ~0.7 h, so a 60 m surface cell (zᵃᵗ ≈ 30 m) is close to this limit; a 30 m surface cell would cross it. `validate_zero_plane_displacement` should check `maximum(d) < minimum(zᵃᵗ)` for `Field` values too.

ewquon:
Hmm, #613 should've closed this but was merged into a (now defunct) feature branch instead of main?

PR #613, opened 2026-08-28
Validate Field-valued zero-plane displacements against the surface layer height

xkykai:
Closes #609.

`validate_zero_plane_displacement` returned early for anything but a `Number`, so a per-cell displacement field with `d ≥ zᵃᵗ` in some cell produced `log` of a non-positive number and NaN fluxes there, silently, where a scalar displacement throws `ArgumentError` at construction. The check is now the per-cell clearance `minimum(zᵃᵗ .- d) > 0`, which covers scalar and `Field` displacements against both a scalar and a per-column surface layer height.

```julia
d = Field{Center, Center, Nothing}(grid); set!(d, (λ, φ) -> ifelse(λ < 11, 0.0, 10.0))
fluxes = SimilarityTheoryFluxes(; zero_plane_displacement = d, …)
AtmosphereLandModel(atmosphere, land; atmosphere_land_fluxes = fluxes)   # surface layer height 10 m → ArgumentError
```

Test added to the per-cell roughness/displacement testset in `test_slab_land.jl`.

### DIFF
--- a/src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
+++ b/src/EarthSystemModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
@@ -271,10 +271,9 @@ end
 validate_zero_plane_displacement(flux_formulation, zᵃᵗ) = nothing
 
 function validate_zero_plane_displacement(fluxes::SimilarityTheoryFluxes, zᵃᵗ)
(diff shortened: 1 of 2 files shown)


xkykai:
`test/test_slab_land.jl` passes on CPU with this change (all testsets, including the new `@test_throws` for a displacement field reaching the surface layer height).

PR #616, opened 2026-08-28
Make ESA WorldCover class fractions shares of the block area

### LINKED ISSUE #608: ESA WorldCover class fractions are zero, not missing, where a block has no classified pixel
`class_fractions` returns all-zero fractions for a block whose pixels are all no-data, and the `:<class>_fraction` / `:vegetation_fraction` products inherit that. Open sea inside a published 3° tile therefore lands on a model grid as "0% water, 0% vegetation, 0% built-up", indistinguishable from an unvegetated land cell, while the unpublished all-ocean tiles come through as `NaN`. Over the CONUS box this turned 44,958 ocean cells into bare land until the consumer added `1 − Σ fractions` back as water.

```julia
using NumericalEarth.DataWrangling.WorldCover: class_fractions, majority_class
pixels = zeros(UInt8, 12, 12)          # all no-data
class_fractions(pixels).permanent_water_bodies   # 0.0
majority_class(pixels)                           # 0 (masked to NaN downstream)
```

The majority-class product already masks these blocks (`_mask_uncovered_class!`); the fraction products should agree — either `NaN` fractions for all-no-data blocks (with the conservative regrid normalizing by the valid area, as the categorical path does), or a documented `unmapped_fraction` = 1 − Σ fractions so a consumer can treat it as sea.

xkykai:
Closes #608.

`class_fractions` and `vegetation_fraction` normalized by the number of *classified* pixels, so a block half covered by unmapped sea reported its land classes as if they filled the block, and an all-no-data block reported zeros — open sea inside a published WorldCover tile reached a model grid as "0% water, 0% vegetation", i.e. bare land (44,958 cells over the CONUS box). Per-valid fractions are also not conservative under the area-weighted `regrid!`: averaging blocks with different denominators does not give the target cell's area fraction.

Fractions are now shares of the block area: no-data pixels belong to no class, `Σ class fractions` is the mapped share of the block, and `1 − Σ` is the unmapped (sea) area. The `:landcover_class` mask for cells with total ≤ ½ now means "less than half mapped".

```julia
codes = UInt8[10 10 40; 10 10 0]      # four tree, one crop, one no-data pixel
class_fractions(codes).tree_cover     # 4/6 (was 4/5)
sum(values(class_fractions(codes)))   # 5/6 (was 1)
```

`class_fractions(counts, pixel_count)` / `vegetation_fraction(counts, pixel_count; …)` replace the counts-only methods; `aggregate_landcover` passes `factor^2`. Tests and the example's wording follow the new semantics.

### DIFF
--- a/examples/esa_worldcover_fractions.jl
+++ b/examples/esa_worldcover_fractions.jl
@@ -13,7 +13,7 @@
 # The 10 m `Map` band is *categorical*: its byte value is a class code, not a
 # quantity, so it is never averaged. The ingest counts codes over each aggregated
 # cell to form area fractions. Onto the model grid the fractions are regridded
-# conservatively (area-weighted, so they still sum to one), and the categorical
+# conservatively (area-weighted, so they still sum to the mapped share), and the categorical
 # majority class is the argmax of those fractions — the class covering the most
 # area of each cell, never a blend, so no intermediate code is ever invented.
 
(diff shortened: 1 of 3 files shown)


xkykai:
`test/test_esaworldcover.jl` passes on CPU with the area-share semantics (all testsets, exit 0).

PR #624, opened 2026-08-30
Materialize a time-varying pressure-level geopotential once per state update

### LINKED ISSUE #623: Time-varying pressure-level geopotential is re-interpolated at every probe of the column bisection
`column_fractional_z_index` (`src/Grids/pressure_level_vertical_discretization.jl`) bisects a column of the geopotential for every cell it interpolates, and each probe of that bisection reads `grid.z.geopotential[i, j, k]` through a `ColumnView`.

When the geopotential is a `TimeSeriesInterpolation` — the default for `ERA5HourlyPressureLevels()`, which passes `z = nothing` and builds a `per_column_geopotential_discretization` — every one of those reads is a `FieldTimeSeries[i, j, k, Time(t)]`, carrying its own binary search over `times` and a two-slice blend. Within a step all of them ask for the same Φ(t), so that **time interpolation is repeated for each probe of each cell and returns the same answer every time**.

It sits on the hot path for a nested run: the Davies relaxation targets and the lateral boundary conditions both interpolate the parent `FieldTimeSeries` onto the child, so this runs for every relaxed field at every cell, every stage.

Interpolating a field from an ERA5-like parent (40×40×37) onto a 150×136×50 child, with the geopotential as a `TimeSeriesInterpolation` against the same setup with a static `Field`:

| | `TimeSeriesInterpolation` | static `Field` |
|---|---|---|
| `interpolate!` end-to-end | 121.5 ms | 57.4 ms |
| `column_fractional_z_index` | 109.3 ms | 20.4 ms |

So **roughly half the interpolation cost is re-interpolating Φ in time**. Materializing Φ into a plain `Field` once per state update, rather than reading the series per probe, should recover it. PR to follow.

ewquon:
Closes #623.

The discretization now always stores a plain `Field`: a `TimeSeriesInterpolation` is materialized at construction and recomputed by `materialize_geopotential!` from `update_state!(::PrescribedAtmosphere)`, which `NestedModel.time_step!` reaches on every nested step. Each probe of the column bisection is a single array read again.

This is a `Field` type alias rather than a new struct, so `rnode`, `ColumnView`, `first_above_surface_level`, `Adapt` and `on_architecture` keep working unchanged. `geopotential_data_for_extrema` forwards through `.operand` so `Lz` and the column-mean profile still span every time slice, not just the resident snapshot, and clipping runs before materialization so every snapshot inherits the clipped columns.

End-to-end on the ERA5 → 12 km nested hindcast of `examples/breeze_downscaling_era5.jl` (150×136×50, `Float32`), A100-SXM4-40GB, exclusive node, 400 iterations, arms alternated A/B/A/B:

| | median s/step | |
|---|---|---|
| `main` | 0.0478, 0.0495 | |
| this branch | 0.0368, 0.0359 | **1.34×** |

Within-arm spread is 0.0017 and 0.0009 s against a between-arm gap of 0.0123 s, and the arms' ranges do not overlap. Physics is unchanged: ρ ∈ [0.0947, 1.1857] and max|u| 98.6–101.9 m/s across all four runs.

The **accuracy cost is zero** rather than small. The parent clock advances only inside `time_step!(atmos, Δt)`, which calls `update_state!` immediately, and the child reads `znode` between parent ticks, so the snapshot always holds Φ at the parent clock time — over 500 steps of ERA5 data, `max |materialized − interpolated|` is 0 exactly.

One behavior change: reading `rnode` after hand-setting `clock.time` returns the previous snapshot until a refresh runs, and the existing clock-following test is updated to assert that. Separately, `FieldTimeSeries` defaults to `Clamp()` time indexing, so past the end of the window the heights freeze on the last snapshot rather than extrapolating; that is now pinned by a test.


bischtob:
NUM-232

PR #646, opened 2026-09-03
[CI] Move CDS downloading jobs to weekly schedule

giordano:


glwagner (approved):
Nice! Note, I do not think this is only CDS (Copernicus Data Store) downloading. It tests downloading from a variety of places. Also with this change we can probably add more downloading tests (or should add all of them) but that can be a future PR

.github/workflows/DataDownload.yml
```diff
@@ -0,0 +1,110 @@
...
+env:
+  ECCO_USERNAME: ${{ secrets.ECCO_USERNAME }}
+  ECCO_WEBDAV_PASSWORD: ${{ secrets.ECCO_WEBDAV_PASSWORD }}
+  COPERNICUSMARINE_SERVICE_USERNAME: ${{ secrets.COPERNICUSMARINE_SERVICE_USERNAME }}
+  COPERNICUSMARINE_SERVICE_PASSWORD: ${{ secrets.COPERNICUSMARINE_SERVICE_PASSWORD }}
+  CDSAPI_URL: "https://cds.climate.copernicus.eu/api"
+  CDSAPI_KEY: ${{ secrets.CDSAPI_KEY }}
+  DESTINE_ACCESS_TOKEN: ${{ secrets.DESTINE_ACCESS_TOKEN }}
+  EARTHDATA_USERNAME: ${{ secrets.EARTHDATA_USERNAME }}
+  EARTHDATA_PASSWORD: ${{ secrets.EARTHDATA_PASSWORD }}
```
giordano, line 26: The bot points out that these should be set in the settings for the `test_asterged_downloading` test to actually be run (but when they are not set, that's not run, so nothing breaks)

glwagner: thanks bot 🤖

PR #647, opened 2026-09-03
Add synthetic datasets so dataset code paths are tested offline

giordano:
> `test/synthetic_datasets.jl` defines three tiny analytic datasets whose `download` writes a NetCDF file on first use: a bathymetry, a 3-D ocean with temperature and salinity, and a 2-D atmosphere with the JRA55 variable names. `Field`, `FieldTimeSeries`, `DatasetRestoring`, `regrid_bathymetry` and the prescribed atmosphere, radiation and land components all run their dataset code paths on them without any network.
> 
> `test_synthetic_dataset.jl` drives the shared helpers in `runtests_setup.jl` with the synthetic ocean and checks exact values on the native grid. The checkpointer, radiation, field cache and distributed-bathymetry tests, which used real data only as a plausible forcing, now use the synthetic datasets; the hand-written bathymetry in `test_distributed_utils.jl` is replaced by the shared one.
> 
> The nonzero restoring assertions in `test_dataset_restoring` become approximate: spatially interpolating a constant field is exact only up to roundoff.
> 

### DIFF
--- a/test/runtests_setup.jl
+++ b/test/runtests_setup.jl
@@ -185,7 +185,7 @@ function test_dataset_restoring(arch, dataset, dates, inpainting;
         field = NamedTuple{fldnames}(ntuple(i->CenterField(grid), length(fldnames)))
 
         # A window-averaged product has no node at its first date: the first sits half a window later.
-        clock = Clock(; time = first(var_restoring.field_time_series.times))
+        clock = Clock(; time = first(on_architecture(CPU(), var_restoring.field_time_series.times)))
 
         @allowscalar begin
             @test var_restoring(1, 1,   10, grid, clock, field) ≈ var_restoring.rate
--- a/test/runtests_setup.jl
+++ b/test/runtests_setup.jl
@@ -185,7 +185,7 @@ function test_dataset_restoring(arch, dataset, dates, inpainting;
         field = NamedTuple{fldnames}(ntuple(i->CenterField(grid), length(fldnames)))
 
         # A window-averaged product has no node at its first date: the first sits half a window later.
-        clock = Clock(; time = first(on_architecture(CPU(), var_restoring.field_time_series.times)))
+        clock = Clock(; time = @allowscalar first(var_restoring.field_time_series.times))
 
         @allowscalar begin
             @test var_restoring(1, 1,   10, grid, clock, field) ≈ var_restoring.rate
--- a/test/runtests_setup.jl
+++ b/test/runtests_setup.jl
@@ -188,12 +188,12 @@ function test_dataset_restoring(arch, dataset, dates, inpainting;
         clock = Clock(; time = first(var_restoring.field_time_series.times))
 
         @allowscalar begin
-            @test var_restoring(1, 1,   10, grid, clock, field) == var_restoring.rate
-            @test var_restoring(1, 11,  10, grid, clock, field) == var_restoring.rate / 2
+            @test var_restoring(1, 1,   10, grid, clock, field) ≈ var_restoring.rate
+            @test var_restoring(1, 11,  10, grid, clock, field) ≈ var_restoring.rate / 2
             @test var_restoring(1, 21,  10, grid, clock, field) == 0
             @test var_restoring(1, 80,  10, grid, clock, field) == 0
-            @test var_restoring(1, 90,  10, grid, clock, field) == var_restoring.rate / 2
-            @test var_restoring(1, 100, 10, grid, clock, field) == var_restoring.rate
+            @test var_restoring(1, 90,  10, grid, clock, field) ≈ var_restoring.rate / 2
+            @test var_restoring(1, 100, 10, grid, clock, field) ≈ var_restoring.rate
             @test var_restoring(1, 1,   5,  grid, clock, field) == 0
             @test var_restoring(1, 10,  5,  grid, clock, field) == 0
         end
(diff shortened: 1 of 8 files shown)


test/synthetic_datasets.jl
```diff
@@ -0,0 +1,142 @@
...
+end
+
+# Each process writes into its own directory.
+function Downloads.download(metadata::Metadata{<:Synthetic})
+    for metadatum in metadata
+        path = metadata_path(metadatum)
+        isfile(path) || write_synthetic_netcdf(path, metadatum)
+    end
+    return metadata_path(metadata)
+end
```
giordano, line 110: @glwagner this is what I suggested when we were talking about this offline

test/runtests_setup.jl
```diff
@@ -185,15 +185,15 @@ function test_dataset_restoring(arch, dataset, dates, inpainting;
         field = NamedTuple{fldnames}(ntuple(i->CenterField(grid), length(fldnames)))
 
         # A window-averaged product has no node at its first date: the first sits half a window later.
-        clock = Clock(; time = first(var_restoring.field_time_series.times))
+        clock = Clock(; time = first(on_architecture(CPU(), var_restoring.field_time_series.times)))
```
glwagner: why does this matter, if allowscalar is used below?

test/synthetic_datasets.jl
```diff
@@ -0,0 +1,142 @@
...
+        0.01
+    elseif name == :eastward_velocity
+        5
+    elseif name == :northward_velocity
+        1
+    elseif name == :downwelling_shortwave_radiation
+        300 * cosd(φ)
+    elseif name == :downwelling_longwave_radiation
+        250
+    else # freshwater fluxes
```
glwagner: could be best for future proof to make this explicit

test/test_synthetic_dataset.jl
```diff
@@ -0,0 +1,84 @@
...
+        @testset "Timestepping with DatasetRestoring" begin
+            test_timestepping_with_dataset_restoring(arch, dataset, dates, inpainting)
+        end
+
+        @testset "Dataset cycling boundaries" begin
+            test_cycling_dataset_restoring(arch, dataset, dates, inpainting)
+        end
+
+        @testset "Inpainting algorithm" begin
+            test_inpainting_algorithm(arch, dataset, dates[1], inpainting)
```
glwagner, line 58: does this get tested properly? do we need synthetic data with NaNs? or perhaps the presence of bathymetry does the job.

glwagner (approved):
possibly @simone-silvestri should take a look as well

test/runtests_setup.jl
giordano: Uhm, the commit message says

> `first` on the times of a GPU 
> `FieldTimeSeries` and indexing the nodes of
> a GPU grid are scalar GPU accesses.

which suggests the two things are orthogonal.

test/test_synthetic_dataset.jl
giordano: I didn't really know how to answer this one, so I asked help from home:

> Verified. Yes, line 59 tests the inpainting properly, and it is the NaN in the synthetic data that does the job, not bathymetry.
> 
> Measured on the 20×20×10 test grid, per variable:
> 
> | Data | Inpainting | Zeros | NaN |
> |---|---|---|---|
> | with land (656 NaN of 2592 native cells) | none | 0 | 1150 |
> | with land | 1 pass | 150 | 0 |
> | with land | unlimited | 0 | 0 |
> | without land (0 NaN native) | none, 1 pass, or unlimited | 0 | 0 |
> 
> So with the synthetic land, one pass fills the coastal ring and leaves 150 interior land cells that the regridder then zeroes, which is what `any(== 0)` catches, and unlimited passes fill every cell, which is what `all(!= 0)` catches. In the counterfactual with the land removed (same dataset, `synthetic_land` redefined to return false in a fresh directory), one pass leaves no zeros, so the `any(== 0)` assertion would fail. The NaN land is therefore load-bearing for the test.
> 
> Bathymetry plays no role here: the helper regrids onto a plain `LatitudeLongitudeGrid` with no immersed boundary, and `set!` from a `Metadatum` reads the file's NaN mask directly. The synthetic bathymetry only enters through `regrid_bathymetry`, which this test never calls. The two are consistent by construction, though, since the bathymetry and the ocean NaN both come from the same `synthetic_land` function.

glwagner: Claude is good. they thought of it.

test/synthetic_datasets.jl
giordano: Done in 8f0bf9c7da27848ee372cb09815163ea1d8677d0 (also, I decided to throw an error in the `else` branch, to be on the safe side)

test/runtests_setup.jl
giordano: Actually, it turns out that the bot was overzealous following this rule: https://github.com/NumericalEarth/NumericalEarth.jl/blob/c35ad718df5071cd9dcc25160231e89aecce35fb/.claude/rules/testing-rules.md?plain=1#L26 Maybe we can avoid the copy to CPU, testing it

giordano: Ok, copy to CPU avoided in bd4f76d32ce004def7e726c4ec7955d396c513f3

simone-silvestri (approved):
So I guess this already speeds up the testing?

giordano:
This + all the other PRs of the stack should help with reliability, because we don't download anything on PRs anymore. As a side effect tests should also be a little bit shorter because we won't have anymore the serial download phase (with all the failure chances attached to that).

PR #648, opened 2026-09-03
Make the default test run offline; real-data tests move to a nightly workflow

giordano:
> `test/synthetic_datasets.jl` defines three tiny analytic datasets whose
> `download` writes a NetCDF file on first use: a bathymetry, a 3-D ocean
> with temperature and salinity, and a 2-D atmosphere with the JRA55
> variable names. `Field`, `FieldTimeSeries`, `DatasetRestoring`,
> `regrid_bathymetry` and the prescribed atmosphere, radiation and land
> components all run their dataset code paths on them without any network.
> 
> `test_synthetic_dataset.jl` drives the shared helpers in
> `runtests_setup.jl` with the synthetic ocean and checks exact values on
> the native grid. The checkpointer, radiation, field cache and
> distributed-bathymetry tests, which used real data only as a plausible
> forcing, now use the synthetic datasets; the hand-written bathymetry in
> `test_distributed_utils.jl` is replaced by the shared one.
> 
> The nonzero restoring assertions in `test_dataset_restoring` become
> approximate: spatially interpolating a constant field is exact only up
> to roundoff.
> 
> Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,3 +1,9 @@
+# The default run is offline: the data directory is a fresh temporary directory, and any file
+# that lands there is an accidental download. Set `NUMERICALEARTH_TEST_REMOTE_DATA=true` to
+# also run the tests that need real remote datasets, in the regular data directory.
+remote_data = parse(Bool, get(ENV, "NUMERICALEARTH_TEST_REMOTE_DATA", "false"))
+remote_data || (ENV["NUMERICALEARTH_DATA_DIRECTORY"] = mktempdir())
+
 # Common test setup file to make stand-alone tests easy
 include("runtests_setup.jl")
 include("download_utils.jl")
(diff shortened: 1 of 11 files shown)


test/runtests.jl
```diff
@@ -1,3 +1,9 @@
+# The default run is offline: the data directory is a fresh temporary directory, and any file
+# that lands there is an accidental download. Set `NUMERICALEARTH_TEST_REMOTE_DATA=true` to
+# also run the tests that need real remote datasets, in the regular data directory.
+remote_data = parse(Bool, get(ENV, "NUMERICALEARTH_TEST_REMOTE_DATA", "false"))
```
glwagner, line 4: this is a good approach

glwagner (approved):
possibly we should open an issue and link to this PR about creating some docs for how to run the tests (like we have for Breeze). this can be a future PR.

navidcy (approved):

PR #651, opened 2026-09-03
Drive the forcing-only tests with the synthetic datasets

giordano:
> Eight tests used JRA55, ETOPO, ECCO or EN4 only as a plausible forcing,
> bathymetry or initial condition and asserted nothing about the real
> values. They now use the synthetic datasets and leave the remote-data
> list in `runtests.jl`, so they run on every pull request again.
> 
> The synthetic bathymetry gains land under the two north poles of the
> default `TripolarGrid` and a south polar cap. The barotropic solver is
> singular at a wet pole, and sea ice over a wet south polar ocean
> exchanges enough salt to break the tracer budget's conservation check.
> 
> `test_mangling` stays a remote-data test: it asserts the layout of
> ECCO's Face-located `v_velocity` file.
> 

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -40,15 +40,6 @@ remote_data_tests = ["test_bathymetry",
                      "test_soilgrids",
                      "test_dataset_region",
                      "test_diagnostics_1",
-                     # Use real data only as forcing or initial condition; to be moved onto the synthetic datasets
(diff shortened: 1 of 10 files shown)


glwagner (approved):
I think this is a good approach! It makes sense!

cc @simone-silvestri may want to comment

simone-silvestri (approved):

PR #652, opened 2026-09-03
Retry every raw dataset download and route AVISO and GloFAS through download_cache

giordano:
> `download_with_retries` moves from `earthdata.jl` into `DataWrangling.jl` and forwards its keyword arguments to `Downloads.download`, so the ten datasets that fetch a plain URL (ETOPO, JRA55, EN4, ECCO, GEBCO, ORCA, SoilGrids, SeaWiFS, IBCSO, IBCAO) retry a failed download three times with backoff instead of failing on the first transient error.
> 
> AVISO and GloFAS called `@get_scratch!` directly and so ignored `NUMERICALEARTH_DATA_DIRECTORY`; they now use `download_cache` like every other dataset module.
> 


src/DataWrangling/DataWrangling.jl
```diff
@@ -156,6 +157,26 @@ function netrc_permission_file(username, password, machine, dir)
...
+function download_with_retries(url, path; attempts = 3, description = "Download", kw...)
+    for attempt in 1:attempts
+        try
+            Downloads.download(url, path; kw...)
+            return path
+        catch error
+            rm(path, force = true)
+            attempt == attempts && rethrow()
+            @warn "$description failed (attempt $attempt of $attempts); retrying..." url error
+            sleep(2attempt)
```
glwagner, line 194: whats' a good time-scale for this? also, exponential backoff?

glwagner (approved):

src/DataWrangling/DataWrangling.jl
giordano: This is the same as it's now, the function was simply moved around.  Exponential backoff is definitely an option, if you prefer

glwagner: your call!

PR #653, opened 2026-09-03
Inject the CDS retrieve function so the download paths are tested offline

giordano:
@glwagner this is another thing we discussed: mock the download functionalities, so that they can be tested offline, without accessing the network

> Every Copernicus Climate Data Store download in the CDSAPI and CopernicusClimateDataStore extensions takes a `retrieve` keyword, the function `(product, request, path)` that answers a request, defaulting to the backend's own `retrieve`. Tests pass a function that records the request and writes a synthetic NetCDF, and the request construction, temporary-file handling, per-date splitting and yearly concatenation run end to end without credentials or network.
> 
> `test_cds_downloading.jl` keeps only the testsets that contact the CDS; its offline testsets move to `test_cds.jl`, which runs in the default suite, together with the injected-retrieve tests for ERA5 single-level and GloFAS. The ERA5-Land one lives with the CopernicusClimateDataStore extension tests.
> 


test/test_cds.jl
giordano, line 1: Specifically, all the tests in this file, moved out from `test_cds_downloading.jl`, don't require network access, and when they need to "retrieve" some data a fake retriever function is supplied.

glwagner: interesting solution! it seems a bit noisier than a dispatch-based solution, but perhaps also a little simpler?

test/test_cds.jl
```diff
@@ -0,0 +1,1251 @@
...
+end
+
+@testset "ERA5 CDSAPIExt download through an injected retrieve" begin
+    dataset = ERA5HourlySingleLevel()
+    dates = [DateTime(2005, 2, 16, 6), DateTime(2005, 2, 16, 18)]
+    requests = []
+
+    # Stands in for the CDS: records the request and answers it with a file holding every
+    # datetime of the request's year × month × day × time product.
+    function fake_retrieve(product, request, path)
```
glwagner, line 1152: 👍

ext/NumericalEarthCDSAPIExt/copernicus_land_albedo.jl
```diff
@@ -3,7 +3,7 @@
 ##### construction, extraction, and repacking live in the CopernicusLandAlbedo module.
 #####
 
-Downloads.download(metadata::CopernicusAlbedoDatasetMetadata; kwargs...) =
+Downloads.download(metadata::CopernicusAlbedoDatasetMetadata; retrieve=CDSAPI.retrieve, kwargs...) =
     download_ten_day_albedo!(metadata; kwargs...) do request, path
-        retrieve_with_retries(ALBEDO_CDS_PRODUCT, request, path)
+        retrieve_with_retries(ALBEDO_CDS_PRODUCT, request, path; retrieve)
```
glwagner, line 8: I guess "retrieve" is pre-existing terminology, but it is interesting that we have both a :"download" and a "retrieve" which are synonyms...

glwagner (approved):
I see no cause not to approve, my only question is what we might do for other products that do not have `retrieve`. Perhaps we just need an interface function.

test/test_cds.jl
```diff
@@ -0,0 +1,1251 @@
...
+    @testset "foreach_nc — zip path extracts and visits each .nc" begin
+        mktempdir() do tmp
+            # Build a ZIP fixture containing two .nc files (and a non-.nc file
+            # that should be ignored).
+            nc1 = joinpath(tmp, "a.nc"); touch(nc1)
+            nc2 = joinpath(tmp, "b.nc"); touch(nc2)
+            other = joinpath(tmp, "readme.txt"); touch(other)
+
+            zip_path = joinpath(tmp, "bundle.zip")
+            run(`zip -j -q $zip_path $nc1 $nc2 $other`)
```
giordano: Bad bot, should have used p7zip here.  ***Edit***: the bot didn't do anything wrong, just copied the test over from `test_cds_downloading.jl`!

ext/NumericalEarthCDSAPIExt/copernicus_land_albedo.jl
giordano: Unfortunately I think that the name similarity is sort of by construction: "download" is the high-level functionality, then you need a mechanism to practically materialize (fetch/retrieve, in the downloading context) the data.

test/test_cds.jl
giordano: This is a more functional approach, rather than dispatch-based, but I think that the nice thing is that it integrates with the dispatched-based interface used by these function because it's orthogonal to that.

PR #666, opened 2026-09-06
Speed up tracer budget test

giordano:
This seems to almost halve the time to run `test_tracer_budget` locally for me, see the individual commits for an explanation of how they help.

### DIFF
--- a/test/test_tracer_budget.jl
+++ b/test/test_tracer_budget.jl
@@ -66,7 +66,7 @@ function test_tracer_budget(coupled_model, Sᵒᶜ, Δt, nsteps; heat_rtol, fres
         # Heat content changes by the surface heat flux plus the enthalpy carried by the freshwater
         # (rain − evaporation at SST). The live Tᴺ Jʷ exchange cancels the z-star ambient carry, so
         # the freshwater's own enthalpy Σᵢ Tᵢ Jʷᵢ is what remains.
-        heat_content_tendency = sum(ρᵒᶜ * cᵒᶜ * ΔVT)
+        heat_content_tendency = ρᵒᶜ * cᵒᶜ * sum(ΔVT)
         expected_heat_content_tendency = (previous_radiative_rate - previous_heat_flux + previous_enthalpy) * last_Δt
         @test isapprox(heat_content_tendency, expected_heat_content_tendency; rtol=heat_rtol)
 
--- a/test/test_tracer_budget.jl
+++ b/test/test_tracer_budget.jl
@@ -118,7 +118,7 @@ end
 
             Δt = 605seconds
             Sᵒᶜ = 35 # reference salinity [psu]
-            free_surface = SplitExplicitFreeSurface(substeps=20)
+            free_surface = SplitExplicitFreeSurface(substeps=5)
 
             # Without shortwave penetration
             @testset "Surface-only fluxes" begin
--- a/test/test_tracer_budget.jl
+++ b/test/test_tracer_budget.jl
@@ -2,6 +2,7 @@ include("runtests_setup.jl")
 
 using CUDA: @allowscalar
 using Oceananigans.AbstractOperations: KernelFunctionOperation
+using Oceananigans.Diagnostics: NaNChecker, nan_detected
 using Oceananigans.Grids: MutableVerticalDiscretization
 using Oceananigans.Operators: volume
 using Oceananigans.Units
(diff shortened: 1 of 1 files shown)


test/test_tracer_budget.jl
```diff
@@ -66,7 +66,7 @@ function test_tracer_budget(coupled_model, Sᵒᶜ, Δt, nsteps; heat_rtol, fres
         # Heat content changes by the surface heat flux plus the enthalpy carried by the freshwater
         # (rain − evaporation at SST). The live Tᴺ Jʷ exchange cancels the z-star ambient carry, so
         # the freshwater's own enthalpy Σᵢ Tᵢ Jʷᵢ is what remains.
-        heat_content_tendency = sum(ρᵒᶜ * cᵒᶜ * ΔVT)
+        heat_content_tendency = ρᵒᶜ * cᵒᶜ * sum(ΔVT)
```
giordano, line 77: In a nutshell, Julia can't infer the type of `ρᵒᶜ * cᵒᶜ * ΔVT` and so it has to dynamic dispatch every single time.  Hoisting the constants out of the sum works around the issue because the type of `ΔVT` is inferred alright.

test/test_tracer_budget.jl
```diff
@@ -118,7 +118,7 @@ end
 
             Δt = 605seconds
             Sᵒᶜ = 35 # reference salinity [psu]
-            free_surface = SplitExplicitFreeSurface(substeps=20)
+            free_surface = SplitExplicitFreeSurface(substeps=5)
```
giordano, line 129: I'm letting the bot investigate the possibility of making the case `substeps + 2 > halo` not very slow directly in Oceananigans, but as far as I understand the change proposed here doesn't affect test coverage and makes the test lighter regardless of upstream improvements. ***Edit***: https://github.com/CliMA/Oceananigans.jl/pull/5934

giordano:
Uhm, the relative impact here on CPU CI seems to be limited (19m -> 17m, 10%) compared to what I observed locally (7m -> 4 m, 40%), but an improvement nonetheless.  ***Edit***: on GPU [697.74s](https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/33990505579/job/101371769884#step:13:544) -> [376.82s](https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/34031234639/job/101481000028?pr=666#step:13:533), 45%

test/test_tracer_budget.jl
navidcy: oh wow! tagging @taimoorsohail in case similar type of type inference (or not inference) are elsewhere in the diagnostics.

would the same issue be present if it was

```Julia
Integral(ρᵒᶜ * cᵒᶜ * ΔVT)
```

you think?

navidcy (approved):

test/test_tracer_budget.jl
navidcy: No, on second thought I think it would be the same issue....

giordano: Short answer, yes, there's the same issue, below are some numbers and more verbose details from the bot:

> Yes, it has exactly the same problem, and it is slightly worse: even `Integral(c)` of a plain field built inside the function re-infers on every call for the first five calls (about 0.1 s each). Building the `Integral` outside and calling `Field(I)` or `compute!` inside is instant from the first call.
> 
> | Expression built inside the function | Calls 1–5 | Call 6 |
> |---|---|---|
> | `sum(2.0 * c)` | 0.07–0.6 s each | 0.0001 s |
> | `2.0 * sum(c)` | 0.04 s once, then 0 | 0 |
> | `first(Field(Integral(2.0 * c)))` | ~0.11 s each | 0.0001 s |
> | `first(Field(Integral(2.0 * c, dims=(1, 2))))` | ~0.11 s each | 0.0001 s |
> | `first(Field(Integral(c)))` | ~0.09 s each | 0.0001 s |
> | `Integral` prebuilt outside, `Field(I)` inside | 0 | 0 |
> | `Field(Integral)` prebuilt, `compute!` inside | 0 | 0 |
> 
> So the trigger is constructing the operation (any `AbstractOperation`, including `Integral`) in the same function that reduces it, not the multiplication by a constant specifically. This is also why the test's existing `heat_rate`, `enthalpy_rate`, and `volume_rate` were never a problem: they are built once before the loop, and only `Field(heat_rate)` runs per iteration. If you preferred `Integral` over the hoisted `sum` for the heat content, the right form would be to build `Integral(ρᵒᶜ * cᵒᶜ * ΔVT)` once next to the other rates and `compute!` it inside the loop, which would be equally fast.

test/test_tracer_budget.jl
simone-silvestri: we need to make sure the test case is stable and does not produce NaNs with this change (most likely, the test grid is coarse)

simone-silvestri (approved):

test/test_tracer_budget.jl
giordano: Would using the NaN checker be sufficient? Also, isapprox by default returns false in preference of NaNs, so the isapprox checks would fail if there were NaNs around, right?

giordano: @simone-silvestri does 1ae1b4ea5bfa470df488bd6570a013d1015635d3 address your concern?

simone-silvestri (approved):

giordano:
Hopefully we'll see more improvements when https://github.com/CliMA/Oceananigans.jl/pull/5934 and https://github.com/CliMA/Oceananigans.jl/pull/5936 make it into a stable release compatible with NumericalEarth.

PR #667, opened 2026-09-06
Allow Breeze 0.10

glwagner:
Widens the Breeze compat to include 0.10, which carries the merged AIVA split-pairing fixes (Breeze #902) and is the version the AR campaign envs pin. One line.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #668, opened 2026-09-06
Key the Breeze energy BC by version: ρs on Breeze ≥0.10, ρe before

glwagner:
Breeze 0.10 renamed static energy e → s (Breeze #926) and the energy-BC interface key followed; unknown boundary_conditions keys are silently dropped (Breeze #956), so the ρe-keyed coupling bottom flux and nested lateral θ values vanish and net_fluxes crashes unwrapping a default no-flux BC. Key by pkgversion(Breeze). Companion to the Breeze-0.10 compat PR — compat alone lets environments resolve into this crash.

### DIFF
--- a/ext/NumericalEarthBreezeExt/breeze_atmosphere_simulation.jl
+++ b/ext/NumericalEarthBreezeExt/breeze_atmosphere_simulation.jl
@@ -26,6 +26,10 @@ function merge_fbcs(coupling::FieldBoundaryConditions, user::FieldBoundaryCondit
                                    immersed = pick(coupling.immersed, user.immersed))
 end
 
+# Breeze 0.10 renamed static energy e → s (Breeze #926), and the energy-BC interface key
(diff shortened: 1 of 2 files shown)


glwagner:
Superseded in shape by Breeze#974, which adds `ρE` — a formulation-agnostic energy key — and validates BC/forcing names at construction (closing Breeze#956, the silent-drop bug this PR works around).

I verified NumericalEarth's exact coupling pattern against that branch (lateral Value BCs and a bottom energy Flux BC merged on one key, potential-temperature formulation): the bottom flux wraps for the `cᵖᵐ` conversion, the lateral values pass through unconverted (300.0 preserved), and a stale `ρe` key now raises instead of vanishing.

So once Breeze#974 lands and tags, this PR should become `ρE` unconditionally rather than a `pkgversion` gate — one key that works on both formulations and doesn't move again. Holding it until then; the Breeze-0.10 compat half is independent and unaffected (#667).

PR #672, opened 2026-09-10
Preserve simulation controls when attaching nested radiation

masonlee277:
Attaching radiation to a nested atmosphere rebuilt its `Simulation` with only `Δt`. This discarded configured stop limits, callbacks, and other simulation controls.

Keep the existing child/nest reconstruction and replace only `atmosphere.model` with `@set`, matching the non-nested materialization path in this file.

The regression extends `test_breeze_coupling.jl` and checks that stop limits, callback ownership, the shared clock, and the radiation-flux reference survive materialization.

Validation: CPU `test/test_breeze_coupling.jl`: 51 passed, two existing tests marked broken. All five ExplicitImports checks passed. GPU execution was not tested.


simone-silvestri (approved):

masonlee277:
Rebased onto current `main` (`56223724`) and taken out of draft.

`test/test_breeze_coupling.jl` passes in full on CPU at this head, with zero failures; the new testset is 8/8.

For context on why the nested branch needed this: the non-nested method immediately above already rebuilds with `@set atmosphere.model.radiation = ...`, so this only makes the nested case behave the same way. Constructing a fresh `Simulation` was discarding `stop_time`, `stop_iteration`, `wall_time_limit`, `align_time_step` and every registered callback.

PR #674, opened 2026-09-10
Add conservative surface flux restoring

taimoorsohail:
Surface salinity restoring can carry a nonzero global mean and drift total ocean salt when the model mean differs from the target climatology. This adds `ConservativeSurfaceFluxRestoring`, which materializes the restoring flux and removes its wet-area mean before the boundary condition applies it.

The accompanying callback initializes the corrected field at the start of `run!` and refreshes it once per timestep. The integration test exercises both supported ocean timesteppers and verifies that restoring changes the salinity pattern while conserving volume-integrated salt on CPU and GPU test jobs.

Validation: `git diff --check`. Repository policy delegates test execution to GitHub CI.

### DIFF
--- a/src/DataWrangling/restoring.jl
+++ b/src/DataWrangling/restoring.jl
@@ -270,6 +270,71 @@ function BoundaryConditions.getbc(sf::SurfaceFluxRestoring, i, j, grid, clock, f
     return - G * Δz
 end
 
+"""
+$(TYPEDSIGNATURES)
+
+Wrap a surface flux and store a corrected flux whose area-weighted mean over wet cells is zero.
+"""
+struct ConservativeSurfaceFluxRestoring{R, F, M} <: Function
+    flux :: R
+    corrected_flux :: F
+    mean_flux :: M
+end
+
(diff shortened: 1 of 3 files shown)


src/DataWrangling/restoring.jl
```diff
@@ -270,6 +270,71 @@ function BoundaryConditions.getbc(sf::SurfaceFluxRestoring, i, j, grid, clock, f
...
+@kernel function _materialize_surface_flux!(buffer, flux, grid, clock, fields)
+    i, j = @index(Global, NTuple)
+    @inbounds buffer[i, j, 1] = BoundaryConditions.getbc(flux, i, j, grid, clock, fields)
+end
+
+"""
+$(TYPEDSIGNATURES)
+
+Evaluate `restoring.flux` from the current model state and subtract its
+area-weighted mean over wet cells.
```
simone-silvestri, line 308:
```suggestion
Evaluate `restoring.flux` from the current model state and subtract its area-weighted mean over wet cells.
```

PR #678, opened 2026-09-12
Write output already low-pass filtered in time: LowPassFilter

enri66:
## What this adds

`LowPassFilter(interval; window = 5days, cutoff = 40hours)` is a schedule for `JLD2Writer` that writes its
outputs low-pass filtered in time. Each frame is a Lanczos-weighted average of the output over a window
centered on the frame time ([Duchon 1979](https://doi.org/10.1175/1520-0450(1979)018%3C1016:LFIOAT%3E2.0.CO;2)).
The use case is long regional runs with tides: writing daily or 5-day output with the tides already removed,
instead of writing hourly output and filtering it afterwards.

```julia
simulation.output_writers[:daily] = JLD2Writer(model, outputs; filename = "daily",
                                               schedule = LowPassFilter(1day))

simulation.output_writers[:five_day] = JLD2Writer(model, outputs; filename = "five_day",
                                                  schedule = LowPassFilter(5days; window = 10days, cutoff = 10days))
```

Amplitude passed by each (5-minute sampling):

| | M2 | S2 | K1 | O1 | 2 d | 5 d | 10 d | 27.6 d |
|---|---|---|---|---|---|---|---|---|
| daily: 5 d window, 40 h cutoff | 0.05% | 0.01% | −0.45% | −1.1% | 78% | 101% | 100% | 100% |
| 5-day: 10 d window, 10 d cutoff | 0.0% | 0.0% | 0.0% | 0.0% | 0.0% | 5% | 56% | 93% |

## How it works

- Frames fall on multiples of `interval`, are stamped with their center time, and are written `window / 2`
  after it.
- Each output keeps `floor(window / interval) + 1` running sums (6 for the daily example, 3 for the 5-day
  one), advanced every time step with weights `sinc(2τ / cutoff) sinc(2τ / window) Δt` and normalized by
  their own sum, so variable time steps are handled and a constant passes exactly.
- Nothing is checkpointed. Frames whose window starts before the first time step are skipped, so a run picked
  up from a checkpoint taken one window before the end of the previous run continues the frame sequence.
- It plugs into the same Oceananigans machinery as `AveragedTimeInterval` (`time_average_outputs`,
  `add_dependency!`), so file splitting and output indices work unchanged.

## Points for review

- Oceananigans stamps every frame with the clock time, so `write_output!` for this schedule calls
  Oceananigans' own write and then restamps the frame with its center time. If Oceananigans schedules could
  supply the output time, that step would go away.
- Oceananigans writes every output at iteration 0 regardless of its schedule; frames are written only once
  their window is complete.
- `grid`, `location` and `indices` forward to the filtered output's operand: the JLD2 writer asks for them
  when it sets up the file, as for `WindowedTimeAverage`.
- I searched Oceananigans' `OutputWriters` (only the boxcar `WindowedTimeAverage`) and NumericalEarth's
  `Diagnostics` for existing time filtering and found none.
- JLD2 output and numeric clocks only; NumericalEarth does not use `NetCDFWriter` or `DateTime` clocks.

## Tests

`test/test_low_pass_filter.jl`:

1. A slow signal carrying an M2 and an O1 tide is sampled every 10 minutes for 20 days; daily and weekly
   frames land on their center times and recover the slow signal to within 0.02.
2. A run picked up from a day-5 checkpoint, appending to the same files, writes frames bitwise identical
   to an uninterrupted run.

## Validation

A 140 × 92 × 24 Mid-Atlantic Bight configuration, forced by 10 TPXO10 tidal constituents at the open
boundaries and as a body force, run for 15 days with CATKE and a variable time step (up to 4 minutes).
Sea surface height was written hourly, daily (`LowPassFilter(1day)`) and every 5 days
(`LowPassFilter(5days; window = 10days, cutoff = 10days)`).

- The frames written during the run agree with the same filter applied afterwards to the hourly output:
  4.1 × 10⁻⁵ m rms difference for the daily frames (the filtered signal is 4.1 mm rms) and
  1.5 × 10⁻⁵ m for the 5-day frames. The difference comes from sampling every time step versus every hour.
- A least-squares fit of M2, S2, N2, K1, O1 and Q1 to the hourly output (0.35 m rms) leaves 5.7 × 10⁻⁴ m rms
  after the daily filter, so 0.16% of the tide leaks into the daily frames.

### DIFF
--- a/docs/src/NumericalEarth.bib
+++ b/docs/src/NumericalEarth.bib
@@ -515,3 +515,13 @@
   year={1977},
   doi={10.1175/1520-0485(1977)007<0952:IMITUO>2.0.CO;2}
 }
+
+@article{duchon1979lanczos,
+  title={Lanczos filtering in one and two dimensions},
+  author={Duchon, Claude E},
+  journal={Journal of Applied Meteorology},
+  volume={18},
+  number={8},
+  pages={1016--1022},
+  year={1979}
+}
--- a/src/Diagnostics/Diagnostics.jl
+++ b/src/Diagnostics/Diagnostics.jl
@@ -4,10 +4,13 @@ export MixedLayerDepthField, MixedLayerDepthOperand
 export meridional_heat_transport
 export net_ocean_heat_flux, sea_ice_ocean_heat_flux, atmosphere_ocean_heat_flux, frazil_heat_flux,
        net_ocean_freshwater_flux, sea_ice_ocean_freshwater_flux, atmosphere_ocean_freshwater_flux
+export LowPassFilter
 
+using DocStringExtensions: TYPEDSIGNATURES
+using JLD2: jldopen
 using KernelAbstractions: @index, @kernel
 
-using Oceananigans: Oceananigans
+using Oceananigans: Oceananigans, AbstractDiagnostic, AbstractModel
 using Oceananigans.AbstractOperations: Integral
 using Oceananigans.Architectures: architecture
 using Oceananigans.BoundaryConditions: DiscreteBoundaryFunction, FieldBoundaryConditions, fill_halo_regions!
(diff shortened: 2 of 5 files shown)


enri66:
Closing this for now. `LowPassFilter` doesn't depend on anything NumericalEarth-specific either —
every import it needs (`AbstractSchedule`, `IterationInterval`, `JLD2Writer`, `fetch_output`,
`AbstractDiagnostic`, `AbstractModel`) is a core Oceananigans abstraction. It belongs in Oceananigans
itself rather than here, since it's a general-purpose output-writing capability with nothing
domain-specific about it.

No action needed on this one — closing rather than leaving it open in a shape that's about to change.
Will follow up upstream in Oceananigans instead.

PR #679, opened 2026-09-12
Add tidal forcing: astronomical body force and TPXO10 boundary conditions

enri66:
## What this adds

Tidal forcing for regional ocean simulations, as the two halves a tidal run needs:

- `TidalHarmonics(reference_date; constituents, ramp_time)` resolves the astronomy — frequencies,
  equilibrium arguments and nodal corrections — for up to ten constituents.
- `tidal_forcing(harmonics)` is the astronomical body force on `u` and `v`, `+g∇η_eq`.
- `tidal_boundary_conditions(grid, harmonics, dataset)` drives the barotropic tide through the
  lateral boundaries from a tidal atlas: Flather conditions on `U` and `V`, radiation on `η`.
- `TPXO10Atlas()` reads the TPXO10-atlas-v2 harmonic constants.

```julia
harmonics = TidalHarmonics(DateTime(2019, 4, 1); ramp_time = 1day)
tide = tidal_boundary_conditions(grid, harmonics, TPXO10Atlas())

                        forcing = tidal_forcing(harmonics),
                        boundary_conditions = merge(tide, (u = u_bcs, v = v_bcs)))
```

The body force and the boundary forcing read one `TidalHarmonics`, so they cannot drift out of phase
with each other. If they used different nodal factors or equilibrium arguments, the astronomical tide
and the boundary tide would beat against one another at a rate no amplitude tuning can fix.

## How it works

- The constituent table carries each constituent's equilibrium amplitude and the multiples of the
  astronomical angles that make up its equilibrium argument. The frequencies are the time derivative
  of those same arguments rather than a second column of numbers, so a frequency and its phase cannot
  disagree. The ten constituents are the eight primary astronomical ones plus `Mf` and `Mm`; a shelf
  generates the compound and overtides internally, so forcing them at the boundary as well would
  double-count them.
- The body force is the discrete gradient of the equilibrium elevation, taken over the same cells as
  the model's pressure gradient (`∂xᶠᶜᶜ`, as `BarotropicPotentialForcing` does), so a resting ocean
  with `η = η_eq` stays at rest, and it works on any grid whose nodes carry longitude and latitude.
- The atlas window is read onto its own `LatitudeLongitudeGrid` — TPXO's C-grid maps onto an
  Oceananigans C-grid cell for cell, `lon_u` being the western faces of the elevation cells and
  `lat_v` their southern faces. Land, which the atlas encodes as zero, is filled with
  `inpaint_mask!`, and the boundary nodes are sampled with `Fields.interpolate`, so a model boundary
  point that the atlas puts on land takes the nearest wet constants.
- Harmonic constants stay complex through the whole path, which avoids interpolating a phase (the
  average of 359° and 1° is 180°, the opposite of the right answer). The boundary condition evaluates
  `real(f z e^{i(ωt + V + u)})`, with the constants on the architecture of the grid.
- Transports enter the Flather condition unconverted, preserving the tidal mass flux where the atlas
  and the model disagree about depth.

## TPXO is licensed

TPXO cannot be downloaded automatically: it is distributed under an academic license from
https://www.tpxo.net. `TPXO10Atlas` therefore has no `download` method, and reports in one sentence
where the files are expected if they are missing. The tests use synthetic files in the atlas's own
format instead, so none of this needs the real atlas.

## Tests

`test/test_tides.jl`, all on synthetic data:

1. The frequencies match the NOAA CO-OPS published angular speeds to 1e-7 relative — an independent
   source, and a check on the astronomical arguments the frequencies are derived from. The phases of
   harmonics built six hours apart differ by exactly the frequencies times six hours, which is what
   keeps the body force and the boundary tide on one clock.
2. A closed basin under the M2 body force fills to the equilibrium tide: least-squares gain 1.02 with
   correlation 0.9996 against `f A cos²φ cos(ωt + V + u + 2λ)` written out independently in the test.
3. A basin far shorter than the tidal wavelength, forced through all four boundaries from a synthetic
   atlas, co-oscillates with the atlas tide: interior amplitude 0.52 m against the atlas's 0.50 m,
   phase lag 0.5°. This exercises the file layout, the millimeter and centimeter²/second units, the
   phase convention and the Flather pair together — a unit error in the transports would show up as
   metres of error in the interior.

## Validation

A 140 × 92 × 24 Mid-Atlantic Bight configuration at 1/12°, CATKE with a constant bottom drag
coefficient of 3e-3 and no stratification, forced both at the open boundaries and as a body force.
Two runs, each repeated with this code:

**M2 alone, 10 days**, harmonically analyzed against TPXO over 9669 wet interior cells, with a
six-cell buffer next to the open boundaries excluded:

| depth class | cells | model A | TPXO A | cRMS | cRMS / A |
|---|---|---|---|---|---|
| deep > 1000 m | 7454 | 0.408 | 0.413 | 0.0100 | 0.02 |
| slope 200–1000 m | 318 | 0.426 | 0.419 | 0.0156 | 0.04 |
| shelf 50–200 m | 987 | 0.448 | 0.418 | 0.0549 | 0.13 |
| shelf < 50 m | 910 | 0.563 | 0.469 | 0.3424 | 0.73 |

Domain complex RMS 0.1069 m against a mean amplitude of 0.419 m, amplitude bias +0.008 m, phase bias
+0.46°, phase RMS 8.0°.

**All ten constituents, 15 days**, compared with the full TPXO prediction in the time domain, which
needs no constituent separation because the forcing is known. Over 9674 wet interior cells and a
12.9-day record, `rms(model − TPXO prediction) / rms(prediction)` has a median of 0.026 and a 90th
percentile of 0.114. At the two open-coast gauges the modeled elevation correlates with the TPXO
prediction at 0.996 (Atlantic City) and 0.986 (Duck), and with the NOAA prediction at 0.992 and 0.977,
against a signal of 0.44 m and 0.36 m rms.

Both runs were repeated with this code rather than quoted from the implementation they were first made
with, because the atlas sampling genuinely changed — inpainting and `Fields.interpolate` in place of a
masked bilinear interpolation with a nearest-wet fallback. The skill did not move: the M2 domain
complex RMS is 0.1069 m against 0.1070 m before, every depth class is unchanged, and the ten-constituent
median ratio and 90th percentile are identical to three decimals (0.026 and 0.114). The ported astronomy
agrees with the earlier one to 2.7e-6 in frequency, 0.02° in phase and exactly in amplitude, the
frequency difference being the port's own gain in accuracy, since `Mf` and `Mm` were tabulated to five
digits before.

The shelf figure is a resolution and bottom-boundary-layer question at 12 km, not a forcing one: a
uniform ν_z = 0.1 scores better there but is not a defensible setting, since it would also erode
surface stratification.

## Points for review

- The open boundary conditions on the baroclinic `u` and `v` are left to the caller: they are a
  choice about the interior flow, not about the tide. Without them the barotropic tide still enters,
  but the depth-mean flow and the 3-D flow disagree at the boundary.
- Self-attraction and loading is not included. The scalar approximation is a force `+gβ∇η` on the
  model's own elevation; the better route for a regional model is per-constituent SAL amplitude and
  phase maps added to the equilibrium tide, which the harmonic sum already accommodates. Either is
  additive to this, and for a domain whose tide arrives through the boundaries an assimilative atlas
  already carries the real SAL there.
- Every lateral boundary gets the tidal conditions, including a boundary that is all land; the MAB
  configuration above runs that way.
- The nodal factors are evaluated once, at the reference date. Over a run of a year or more they
  drift, and the harmonics should be rebuilt.

### DIFF
--- a/docs/src/NumericalEarth.bib
+++ b/docs/src/NumericalEarth.bib
@@ -515,3 +515,50 @@
   year={1977},
   doi={10.1175/1520-0485(1977)007<0952:IMITUO>2.0.CO;2}
 }
+
+@book{schureman1958manual,
+  title={Manual of Harmonic Analysis and Prediction of Tides},
+  author={Schureman, Paul},
+  series={Special Publication 98},
+  publisher={United States Government Printing Office},
+  year={1958}
+}
+
+@book{kowalik2019modern,
+  title={Modern Theory and Practice of Tide Analysis and Tidal Power},
+  author={Kowalik, Zygmunt and Luick, John},
+  publisher={Austides Consulting},
+  address={Adelaide, Australia},
+  year={2019}
+}
+
+@article{egbert2002efficient,
+  title={Efficient inverse modeling of barotropic ocean tides},
+  author={Egbert, Gary D and Erofeeva, Svetlana Y},
+  journal={Journal of Atmospheric and Oceanic Technology},
+  volume={19},
+  number={2},
+  pages={183--204},
+  year={2002},
+  doi={10.1175/1520-0426(2002)019<0183:EIMOBO>2.0.CO;2}
+}
+
+@article{flather1976tidal,
+  title={A tidal model of the north-west European continental shelf},
+  author={Flather, Roger A},
+  journal={M{\'e}moires de la Soci{\'e}t{\'e} Royale des Sciences de Li{\`e}ge},
+  volume={10},
+  pages={141--164},
+  year={1976}
+}
+
+@article{chapman1985numerical,
+  title={Numerical treatment of cross-shelf open boundaries in a barotropic coastal ocean model},
+  author={Chapman, David C},
+  journal={Journal of Physical Oceanography},
+  volume={15},
+  number={8},
+  pages={1060--1075},
+  year={1985},
+  doi={10.1175/1520-0485(1985)015<1060:NTOCSO>2.0.CO;2}
+}
--- a/src/DataWrangling/DataWrangling.jl
+++ b/src/DataWrangling/DataWrangling.jl
@@ -417,6 +417,7 @@ include("GloBFP3D/GloBFP3D.jl")
 include("GHSL/GHSL.jl")
 include("CopernicusLandAlbedo/CopernicusLandAlbedo.jl")
 include("WorldCover/WorldCover.jl")
+include("TPXO/TPXO.jl")
 
 using .ETOPO
 using .ECCO
@@ -439,6 +440,7 @@ using .GloBFP3D
 using .GHSL
 using .CopernicusLandAlbedo
 using .WorldCover
+using .TPXO
 
 function dataset_modules()
     modules = Module[]
--- a/src/DataWrangling/TPXO/TPXO.jl
+++ b/src/DataWrangling/TPXO/TPXO.jl
@@ -1,0 +1,107 @@
+module TPXO
+
+export TPXO10Atlas
+
+using DocStringExtensions: TYPEDSIGNATURES
+using NCDatasets: Dataset
+using Oceananigans.Fields: Field, interior
+using Oceananigans.Grids: Bounded, Center, Face, Flat, LatitudeLongitudeGrid
+
+using NumericalEarth.Oceans: Oceans
+using ..DataWrangling: DataWrangling, AbstractStaticDataset, BoundingBox, inpaint_mask!
+
+download_TPXO_cache::String = ""
+function __init__()
+    global download_TPXO_cache = DataWrangling.download_cache("TPXO10")
+end
+
+"""
+    TPXO10Atlas()
+
+The TPXO10-atlas-v2 solution for the barotropic tide, on a 1/30° grid
+[Egbert and Erofeeva (2002)](@cite egbert2002efficient).
+
+TPXO is licensed, so its files cannot be downloaded: request them at https://www.tpxo.net, then
+unpack `grid_tpxo10atlas_v2.nc` together with the elevation and transport files of the constituents
+to be forced into `dir`.
+"""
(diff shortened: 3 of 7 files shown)


enri66:
The Codecov report here is not actually blocking: the `codecov/patch` status check passed (95.37%,
clears target), and the report comment's ❌ is only Codecov's own informational card.

The check that is actually failing and blocking merge is "Build documentation", and it is unrelated to
this patch. The traceback shows it failing during package precompilation, before any doc page is
built: `NumericalEarthVerosExt.jl`'s `__init__` calls into `PythonCall.C`, which fails inside
`CondaPkg.resolve()`. `NumericalEarthVerosExt.jl` has not changed since June 8, and the same
Documentation workflow fails on `main` itself as of three days ago
(https://github.com/NumericalEarth/NumericalEarth.jl/actions/runs/34337745039) — before this branch
existed. So this is a pre-existing CondaPkg/PythonCall environment issue on the doc-build runner, not
something introduced by `tides.jl` or `TPXO.jl`.

enri66:
Closing this for now. The tidal machinery in this PR (`TidalHarmonics`, `tidal_forcing`,
`tidal_boundary_conditions`) doesn't actually depend on anything NumericalEarth-specific — it's built
entirely on Oceananigans primitives, with only `TPXO10Atlas` (the dataset reader) belonging here. On
reflection that machinery should live in Oceananigans core, alongside the open-boundary schemes it
already uses (`GravityWaveRadiationBoundaryCondition`, `SurfaceWaveRadiationBoundaryCondition`), the
same way `ObliqueRadiation`/`TracerReservoir` do.

I'll reopen a smaller PR here containing only `TPXO10Atlas`, once the generic half has a home
upstream in Oceananigans. No action needed on this one — closing rather than leaving it open in a
shape that's about to change.

PR #680, opened 2026-09-12
Try to fix the CondaPkg error in the docs and remove near global example

simone-silvestri:


simone-silvestri:
This fix seems to work! (Examples should take about 7 hours to complete)

giordano (approved):

simone-silvestri:
Apparently, once we lift the CondaPkg error, an OOM error materializes while running the near-global config. I will remove it for the moment. When we move the examples to a bigger server and not only the poor 12GB gpus of tartarus we can re-enable it

glwagner (approved):

PR #682, opened 2026-09-12
Fix orca grid

simone-silvestri:
Apparently the ORCA grids have a duplicated periodic row. This PR removes it

### DIFF
--- a/src/Bathymetry/orca_grid.jl
+++ b/src/Bathymetry/orca_grid.jl
@@ -17,7 +17,7 @@ using ..DataWrangling.ORCA: ORCAOne, default_south_rows_to_remove
 # NEMO C-grid: T is the cell center, U the east face of T, V the north face of T, F the northeast corner.
 # eORCA quirks handled before constructing the grid:
 #
-#   - Duplicated east-edge periodic columns (`periodic_overlap_index`, `shift_face_x`).
+#   - Duplicated east-edge periodic columns (`periodic_overlap_index`, `shift_face_x`, `chop`).
 #   - Optional southern land padding rows (`south_rows_to_remove`, `chop`).
 #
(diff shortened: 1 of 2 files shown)


src/Bathymetry/orca_grid.jl
```diff
@@ -426,23 +428,21 @@ function ORCAGrid(arch = CPU(), FT::DataType = Float64;
...
     pole_idx = argmin(φFF[:, end])
-    north_poles_latitude = φFF[pole_idx]
-    first_pole_longitude = Float64(λFF[pole_idx])
-
-    Nx, Ny = size(λCC)
+    north_poles_latitude = φFF[pole_idx, end]
+    first_pole_longitude = Float64(λFF[pole_idx, end])
 
+    # eORCA repeats its first `ir` columns at the east edge, and a Periodic topology carries distinct cells only
+    ir = periodic_overlap_index(λCC)
```
xkykai: looking at this function I was slightly confused at first:

https://github.com/NumericalEarth/NumericalEarth.jl/blob/dbd2eb0f20cefb42f695332a8cd39feb7c330e98/src/Bathymetry/orca_grid.jl#L299-L307

why `min(div(Nx, 4), 10)`? are `4` and `10` magic numbers?

xkykai (approved):

src/Bathymetry/orca_grid.jl
simone-silvestri: You are right, this was trying to search what is the repeated index in the periodic direction of the grid. However we deal with immutable grids so we can hardcode what these rows are (through a method `periodic_overlap(::ORCADateset) = 2`)

jagoosw (approved):

PR #683, opened 2026-09-13
Update SoilGrids dataset to download from original ISRIC data source

bgroenks96:
This PR augments the SoilGrids2 dataset's dependency on our LRZ mirror with direct downloads from ISRIC (files.isric.org), reprojecting the native Homolosine (IGH) VRTs to EPSG:4326 via a new NumericalEarthArchGDALExt extension file, following the existing GHSL/OpenLandMap pattern.

It adds a resolution option (Grid250m, Grid1000m default, Clenshaw10km legacy) so callers can trade fidelity against download volume; Grid250m requires a bounded region (a global read would be ~300 GB/variable).

It also fixes a depth-unit bug: z_interfaces were in raw centimeters where meters were expected (bottom face at -200 m instead of -2 m), as well as two pre-existing export gaps that blocked `using NumericalEarth; SoilGrids2()` from working at all.

No Project.toml changes are necessary.

## Test plan

- [x] Offline unit tests (dataset interface, unit conversions, filename/region validation, raster geometry) — all pass
- [x] Real ISRIC download verified end-to-end (Grid250m) — physically plausible clay fractions in [0,1]
- [x] Legacy Clenshaw10km real-network path re-verified unchanged
- [ ] Grid1000m's full global download (~19 GB/variable) — impractical routinely; same code path is proven correct at Grid250m scale
- [ ] Verify docs build

This PR was co-authored by Claude Sonnet and Opus :robot:


src/DataWrangling/SoilGrids/SoilGrids.jl
```diff
@@ -27,16 +29,66 @@ remote islands, deserts, and the Arctic.
...
+- `Grid1000m`: downloads directly from ISRIC (`files.isric.org`) at 1000 m, reprojecting
+  each depth's Homolosine (IGH) VRT to EPSG:4326 with GDAL. Global by default (no region needed);
+  at 1000 m the full global 6-depth stack for one variable+statistic is ~19 GB.
+- `Grid250m`: the same ISRIC-direct pipeline at native resolution. A global read at 250 m
+  would be ~300 GB per variable, so `Grid250m` requires a bounded `region::BoundingBox`.
+
+Note that `Grid1000m`/`Grid250m` require `using ArchGDAL` (for the Homolosine → EPSG:4326
+reprojection).
+"""
+@enum Resolution Grid250m Grid1000m Clenshaw10km
```
xkykai, line 48: I feel like `Grid250m`,  `Grid1000m`, `Clenshaw10km` sound too generic for what they do. Perhaps `SoilGrids2Grid250m`, etc.? 

Also is `Resolution` necessary?

src/DataWrangling/SoilGrids/SoilGrids.jl
```diff
@@ -48,32 +100,140 @@ SoilGrids2_dataset_variable_names = Dict(
...
+end
+
+function DataWrangling.longitude_interfaces(dataset::SoilGrids2)
+    dataset.resolution === Clenshaw10km && return (0, 360)
+    return (-180, 180)
+end
+
+function DataWrangling.z_interfaces(dataset::SoilGrids2)
+    dataset.resolution === Clenshaw10km && return [-200, -100, -60, -30, -15, -5, 0]
+    return SoilGrids2_z_interfaces
```
xkykai, line 184:
```suggestion
    return [-2.0, -1.0, -0.6, -0.3, -0.15, -0.05, 0]
```

src/DataWrangling/SoilGrids/SoilGrids.jl
```diff
@@ -48,32 +100,140 @@ SoilGrids2_dataset_variable_names = Dict(
     :soil_organic_carbon     => "soc"
 )
 
-Base.size(::SoilGrids2, variable) = (3956, 1979, 6, 4)
+# Faces of the 100-200 / 60-100 / 30-60 / 15-30 / 5-15 / 0-5 cm intervals, increasing upward (m).
+const SoilGrids2_z_interfaces = [-2.0, -1.0, -0.6, -0.3, -0.15, -0.05, 0.0]
+
```
xkykai, line 105:
```suggestion
```

src/DataWrangling/SoilGrids/SoilGrids.jl
```diff
@@ -48,32 +100,140 @@ SoilGrids2_dataset_variable_names = Dict(
...
+# ISRIC depth-range URL fragments, deepest first (matches SoilGrids2_z_interfaces above).
+const SoilGrids2_depth_ranges = ("100-200cm", "60-100cm", "30-60cm", "15-30cm", "5-15cm", "0-5cm")
```
xkykai, line 107: localize this also?

src/DataWrangling/SoilGrids/SoilGrids.jl
```diff
@@ -48,32 +100,140 @@ SoilGrids2_dataset_variable_names = Dict(
...
 
 const SoilGrids2Metadatum = Metadatum{<:SoilGrids2}
 
-const SoilGrids2_url = "https://syncandshare.lrz.de/dl/fiVMyHskjL3FNbceuUFJev/soilgrids2_clenshaw989_10km.nc"
+const SoilGrids2_10km_url = "https://syncandshare.lrz.de/dl/fiVMyHskjL3FNbceuUFJev/soilgrids2_clenshaw989_10km.nc"
```
xkykai, line 111: localize this too?

ext/NumericalEarthArchGDALExt/soilgrids.jl
```diff
@@ -0,0 +1,62 @@
+#####
+##### SoilGrids2 ISRIC ingest: warp each depth's remote Homolosine (IGH) VRT directly to
+##### EPSG:4326 via GDAL's `/vsicurl` (no local intermediate GeoTIFF is downloaded — GDAL
+##### streams the needed bytes via HTTP range requests during the warp itself), stacking the
+##### six depths into the global or regional NetCDF the read path expects.
+#####
+
+const SOILGRIDS_IGH_PROJ4 = "+proj=igh +lat_0=0 +lon_0=0 +datum=WGS84 +units=m +no_defs"
```
xkykai, line 8:
```suggestion
```

ext/NumericalEarthArchGDALExt/soilgrids.jl
```diff
@@ -0,0 +1,62 @@
...
+
+# Warp one depth's remote VRT directly onto `raster`'s EPSG:4326 grid.
+function soilgrids_depth_window(source, raster)
+    configure_vsicurl!()
+    dataset = ArchGDAL.read(source)
+    data = try
+        ArchGDAL.gdalwarp([dataset],
+            ["-s_srs", SOILGRIDS_IGH_PROJ4,
```
xkykai, line 16:
```suggestion
            ["-s_srs", "+proj=igh +lat_0=0 +lon_0=0 +datum=WGS84 +units=m +no_defs",
```

src/DataWrangling/SoilGrids/SoilGrids.jl
bgroenks96: They are unexported, so the idea was to avoid the redundancy of `SoilGrids.SoilGrids250m` since the module qualifier already captures ownership. This would only be a problem if someone explicitly imported the name via `using NumericalEarth.DataWrangling.SoilGrids: Grids250m`.

That said, I don't mind changing it if you're unconvinced.

src/DataWrangling/SoilGrids/SoilGrids.jl
bgroenks96: Is that the standard style in `NumericalEarth`? I think for something like a URL I would prefer to have it explicitly listed as a constant at the top level for visibility rather than inlined in the code?

src/DataWrangling/SoilGrids/SoilGrids.jl
```diff
@@ -27,16 +29,66 @@ remote islands, deserts, and the Arctic.
...
+"""
 @kwdef struct SoilGrids2 <: AbstractStaticDataset
     "Specifies which dataset layer to load variables from; see [SoilGrids.Statistic](@ref). Defaults to `Mean`"
     statistic::Statistic = Mean
+    "Specifies how the data is sourced; see [`Resolution`](@ref). Defaults to `Grid1000m`"
+    resolution::Resolution = Grid1000m
 end
 
+Base.summary(dataset::SoilGrids2) =
+    string("SoilGrids2(statistic = ", dataset.statistic, ", resolution = ", dataset.resolution, ")")
```
glwagner, line 89:
```suggestion
Base.summary(dataset::SoilGrids2) = string("SoilGrids2(statistic = ", dataset.statistic, ", resolution = ", dataset.resolution, ")")
```

src/DataWrangling/SoilGrids/SoilGrids.jl
glwagner: as long as it can be changed by the user I feel it doesn't matter much. Note the name "SoilGrids2_10km_url" is slightly out of convention for variable names since it includes TitleCase. No big deal but consider a name like "clenshaw_10km_url" without the module prefix (since it is already in SoilGrids2 module)

src/DataWrangling/SoilGrids/SoilGrids.jl
glwagner: I also find the names to be cryptic. If they are only internal to the module, it's not a big deal. If they come out into user scripts, then we indeed will encounter `Grids250m` in the wild.

bgroenks96: I suppose my thinking was that a user would write `SoilGrids.Grid250m`. There might be a better word than "grid" here; the ideal would be to somehow convey that it's a grid based on meter spacing rather than spherical.

PR #684, opened 2026-09-14
Accounting for latent heat of fusion from PrescribedAtmosphere iceberg freshwater flux 

simone-silvestri:
At the moment, snow and icebergs which melt in the ocean provide only a freshwater flux, but the ocean spends heat to melt them equivalent to the freshwater flux times the latent heat of fusion.

This PR adds this heat component to the coupled model. This requires splitting the freshwater fluxes into rivers and icebergs and treat them separately


xkykai (approved):

glwagner (approved):
seems right

glwagner:
The title was pretty alarming until you read the PR description.

PR #685, opened 2026-09-14
Ice meltwater comes in at the correct temperature

simone-silvestri:
Ice meltwater comes in the ocean at 0 degrees at the moment, which is a bit high compared to the freezing point of water (~1.8 degrees). This means that freshwater coming from the ice, as it stands, heats the ocean by the difference in that temperature.

This PR changes the behavior by making ice meltwater enter at the interface temperature making meltwater essentially neutral (not heating not cooling)


xkykai (approved):

src/EarthSystemModels/InterfaceComputations/component_interfaces.jl
```diff
@@ -145,6 +145,7 @@ struct SeaIceOceanFluxes{C, FX, FY}
     frazil_heat    :: C
     salt           :: C
     freshwater     :: C
+    freshwater_heat_content :: C
```
glwagner, line 148: should this be `freshwater_temperature` ? What is "heat content"?

src/Oceans/assemble_net_ocean_fluxes.jl
```diff
@@ -129,7 +129,7 @@ Base.@propagate_inbounds get_land_freshwater_flux(i, j, flux) = flux[i, j, 1]
     Jᵀ = net_ocean_fluxes.T
     Jˢ = net_ocean_fluxes.S
     Jʷ = net_ocean_fluxes.η
-    Jᴴ = net_ocean_fluxes.freshwater_heat_content # Σᵢ Tᵢ Jʷᵢ — atmosphere freshwater enters at SST
+    Jᴴ = net_ocean_fluxes.freshwater_heat_content # Σᵢ Tᵢ Jʷᵢ — atmosphere freshwater enters at SST, ice meltwater at Tb
```
glwagner, line 132: why is a flux referred to as "heat content"?

src/EarthSystemModels/InterfaceComputations/component_interfaces.jl
simone-silvestri: Well, in practice it is T * Fw, which is the "heat content of the incoming or outgoing freshwater" measured as a temperature flux

simone-silvestri: freshwater_temperature_content?

glwagner: shouldn't we specify the temperature, since Fw is given by the flux? That way the properties are truly independent.

glwagner: In other words, lets specify `Tw` and `Fw`, rather than `Fw` and `T * Fw` ?

PR #686, opened 2026-09-14
Speed up our examples

simone-silvestri:
A one degree simulation should be able to handle a 2 hours timestep.

Updated the eddy closure to use TriadIsopycnalSkewSymmetricDiffusivity and changed the simulation timestep from 20 minutes to 120 minutes.


glwagner:
> A one degree simulation should be able to handle a 2 hours timestep.
> 
> Updated the eddy closure to use TriadIsopycnalSkewSymmetricDiffusivity and changed the simulation timestep from 20 minutes to 120 minutes.

I thought we were removing "triad" from the name?

simone-silvestri:
yeah, we can wait for https://github.com/CliMA/Oceananigans.jl/pull/5955

PR #687, opened 2026-09-15
Include Breeze `0.11`

dkytezab:
cc @jtbuch

PR #689, opened 2026-09-15
Support Breeze 0.11: `ρE` boundary-condition keys and the `base_pressure` datum

### LINKED ISSUE #488: Nested child's hydrostatic anchor uses terrain-surface pressure where Breeze expects a z = 0 datum
`mean_surface_pressure` (`ext/NumericalEarthBreezeExt/breeze_nested_atmosphere.jl:273`) computes the
domain-mean **surface** pressure over the child grid and passes it to `CompressibleDynamics` as
`surface_pressure`. Breeze defines that argument at **z = 0**, not at the terrain.

## The contract

`converged_hydrostatic_pressure` integrates from `z = 0` with `p(0) = p₀` up to each column's terrain
height, and `terrain_surface_reference_fields` uses it to derive the per-column surface reference
`pˢ`. Breeze's regression test asserts this directly: with the default `p₀`, it checks

```julia
z = znode(i, 1, k, grid, Center(), Center(), Center())   # absolute height above the z = 0 datum
p_exact = hydrostatic_pressure(z, 101325.0, θ₀, pˢᵗ, constants)
@test pᵣ[i, 1, k] ≈ p_exact rtol=1e-2
```

over terrain rising 0 to 1000 m (`test/terrain_following_reference_state.jl:25-72`). It then asserts
that at a given `k` the reference must differ between the flat and peak columns while agreeing at
equal physical height, which is the `pᵣ = pᵣ(z)` invariant the perturbation form depends on.

Two distinct steps are involved, and only the first is wrong here. `p₀` must be a sea-level (z = 0)
pressure, and each column must then be anchored at the value hydrostatically *reduced* from it to
that column's terrain height. Breeze does the second correctly. The caller supplies the wrong
quantity for the first, passing a value that is already at terrain level, so the reduction is
applied twice.

## Why this matters now

Before #863 in Breeze the reference state was not used by the pressure-gradient force by default, so
the anchor was largely inert. #863 made the perturbation form the default specifically to "suppress
the small-difference-of-large-numbers cancellation in the terrain-following PGF that otherwise
drives spurious slope-correlated surface winds over steep terrain". A mis-anchored reference
delivers a fraction of that benefit.

## Measured

8x6x50 nested child at 1/4 degree over a Colorado box (mean terrain 1921 m), Breeze#main +
NumericalEarth 0.6.0, CPU:

| | current | corrected `p₀` |
|---|---|---|
| `p` at k=1 | 80839.8 Pa | 80839.8 Pa (identical) |
| `pᵣ` at k=1 | 64615.2 Pa | 82445.6 Pa |
| \|p'\| at k=1 | 16224.6 Pa (**20.1 % of p**) | 1630.9 Pa (**2.0 %**) |
| \|p'\| domain-wide | 10041.1 Pa | 760.9 Pa |

`p` is unchanged between the two, confirming this is a decomposition and conditioning problem rather
than a wrong force: the reference state sits 16.2 kPa below the field it exists to track at the
terrain surface, and correcting the anchor collapses that by roughly 10x at the surface and 13x
domain-wide.

The reference value of 64615 Pa is approximately 81159 Pa reduced through another ~1900 m, which is
the signature of the reduction being applied a second time.

## Four independent routes agree on the magnitude

1. `converged_hydrostatic_pressure` integrates from `z = 0` with `p(0) = p₀`.
2. Breeze's default `p₀` is 101325 Pa, and the test above asserts against it at absolute height.
3. For this box, mean sea level pressure from the forcing data is 100798 Pa and the hydrostatically
   reduced surface mean is 102033 Pa, agreeing within 1.2 %, while the value currently passed
   (81159 Pa) sits 19.6 kPa below both. That gap is the O(rho g h) scale with h the domain's mean
   elevation: 1.0 * 9.81 * 1921 = 18.8 kPa.
4. The measured 10x collapse in |p'| when the anchor is corrected.

Worth noting that Breeze's default of 101325 Pa is 527 Pa from the mean sea level value here, so
passing nothing at all would be about 37x closer than the value currently computed.

## Scope

The defect is in the shared anchor computation, not in any dataset reader, so it affects any nested
parent dataset. It is invisible over ocean or low terrain, where surface and sea-level pressure
nearly coincide, which is likely why it has gone unnoticed. It is also invisible to Breeze's own
tests, whose terrain rises from z = 0 where the two conventions are the same number.

## Fix options

**Superseded, see the next section.** The original recommendation here was to read the single-level
dataset's mean-sea-level pressure field, or hydrostatically reduce the domain-mean surface pressure
to z = 0 using Breeze's own reference integration. That is correct for the terrain reference state,
but it is not safe on its own and cannot be done in NumericalEarth alone.

## This cannot be fixed here alone

`surface_pressure(dynamics)` is read as three different quantities in Breeze, tracked as
NumericalEarth/Breeze.jl#887:

| reading | consumers |
|---|---|
| pressure at the column's bottom face | flat reference state, `HydrostaticallyBalancedDensity` |
| pressure at absolute z = 0 | terrain reference state (what this issue is about) |
| pressure at the ground | bulk sensible heat, vapour, drag, and stability functions |

`atmosphere_model.jl:184` passes the same scalar into the boundary conditions, so the value this
issue asks NumericalEarth to change is also the one the surface fluxes use. On a plateau at 1921 m,
with true ground pressure 79672 Pa:

| anchor | reference error | θ₀ error | ρ₀ error (drag) |
|---|---|---|---|
| terrain-level 81159 Pa (passed today) | -16236.2 Pa | -1.69 K | 1.9 % |
| sea-level 100798 Pa (the fix proposed above) | -0.0 Pa | -20.79 K | 26.5 % |

Bulk drag is `-ρ₀ Cᴰ |U| u`, so the 26.5 % density error is a 26.5 % error in surface momentum drag,
and it scales the heat and moisture fluxes too. No single scalar zeroes both columns, so any value
NumericalEarth passes only chooses which consumer to break.

## Recommended sequencing

1. **Breeze disambiguates** (NumericalEarth/Breeze.jl#887). That issue also documents a second
   problem this depends on: the flat and terrain reference states currently use different datums, so
   the meaning of the parameter is not settled even within Breeze.
2. **NumericalEarth then passes a genuine z = 0 pressure** to the datum accessor: single-level `msl`,
   or the domain-mean surface pressure hydrostatically reduced via Breeze's own reference
   integration. The latter avoids MSLP's below-ground extrapolation assumptions, which are least
   reliable over exactly the high terrain where this matters.

Doing (2) before (1) trades a conditioning problem for a larger physical error in any configuration
that uses bulk surface fluxes. It is safe in configurations carrying none, which is why the nested
LAM case in this report is unaffected today.

## Solution-level evidence: retracted

An earlier version of this section reported that correcting the anchor reduced peak |w| at 1500 m by
2.5x. **That attribution was wrong.** The two runs compared differed in two variables rather than
one: they had different anchors, and one of them also carried an unrelated fix to the nested child's
terrain blending. The effect belongs to the terrain fix.

Completing the 2x2 on the same case (8x6x50 child at 1/4 degree over the Colorado Rockies, 3 s, peak
interior |w| at 1500 m):

| | terrain-level anchor (81159 Pa) | sea-level anchor (100798 Pa) |
|---|---|---|
| without the terrain-blend fix | 1.3529 m/s | 1.3529 m/s |
| with the terrain-blend fix | 0.5402 m/s | 0.5402 m/s |

The anchor has no measurable effect on |w| in either row; the terrain fix accounts for the entire
reduction. Runs within a row agree to every digit.

The honest position is therefore stronger than the "not established" this section originally
replaced: swapping the anchor across a ~19.6 kPa datum change produced **no measurable change** in
near-surface vertical motion in this configuration. That is one domain, one case and 3 s of
integration, so it does not rule out an effect on longer runs or steeper terrain, but it is a
measurement rather than a gap.

The reference-state error itself is unaffected by this retraction: the 16.2 kPa figure above is
measured directly from `pᵣ` against the state, not inferred from |w|.

ewquon:
Updates NumericalEarth for Breeze 0.11, which is breaking in two ways that reach this package.

**Boundary-condition and forcing keys** (NumericalEarth/Breeze.jl#974). Breeze now validates every key at construction. The `energy_bc_key` returned was a `pkgversion` gate resolving to `ρs`, which is only valid with `formulation = :StaticEnergy` — not on the potential-temperature model the nested LAM runs. It now returns Breeze's exported `total_energy_density_name` (`ρE`), the interface key that names the energy input rather than a prognostic, so Breeze routes it to whichever thermodynamic variable the formulation makes prognostic.

**The reference-pressure keyword** (NumericalEarth/Breeze.jl#893). `surface_pressure` was the datum at z = 0; it is now `base_pressure`, while `surface_pressure` names the derived 2D field at each column's bottom face. `CompressibleDynamics`, `ReferenceState`, `ExnerReferenceState`, `MoistAirBuoyancy` and `ParcelDynamics` all raise on the old spelling.

Only the Breeze-facing calls move. NumericalEarth's own `surface_pressure` keywords keep their spelling, so `atmosphere_model` and `nested_atmosphere_model` are unchanged for callers; the translation happens where the dynamics is built. `HydrostaticallyBalancedDensity` needs no change — it wants the pressure at the ground and kept the old spelling.

Closes #690 and supersedes #687 (and the since-closed #677). Those widen the bound to `0.10, 0.11` and change nothing else, which is potentially problematic on its own: the code they would ship still passes `ρs` and `surface_pressure`, so allowing 0.11 lets an environment resolve to a combination that raises at model construction rather than adding support for it. The range is also not one this package can honor after the migration: `total_energy_density_name` does not exist in 0.10, so `energy_bc_key` raises `UndefVarError` the first time a model is built. None of them moves the Oceananigans bound either, which Breeze 0.11 requires at 0.112. Hence `0.11` replacing the bound rather than extending it — which also stops CompatHelper regenerating the same PR.

Closes #488. This is also the `ρE`-unconditionally follow-up noted in the closing comment of #668, now that Breeze#974 has landed and tagged.

## Behavior

The first three commits are renames. The fourth is not, and it is here because 0.11 reinterprets the datum: 0.10 applied a scalar `surface_pressure` at the domain bottom; 0.11 takes a scalar `base_pressure` as the value at z = 0 then reduces it hydrostatically to each column's bottom face, giving a 2-D `surface_pressure` field. That reduction is what lets the reference surface pressure follow the terrain, but it means the input scalar means something different.

NumericalEarth was passing the domain mean of ERA5 `sp`, the pressure at the dataset's ground — approximately right under 0.10, where nothing reduced it. Under 0.11, the correct base pressure is the mean sea level pressure, which is by definition at z = 0 and already available alongside `surface_pressure` in the ERA5 single-level tables. 93dee10 changes the applied scalar from `sp` to `msl`. Because the reference state moves, results with ERA5 also change.

Note: A per-column datum is supported but is the wrong tool. An `sp` map double-counts terrain outright, and an `msl` map bakes the synoptic pattern into a reference state meant to be smooth and time-invariant. The scalar keeps the weather in the perturbation.

## Testing

`test/test_nested_simulation.jl` built its boundary conditions on `ρ` and `ρe`, which 0.10 dropped silently and 0.11 rejects; they become `ρᵈ` and `ρE`. Checked directly against Breeze 0.11 that the migrated shape constructs, passes lateral values through unconverted, and steps finite — and that each pre-migration spelling (`ρs`, `ρe`, `ρ`, and `CompressibleDynamics(; surface_pressure)`) raises the `ArgumentError` 0.11 intends.

**The suite passes against a ClimaSeaIce build carrying CliMA/ClimaSeaIce.jl#175** — 3226 pass, 2 broken, 0 fail, including `test_nested_simulation`. **Until that merges and is released, NumericalEarth cannot resolve Breeze 0.11 at all**: Breeze 0.11 requires Oceananigans 0.112, and released ClimaSeaIce caps it at 0.111. That is also why this PR moves the Oceananigans bound in each workspace subproject, which had pinned `0.111` independently of the root.

The datum change is not covered by CI — it needs ERA5, which lives in `DataTests.yml` (nightly, or on a PR labeled `run data tests`). It was checked independently instead: over a terrain-spanning ERA5 box the `sp`–`msl` difference matches that box's measured mean surface geopotential to within 1%, confirming the gap is the hydrostatic terrain reduction Breeze now performs.

https://claude.ai/code/session_01AWWn9XNLXVywosTVY8uS9z


docs/Project.toml
```diff
@@ -35,13 +35,13 @@ Suppressor = "fd094767-a336-5f1f-9728-57cf17d0bbfb"
...
+Breeze = "0.11"
 CUDA = "6"
 CairoMakie = "0.15"
 ConservativeRegridding = "0.2"
 Documenter = "1"
 DocumenterCitations = "1.3"
 Enzyme = "0.13.166"
 Literate = "2.2"
-Oceananigans = "0.111"
+Oceananigans = "0.111, 0.112, 0.113"
```
giordano: Can't support 0.111 and v0.112 at the same time since various keyword arguments were renamed

navidcy:
Atm ClimaSeaIce imposes Oceananigans v0.111 and Breeze Oceananigans v0.112

https://github.com/CliMA/ClimaSeaIce.jl/commit/db634bc050accb17198f0a9bca1497b69fe4b2f8#commitcomment-200807540 should resolve this issue.

docs/Project.toml
giordano: Sorry, I got mixed up between Breeze and Oceananigans. But Oceananigans v0.113 did rename one keyword argument (see https://github.com/CliMA/Oceananigans.jl/releases/tag/v0.113.0), I don't know if it's used here, but we used it in Breeze.jl

giordano: > But Oceananigans v0.113 did rename one keyword argument (see [CliMA/Oceananigans.jl@`v0.113.0` (release)](https://github.com/CliMA/Oceananigans.jl/releases/tag/v0.113.0)), I don't know if it's used here, but we used it in Breeze.jl

Yes, [it's used in a few examples](https://github.com/search?q=repo%3ANumericalEarth%2FNumericalEarth.jl+overwrite_existing&type=code)

navidcy: I'll deal with this in a new PR

navidcy: https://github.com/NumericalEarth/NumericalEarth.jl/pull/691

ewquon:
@giordano stacked #691 on top of this -- any other recommendations or is this OK to merge?

giordano:
It'd be good to get https://github.com/CliMA/ClimaSeaIce.jl/pull/177 merged and tag a new release, so that we can do a full testing.  I think that's ready, I just don't know whether @navidcy wanted to wait for something else.

giordano (approved):

PR #691, opened 2026-09-17
Supports Oceananigans v0.113 by changing `overwrite_existing` -> `overwrite_files`

navidcy:


src/DataWrangling/DataWrangling.jl
```diff
@@ -210,8 +210,8 @@ end
 ##### FieldTimeSeries utilities
 #####
 
-function save_field_time_series!(fts; path, name, overwrite_existing=false)
-    overwrite_existing && rm(path; force=true)
+function save_field_time_series!(fts; path, name, overwrite_files=false)
+    overwrite_files && rm(path; force=true)
```
giordano, line 214: Note: this is in the source code of this package, doesn't come from Oceananigans, but I guess it could be changed for consistency (but this will be in turn a breaking change in this package)

giordano (approved):

src/DataWrangling/DataWrangling.jl
navidcy: oh! that was just me blindly search-and-replace!
it does make sense to change it though... shall we?

navidcy:
I guess I hit package resolve issues; perhaps all changes in this PR should be moved to #689?

giordano:
The comment in https://github.com/CliMA/ClimaSeaIce.jl/pull/177#issuecomment-5714545735 was meant to be posted here:

> We still need to wait for https://github.com/NumericalEarth/Breeze.jl/pull/1001 to actually test the new version via Breeze.jl (waiting for approval about some slightly different numerical results, but after that we can quickly release a new version).

ewquon:
Thanks for moving this along, @giordano @navidcy

PR #695, opened 2026-09-18
Rotate parent winds for a child with `LambertConformalConicGrid`

ewquon:
The state exchanger copies the parent's eastward/northward winds into the child-prognostic `ρu`/`ρv` and the specific `u`/`v` unchanged. A `LatitudeLongitudeGrid` child's axes are east/north, so that is exact. A `LambertConformalConicGrid` child's axes are rotated from east by the cone convergence angle `n (λ − λ₀)`, which reaches ±20° at the walls of a CONUS-size panel (450 × 290 at 13 km, tangent cone at 38.5° N). Every consumer of the exchanger's FTS — the lateral boundary conditions, the Davies relaxation targets, and the initial condition — then imposes the wrong components: a cross-component error of 0.34 |V| at 20°, and the wall-normal inflow taken from the wrong component.

This applies the rotation once, in the exchanger kernel, from the child grid's `conformal_mapping`; the three consumers are unchanged. `nested_atmosphere_model` passes the child grid through to `state_exchanger`, whose `child_grid` keyword defaults to `nothing` (identity, as for a lat-lon child).

The analytic angle agrees with Oceananigans' finite-difference `rotation_angle` to 5 digits and follows the `intrinsic_vector` sign convention. The test drives a uniform 10 m s⁻¹ eastward parent wind into a child sitting 200–600 km east of its central meridian, interpolates the exchanged `u`/`v` onto the child, and checks that `extrinsic_vector` on the child grid recovers (10, 0) at every node to 1e-2.

The testset runs on every test architecture, so the CI GPU job exercises the rotation inside the exchanger kernel (the child-frame map is an isbits `LambertConformalConic{FT}` kernel argument).


glwagner:
@ewquon just want to make sure this PR is correct --- the winds should be rotated for _any `OrthogonalSphericalShellGrid`_ --- not only the Lambert conformal conic. Does the PR account for this?

ext/NumericalEarthBreezeExt/breeze_state_exchanger.jl
```diff
@@ -64,9 +66,23 @@ end
 const PrognosticStateFTS = FieldTimeSeries{<:Any, <:Any, <:Any, <:Any, <:PrognosticStateBackend}
 update_field_time_series!(::PrognosticStateFTS, ::Time) = nothing
 
+# The child's velocity components follow its grid axes: eastward/northward on a `LatitudeLongitudeGrid`,
+# rotated by the cone convergence angle on a `LambertConformalConicGrid`.
+child_frame(grid) = nothing
+child_frame(grid::LambertConformalConicGrid) = grid.conformal_mapping
```
glwagner, line 72:
```suggestion
child_frame(grid::OrthogonalSphericalShellGrid) = grid.conformal_mapping
```

ext/NumericalEarthBreezeExt/breeze_state_exchanger.jl
```diff
@@ -64,9 +66,23 @@ end
...
+
+@inline rotate_to_child_frame(::Nothing, i, j, k, grid, u, v) = (u, v)
+
+@inline function rotate_to_child_frame(map::LambertConformalConic, i, j, k, grid, u, v)
```
glwagner, line 76: is this the same as the `extrinsic_vector` helper in Oceananigans?

ext/NumericalEarthBreezeExt/breeze_state_exchanger.jl
glwagner: this may not be necessary if the vector helpers in Oceananigans already perform this dispatch.

glwagner (commented):
I think this is partly duplicating functionality in Oceananigans here:

https://github.com/CliMA/Oceananigans.jl/blob/9c3be77cbfb66fca0352487036995bb5c666759c/src/Models/HydrostaticFreeSurfaceModels/set_hydrostatic_free_surface_model.jl#L108-L139

although, that stuff is within `Models/HydrostaticFreeSurfaceModels`. I am not sure if there is an opportunity to reduce overlap. To start, perhaps we can use the same terminology here.

ext/NumericalEarthBreezeExt/breeze_state_exchanger.jl
glwagner: I think the function signature is just 

```julia
extrinsic_vector(i, j, k, grid, u, v)
```

note that it is redundant to provide the conformal map AND the grid. I also believe that `extrinsic_vector` performs the dispatch you need and you do not need to add extra dispatch here.

Note, this is not specific to the concept of a "child simulation" and therefore is named so

